PR #2, opened 2019-09-30
Put profiling initialization behind DMTR_PROFILE macro

zhangwen0411:
In `posix_queue.cc`, define `alloc_latency` only when the `DMTR_PROFILE` macro is set (otherwise, profiling-related type and variables are undefined).


iyzhang:
Seems fine to me

PR #4, opened 2019-10-22
Segmentation fix

isaac-ped:
This PR addresses three issues:

1) Upgrades to dpdk 19.08
2) Fixes the segmentation issue. Packets larger than an MTU are now able to be received
3) sga segment buffers are now pointers to location in larger sgaseg_buf (in lwip stack), which matches the implementation in the other stacks.


.gitmodules
```diff
@@ -18,18 +18,22 @@
 	path = submodules/redis-posix
 	url = https://github.com/mlr-msft/redis.git
 	branch = msft/4.0.9/mirobert
+    ignore = dirty
```
maxdml: Should we update the submodules repositories with a fork of yours?

src/c++/libos/lwip/lwip_queue.cc
```diff
@@ -1399,35 +1481,42 @@ dmtr::lwip_queue::parse_packet(struct sockaddr_in &src,
...
+        // Allocate enough space for all sga segments
+        size_t sga_total_size = ntohs(ip_hdr->total_length) - sizeof(*ip_hdr) - sizeof(*udp_hdr) - sizeof(uint32_t);
+        DMTR_OK(dmtr_malloc(&sga.sga_buf, sga_total_size));
+
+        // Read into the sga buffer
+        const void *read_buf = rte_pktmbuf_read(pkt, offset, sga_total_size, sga.sga_buf);
+        // If read_buf is not pointing to sga_buf, it's because no copy was necessary
+        // (to avoid freeing problems later, copy into sga_buf anyway)
+        if (read_buf != sga.sga_buf) {
+            rte_memcpy(sga.sga_buf, read_buf, sga_total_size);
```
maxdml: Can't we just free sga.sga_buf and set it to be read_buf? The documentation says when no copy is necessary, it's because the data was continuous (in one segment, I assume). We would have to, instead of free(), call the mbuf free to return the mbuf to the pool, but that can be part of a dmtr_free() call that knows how to distinguish a mbuf from a hoard/glibc malloc'ed returned pointer. A mbuf would not be freed, as tx_burst would free it automatically.

That's is likely not beneficial if freeing is more expensive than copying (which I assume not w/o really being sure)

src/c++/libos/lwip/lwip_queue.cc
```diff
@@ -1399,35 +1481,42 @@ dmtr::lwip_queue::parse_packet(struct sockaddr_in &src,
...
+        }
+
+        char *offset_buf = static_cast<char*>(sga.sga_buf);
         for (size_t i = 0; i < sga.sga_numsegs; ++i) {
-            // segment length
-            auto seg_len = ntohl(*reinterpret_cast<uint32_t *>(p));
+            uint32_t seg_len = ntohl(*(uint32_t*)offset_buf);
```
maxdml: Isn't it a case where we want static_cast?

src/c++/libos/lwip/lwip_queue.cc
```diff
@@ -1399,35 +1481,42 @@ dmtr::lwip_queue::parse_packet(struct sockaddr_in &src,
...
+            offset_buf += sizeof(seg_len);
```
maxdml: Isn't it clearer to explicitly give `uint32_t` as an argument to `sizeof()`? I am always confused with regard to best practices here.

src/c++/libos/lwip/lwip_queue.cc
```diff
@@ -1502,8 +1591,10 @@ int dmtr::lwip_queue::rte_eth_tx_burst(size_t &count_out, uint16_t port_id, uint
     DMTR_TRUE(ERANGE, ::rte_eth_dev_is_valid_port(port_id));
     DMTR_NOTNULL(EINVAL, tx_pkts);
 
-    size_t count = ::rte_eth_tx_burst(port_id, queue_id, tx_pkts, nb_pkts);
+    // If count_out packets were sent on the previous burst, they don't have to be sent again
+    size_t count = ::rte_eth_tx_burst(port_id, queue_id, &tx_pkts[count_out], nb_pkts - count_out);
```
maxdml: I think we should handle partial writes outside of this function, i.e. update nb_pkts according to the sent count.

src/c++/libos/lwip/lwip_queue.hh
```diff
@@ -66,8 +67,11 @@ class lwip_queue : public io_queue {
     private: static std::map<lwip_4tuple, std::queue<dmtr_sgarray_t> *> our_recv_queues;
     private: static std::unordered_map<std::string, struct in_addr> our_mac_to_ip_table;
     private: static std::unordered_map<in_addr_t, struct rte_ether_addr> our_ip_to_mac_table;
-    private: static uint16_t my_app_port; //FIXME this could/should be a list
-    public: static void set_app_port(uint16_t port) { my_app_port = port; }
+
+    // ???: Should this be static? (all queues share the same my_app_ports instance?)
```
maxdml: The reason why sharing the port list across queues is needed at the moment is because a rx_burst can dequeue packets destined to other queues. If we multithread this stack, each net worker should have its own list of ports.

src/c++/libos/lwip/lwip_queue.hh
```diff
@@ -66,8 +67,11 @@ class lwip_queue : public io_queue {
...
+    // TODO: Some mechanic for unregistering ports from the application?
```
maxdml: dmtr_close() should

src/c++/apps/echo/config.yaml
```diff
@@ -9,19 +9,27 @@ server:
     bind:
         host: 192.168.2.10
         port: 6789
-dpdk:
-    eal_init: [
+
+? &IPED_EAL
+    ["-c", "0xe",
```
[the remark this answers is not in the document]
isaac-ped: Got it. Changed to match yours

.gitmodules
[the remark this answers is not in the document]
isaac-ped: I do not actually have my own fork at this time.
I'm working on the multithreaded version but I think that can be managed as a separate submodule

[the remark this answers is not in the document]
isaac-ped: The differences with the version automatically checked out are managed as a patch, similar to the other repos

src/c++/apps/echo/dmtr_http_server_threaded.cc
```diff
@@ -772,6 +775,8 @@ int work_setup(Psp &psp, bool split,
         if (pthread_create(&worker->me, NULL, &net_worker, (void *) worker.get())) {
             log_error("pthread_create error: %s", strerror(errno));
         }
+
+        //FIXME: optional pinning has been removed. Remove from options to reflect that
```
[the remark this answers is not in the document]
isaac-ped: Was talking about the --no-pin-* features which I had already removed, and forgot to remove this FIXME

src/c++/apps/echo/dmtr_rate_client.cc
```diff
@@ -678,6 +679,8 @@ int long_lived_processing(double interval_ns, uint32_t n_requests, std::string h
         token = tokens[idx];
         tokens.erase(tokens.begin()+idx);
         if (status == 0) {
+            token = tokens[idx];
+            tokens.erase(tokens.begin()+idx);
```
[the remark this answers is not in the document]
isaac-ped: Think this got put back in during an ugly merge. Removed again.

src/c++/apps/echo/dmtr_rate_client.cc
```diff
@@ -756,6 +760,8 @@ int long_lived_processing(double interval_ns, uint32_t n_requests, std::string h
         } else {
             assert(status == ECONNRESET || status == ECONNABORTED);
             DMTR_OK(dmtr_close(wait_out.qr_qd));
+            token = tokens[idx];
+            tokens.erase(tokens.begin()+idx);
```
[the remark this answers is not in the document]
isaac-ped: likewise

src/c++/libos/lwip/lwip_queue.cc
isaac-ped: you mean instead of the c-style cast?
Since we're casting from a `void*`, either static_cast or reinterpret_cast should work. I think they compile to the same thing.
You're right though - using c-style cast is not recommended. I did it because I was lazy and think that static_cast looks ugly, but I just found this quote from Stroustrup:
```
This form of explicit type conversion is inherently unsafe and ugly. Consequently, the notation used, static_cast, was designed to be ugly.
```

isaac-ped: Oops, actually since `offset_buff` is a `char *`, not a `void *` (because we want to do arithmetic with it), we do have to use reinterpret_cast

src/c++/libos/lwip/lwip_queue.cc
isaac-ped: My thoughts are that the pointer should be advanced by the size of `seg_len`, regardless of what type `seg_len` is. If we later change it to be a uint16_t, then the pointer should advance by that much, and we'd have to change code in two places to make that change. This way, we just have to change the type of seg_len directly.
I'm not incredibly tied to it though. I don't think there are strict guidelines on it -- especially in c++, where this type of pointer arithmetic is sorta frowned upon anyway.

src/c++/libos/lwip/lwip_queue.hh
isaac-ped: Yeah, makes sense. Adding a note to that effect in the code.

src/c++/libos/lwip/lwip_queue.hh
isaac-ped: That requires looking up the port to which a socket is connected upon `close()`.
Also:
* accepted sockets close() should not unregister a port
* listening sockets close() should unregister the listening port, if and only if there are no active accepted sockets
* connect'ed sockets close() should unregister the connected-to port, if and only if there are no active accepted sockets.

I think that the proper way to do this would be to keep a counter of the number of connections per-port. It is more complicated than just unregistering on close() though

src/c++/libos/lwip/lwip_queue.cc
```diff
@@ -1506,7 +1506,7 @@ dmtr::lwip_queue::parse_packet(struct sockaddr_in &src,
 
         char *offset_buf = static_cast<char*>(sga.sga_buf);
         for (size_t i = 0; i < sga.sga_numsegs; ++i) {
-            uint32_t seg_len = ntohl(*(uint32_t*)offset_buf);
+            uint32_t seg_len = ntohl(*reinterpret_cast<uint32_t*>(offset_buf));
```
maxdml: Shouldn't that be a static cast, if reinterpret cast is made for "complex" data structures (e.g. non native types)?

PR #6, opened 2020-05-20
Optimize catnip latency

kwzhao:
The first and second commits bring the rustc compiler and all dependencies up-to-date.

The third commit replaces the hashing algorithm with a faster one. This speeds things up quite a bit.

The fourth and fifth commits reduce the number of hashmap and hashset operations on the critical path for a few more microseconds' worth of speed-up. I'm not 100% confident in the correctness of these transformations, but they don't cause any extra test failures.

The last commit changes catnip's allocator to mimalloc.

Prior to these changes, there were two unit test failures, and they still remain:

```
failures:
    protocols::tcp::peer::tests::multiple_writes
    protocols::tcp::peer::tests::retransmission_ok
```


iyzhang:
Thanks for these updates!

PR #15, opened 2021-06-08
Updating Git Ignore

ppenna:
# Description

In this pull request I add the following changes:

- Cleanup vscode files from source tree
- Update git ignore file accordingly


gatoWololo (approved):
Looks good!

PR #18, opened 2021-06-08
Bad CMake Target for LibOS Catnip

ppenna:
# Description

In this Pull Request, I fix the bad target name for LibOS Catnip.

### DIFF
--- a/src/rust/catnip_libos/CMakeLists.txt
+++ b/src/rust/catnip_libos/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.10)
 
-set(CATNIP_LIBOS_TARGET catnip_libos)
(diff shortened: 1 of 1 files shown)


gatoWololo (approved):
It seems this PR is a strict superset of #15, i.e. it contains all the changes in #15 and more. Not sure what will happen when we try to merge it to master... let's see...

gatoWololo:
I see. It seems Git does correctly handle when multiple PRs include the same changes. It does make the PR a bit harder to review because it is unclear which are the "real" new changes and which changes we already reviewed in previous PRs. @ppenna thoughts?

ppenna:
@gatoWololo I think that we can overcome this issue by always opening Pull Requests *from* base branches... let me rebase...

gatoWololo (approved):

PR #52, opened 2022-03-24
[contrib] Updating Contrib

ppenna:
Description
========

In this commit I update the `runtime` crate so as to match the latest `dev` release in https://github.com/demikernel/runtime. As well, I've patched both mainline and test code accordingly.

Overall, this bumps scatter-gather allocation/free to fail instead of panicking.

Related Issues
===========

- https://github.com/demikernel/runtime/issues/12

### DIFF
--- a/src/catnip/runtime/memory/manager.rs
+++ b/src/catnip/runtime/memory/manager.rs
@@ -248,7 +248,7 @@ impl MemoryManager {
         {
             let mbuf_ptr = unsafe { rte_pktmbuf_alloc(self.inner.body_pool) };
             if mbuf_ptr.is_null() {
-                return Err(Fail::new(ENOMEM, "cannot allocate more mbufs"));
+                return Err(Fail::new(ENOMEM, "mbuf allocation failed"));
             }
             unsafe {
                 let num_bytes = (*mbuf_ptr).buf_len - (*mbuf_ptr).data_off;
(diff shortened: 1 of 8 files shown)


src/catnip/runtime/memory/manager.rs
```diff
@@ -238,14 +240,16 @@ impl MemoryManager {
...
+    pub fn alloc_sgarray(&self, size: usize) -> Result<dmtr_sgarray_t, Fail> {
         assert!(size <= self.inner.config.get_max_body_size());
 
         let sgaseg = if self.inner.config.get_inline_body_size() < size
             && size <= self.inner.config.get_max_body_size()
         {
             let mbuf_ptr = unsafe { rte_pktmbuf_alloc(self.inner.body_pool) };
-            assert!(!mbuf_ptr.is_null());
+            if mbuf_ptr.is_null() {
+                return Err(Fail::new(ENOMEM, "cannot allocate more mbufs"));
```
BrianZill, line 251: I'd make this text be "mbuf allocation failed".  All we know is that this one attempt to allocate a mbuf failed at this point in time.  The next attempt might succeed.

ppenna: Sure.

BrianZill (commented):
Looks good.

Issue #54, opened 2022-03-28
[demikernel] Mismatch `OperationResult` Structure Across LibOSes


ppenna:
Description
========

Currently, the `OperationResult` structure has a different layout in `catnip`, `catpowder` and `catnap`.

This makes applications non-uniform across these different libOSes.

Catnap
--------

```
pub enum OperationResult {
    Connect,
    Accept(RawFd),
    Push,
    Pop(Option<(Ipv4Addr, Port16)>, Bytes),
    Failed(Fail),
}
```
Reference: https://github.com/demikernel/demikernel/blob/dev/src/catnap/futures/mod.rs#L54

Catnip and Catpowder
-------------------------

```
pub enum OperationResult<T: Buffer> {
    Connect,
    Accept(QDesc),
    Push,
    Pop(Option<Ipv4Endpoint>, T),
    Failed(Fail),
}
```

Reference: https://github.com/demikernel/catnip/blob/dev/src/operations.rs#L16

ppenna:
Issue fixed in commit 515f989028bbc18e51d35f7b58bf23efe422f21a.

PR #55, opened 2022-03-28
[demikernel] Uniform `OperationResult` Across LibOSes

ppenna:
Description
========

In this Pull Request, I fixed issue https://github.com/demikernel/demikernel/issues/54.

### DIFF
--- a/src/catnap/futures/mod.rs
+++ b/src/catnap/futures/mod.rs
@@ -22,6 +22,7 @@ use self::{
     push::PushFuture,
     pushto::PushtoFuture,
 };
+use crate::Ipv4Endpoint;
 use ::catwalk::{
     FutureResult,
     SchedulerFuture,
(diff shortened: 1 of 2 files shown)


src/catnap/futures/mod.rs
```diff
@@ -55,7 +52,7 @@ pub enum OperationResult {
     Connect,
     Accept(RawFd),
     Push,
-    Pop(Option<(Ipv4Addr, Port16)>, Bytes),
+    Pop(Option<Ipv4Endpoint>, Bytes),
```
BrianZill, line 55: Glad you chose to standardize on the Endpoint option versus the Address, Port combo.

But looking to the future, would it be possible to standardize on a more generic IpEndpoint here, so we don't have to change this again to add IPv6 support?

ppenna: Yeah, it would be better to have it based in a generic `IpEndpoint` structure. Unfortunately, catnip was built on top of the `Ip´v4Endpoint` structure and we should first change this.

I've filed an issue for this: https://github.com/demikernel/catnip/issues/new

BrianZill: Great, thanks.

Issue #53, opened 2022-03-28
[demikernel] Expose C Bindings for `getaddrinfo()`


iyzhang:
Description
========

We should support `getaddrinfo()` in Demikernel.

ppenna:
As discussed in the core meeting, we don't need to provide `getaddrinfo()`. One can rely on the Linux system call for this.

PR #56, opened 2022-03-30
[demikernel] Support for `getaddrinfo()` and `freeaddrinfo()`

ppenna:
Description
========

In this PR I introduce support in Catnap for `getaddrinfo()` and `freeaddrinfo()`.

Related Issues
==========

- https://github.com/demikernel/demikernel/issues/53


ppenna:
As discussed in the core meeting, we don't need to provide `getaddrinfo()`. One can rely on the Linux system call for this.

Issue #58, opened 2022-03-30
[catnap] Bad OperationResult for `accept()`


ppenna:
Description
========

In Catnap, the `OperationResult` for `accept()` is wrong.

Currently, this constructed from the file descriptor supplied to `accept()`. Instead, the `OperationResult` structure should be constructed from a `QToken` that is associated to the underlying file descriptor that was allocated on `accept()`.

See code snippets bellow  for more info.

[catnap::futures::OperationResult](https://github.com/demikernel/demikernel/blob/1924bd04ceb357483c1334213f36f38fac445cfd/src/catnap/futures/mod.rs#L51)
```

/// Operation Result
pub enum OperationResult {
    Connect,
    Accept(RawFd),
    Push,
    Pop(Option<Ipv4Endpoint>, Bytes),
    Failed(Fail),
}
```

[catnap::futures::OperationResult::get_result()](https://github.com/demikernel/demikernel/blob/1924bd04ceb357483c1334213f36f38fac445cfd/src/catnap/futures/mod.rs#L82)
```
// Accept operation.
Operation::Accept(FutureResult {
    future,
    done: Some(Ok(fd)),
}) => (future.get_qd(), OperationResult::Accept(fd)),
Operation::Accept(FutureResult {
    future,
    done: Some(Err(e)),
 }) => (future.get_qd(), OperationResult::Failed(e)),
```

ppenna:
This bug was fixed in commig 364fef2ea67f6d39bc4881bc6ca7b05a74f725aa.

PR #59, opened 2022-03-30
[catnap] Bad `OperationResult` for `accept()`

ppenna:
Description
========

This pull request fixes issue https://github.com/demikernel/demikernel/issues/58.

### DIFF
--- a/src/catnap/futures/accept.rs
+++ b/src/catnap/futures/accept.rs
@@ -33,6 +33,8 @@ pub struct AcceptFuture {
     qd: QDesc,
     /// Underlying file descriptor.
     fd: RawFd,
+    /// Queue descriptor of incoming connection.
+    new_qd: QDesc,
 }
 
 //==============================================================================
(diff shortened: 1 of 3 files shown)


src/catnap/futures/accept.rs
```diff
@@ -33,6 +33,8 @@ pub struct AcceptFuture {
     qd: QDesc,
     /// Underlying file descriptor.
     fd: RawFd,
+    /// Queue descriptor of incoming connection
```
BrianZill, line 36: This comment is missing a period.

ppenna: Sure.

src/catnap/mod.rs
```diff
@@ -413,7 +414,23 @@ impl CatnapLibOS {
...
-        boxed_concrete_type.get_result()
+
+        let (qd, new_qd, new_fd, qr): (QDesc, Option<QDesc>, Option<RawFd>, OperationResult) =
+            boxed_concrete_type.get_result();
+
+        // Handle accept operation.
+        if let Some(new_qd) = new_qd {
+            // Associate raw file descriptor with queue descriptor.
+            if let Some(new_fd) = new_fd {
+                assert_eq!(self.sockets.insert(new_qd, new_fd).is_none(), true);
```
BrianZill, line 425: Shouldn't this be "assert!(self.sockets.insert(new_qd, new_fd).is_none());" ?

src/catnap/mod.rs
```diff
@@ -413,7 +414,23 @@ impl CatnapLibOS {
...
+            }
+            // Release entry in queue table.
```
BrianZill, line 427: This is a cleanup after some failure?

src/catnap/mod.rs
ppenna: Both of these asserts would be fine, but I'm happy to change it to the way you've suggested.

src/catnap/mod.rs
ppenna: Yes it is. I'm gonna move the comment to the line bellow.

BrianZill (approved):
Looks good now.  :-)

src/catnap/mod.rs
BrianZill: Yes, it was just a recommended style thing.

BrianZill (approved):

PR #60, opened 2022-03-30
[catnap] Dropping Deprecated Functions

ppenna:
Description
========

In this PR I drop deprecated functions in Demikernel. 

The exposed C interface will be fixed in another PR.


BrianZill (approved):
This PR is just dropping some deprecated API functionality and looks fine.  I'll approve it from a code standpoint.

I'm not aware of the reasoning around why this functionality was deprecated, however, I'm assuming Irene was involved/aware of this.

PR #63, opened 2022-04-01
[demikernel] Moving In SGA System Calls

ppenna:
Description
========

This PR is a follow up of https://github.com/demikernel/catnip/pull/136.

Note that pointers to the network stack were update, as well as there is some re-architecting required in order to merge similar logic.


BrianZill (approved):
I'm not sure I fully understand the purpose of these changes, but the code looks fine.

Issue #65, opened 2022-04-05
[build] Malformed Paths When Multiple Locations Are Found


ppenna:
Description
========

The auto-search mechanism of our Makefile malforms the `LD_LIBRARY_PATH` and the `PKG_CONFIG_PATH` if multiple matching locations exist.

For instance, in the example bellow, the `LD_LIBRARY_PATH` environment ends up with a whitespace character, because two matching locations were found:

![image](https://user-images.githubusercontent.com/4939789/161820491-3f937958-0fa6-489c-9334-d6fcaf7e7e66.png)

ppenna:
Issue fixed in commit 0f1a8bf696129c92da81be2cf7b4f402b7c92a72.

PR #66, opened 2022-04-05
[build] Malformed Paths When Multiple Locations Are Found

ppenna:
Description
========

In this PR I fix issue https://github.com/demikernel/demikernel/issues/65.


BrianZill (approved):
Yuck.  But okay.  :-)

Issue #67, opened 2022-04-05
[catnap] `push2()` Is Not Implemented


ppenna:
Description
========

Catnap is missing an implementation for `push2()`.

ppenna:
This bug was fixed in commit fbf1a572680e448d91223c2074c73f4ca3b09985.

Issue #64, opened 2022-04-05
[catnap] wait_any is not implemented


iyzhang:
wait_any is not implemented in catnap, only wait_any2; however, there are no C bindings for wait_any2, only wait_any. Need one of these to make more progress on Redis porting.

ppenna:
This bug was fixed in commit fbf1a572680e448d91223c2074c73f4ca3b09985

PR #68, opened 2022-04-05
[catnap] Implementing Missing System Calls

ppenna:
Description
========

In this PR I fix the following issues:
- https://github.com/demikernel/demikernel/issues/64
- https://github.com/demikernel/demikernel/issues/67

### DIFF
--- a/src/catnap/mod.rs
+++ b/src/catnap/mod.rs
@@ -364,8 +364,30 @@ impl CatnapLibOS {
     }
 
     /// Waits for any operation to complete.
-    pub fn wait_any(&mut self, _qts: &[QToken]) -> (usize, dmtr_qresult_t) {
-        todo!() // TODO: implement this function.
+    pub fn wait_any(&mut self, qts: &[QToken]) -> (usize, dmtr_qresult_t) {
+        #[cfg(feature = "profiler")]
+        timer!("catnap::wait_any");
+        trace!("wait_any(): qts={:?}", qts);
+
+        loop {
+            // Poll first, so as to give a chance of pending operations to complete.
+            self.runtime.poll();
+
+            // Search for any operation that has completed.
+            for (i, &qt) in qts.iter().enumerate() {
+                let mut handle: SchedulerHandle = self.runtime.get_handle(qt.into()).unwrap();
+
+                // Found one, so extract the result and return.
+                if handle.has_completed() {
+                    let (qd, r): (QDesc, OperationResult) = self.take_result(handle);
+                    return (i, pack_result(self.rt(), r, qd, qt.into()));
+                }
+
+                // Return this operation to the scheduling queue by removing the associated key
+                // (which would otherwise cause the operation to be freed).
+                handle.take_key();
+            }
+        }
     }
 
     /// Waits for any operation to complete.
--- a/src/catnap/mod.rs
+++ b/src/catnap/mod.rs
@@ -228,6 +228,18 @@ impl CatnapLibOS {
         }
     }
 
+    // Handles a push operation.
+    fn do_push(&mut self, qd: QDesc, buf: Bytes) -> Result<QToken, Fail> {
+        match self.sockets.get(&qd) {
+            Some(&fd) => {
+                let future: Operation = Operation::from(PushFuture::new(qd, fd, buf));
+                let handle: SchedulerHandle = self.runtime.schedule(future);
+                Ok(handle.into_raw().into())
+            },
+            _ => Err(Fail::new(EBADF, "invalid queue descriptor")),
+        }
+    }
+
     /// Pushes a scatter-gather array to a socket.
     pub fn push(&mut self, qd: QDesc, sga: &dmtr_sgarray_t) -> Result<QToken, Fail> {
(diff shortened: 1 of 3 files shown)


src/demikernel/libos.rs
```diff
@@ -52,18 +52,21 @@ pub enum LibOS {
 impl LibOS {
     cfg_if::cfg_if! {
         if #[cfg(feature = "catnip-libos")] {
+            /// Waits a pending operation in a I/O queue.
```
BrianZill, line 55: This sentence isn't grammatically correct.  Do you mean "Waits for a pending operation" or "Waits on a pending operation"?  This same comment is repeated 5 times below.

src/demikernel/libos.rs
```diff
@@ -159,6 +180,7 @@ impl LibOS {
         }
     }
 
+    /// Pushes a raw data to a UDP socket.
```
BrianZill, line 183: Delete the extra "a": "Pushes raw data"...

src/catnap/mod.rs
```diff
@@ -346,8 +364,29 @@ impl CatnapLibOS {
...
+        timer!("catnap::wait_any");
+        trace!("wait_any(): qts={:?}", qts);
+
+        loop {
+            // Poll first, so as to give a chance of pending operations to complete.
+            self.runtime.poll();
+
+            // Search for any operation that has completed.
+            for (i, &qt) in qts.iter().enumerate() {
+                let handle: SchedulerHandle = self.runtime.get_handle(qt.into()).unwrap();
```
BrianZill, line 378: Is this unwrap() guaranteed to never fail?

src/catnap/mod.rs
```diff
@@ -346,8 +364,29 @@ impl CatnapLibOS {
...
+
+                // Found one, so extract the result and return.
+                if handle.has_completed() {
+                    let (qd, r): (QDesc, OperationResult) = self.take_result(handle);
+                    return (i, pack_result(self.rt(), r, qd, qt.into()));
+                }
+
+                // Put back operation in the scheduling queue.
+                handle.into_raw();
```
BrianZill, line 387: What is this doing?  If the idea is to remove the key from the SchedulerHandle, "handle.take_key()" would be clearer.  The comment doesn't really explain what is happening either.

src/demikernel/libos.rs
ppenna: Sure

src/demikernel/libos.rs
ppenna: I mean "waits on". Thanks.

src/catnap/mod.rs
ppenna: That's a catch and it is indeed a bug.  We cannot assume that the supplied `qtoken` is valid at all. I've created issue https://github.com/demikernel/demikernel/issues/71 to keep track of this.

src/catnap/mod.rs
ppenna: Sure, I'll change it to `handle.take_key()`, but the bad semantics that you have noticed should be fixed in the scheduler, when a better reference counting mechanism is implemented and the scheduling system is re-architected. For now,  I have just decoupled it and made it more to work on abstract futures.

Overall, the use of the `SchedulerHandle` is overload and it is used to refer to both, short-living and long-living operations. In the latter case the reference count in the underlying `WakerPageRef` is bumped, and in the former case the reference count is not.

To summary up, this line is doing what the comment says: putting back the underlying operation in the scheduling queue (instead of dropping it). What is going on is the following:

- When we call `get_handle()` the scheduler performs a lookup in its internal tables and gets a unique identifier for the given operation. When it does so, we create a `SchedulerHandle` from a given `WakerPageRef`, but we don't bump the associated reference count.
- When the `SchedulerHandle` is dropped, it would call the custom `Drop` trait implementation. This would decrement the reference counting of the underlying `WakerPageRef`, if `key` is not a `None` value. 
- If the target `ScheduleHandle` is a short-living one, the drop will happen afterwards, when the associated future completes. Thus, to avoid a double-free scenario, we take out the underlying `key` in the `ScheduleHandle`, leaving a `None` on its place and preventing `drop()` to run.

This solution is ugly, the exposed interface is not clear and I'm not happy with it.

For now, I've created  issue https://github.com/demikernel/catwalk/issues/2 to keep track of this.

Let me know if you have any suggestions on the comment to make the whole mess clearer.

src/catnap/mod.rs
BrianZill: Great, thanks.

src/catnap/mod.rs
BrianZill: Okay, I see.  Yes, that is ugly.  I would change the comment to say "Return this operation to the scheduling queue by removing the associated key (which would otherwise cause the operation to be freed)" or something like that.

I would also change the code to be just "handle.take_key();".  There's no need to "into_raw()" or "unwrap()" because you're not doing anything with the result.

ppenna: Sure.

BrianZill (approved):
Thanks for making those changes.

Issue #43, opened 2022-01-14
[test] Ping Pong Workflow for TCP


ppenna:
Description
========

We are missing a GitHub Action workflow for TCP that exercises ping-pong.

ppenna:
Feature introduced in commit f374986741b4d7e5bf8581d43882aa45d8e27419.

Issue #42, opened 2022-01-14
[test] Push Pop Workflow for TCP


ppenna:
Description
========

We are missing a GitHub Action workflow for TCP that exercises push-pop.

ppenna:
Feature introduced in commit f374986741b4d7e5bf8581d43882aa45d8e27419.

PR #75, opened 2022-04-07
[test] System Testing for TCP

ppenna:
Description
========

In this PR I introduce the following features:
- https://github.com/demikernel/demikernel/issues/42
- https://github.com/demikernel/demikernel/issues/43

### DIFF
--- a/tests/pingpong.rs
+++ b/tests/pingpong.rs
@@ -1,8 +1,6 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-#![feature(try_blocks)]
-
 mod common;
 
 //==============================================================================
--- a/tests/pingpong.rs
+++ b/tests/pingpong.rs
@@ -30,7 +30,7 @@ use ::std::{
 };
 
 //==============================================================================
-// Ping Pong
+// UDP Ping Pong
 //==============================================================================
 
 #[test]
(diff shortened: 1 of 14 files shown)


tests/pingpong.rs
```diff
@@ -136,3 +145,117 @@ fn udp_ping_pong() {
...
+    let nrounds: usize = 1024;
+    let local_addr: Ipv4Endpoint = test.local_addr();
+    let remote_addr: Ipv4Endpoint = test.remote_addr();
+    let expectbuf: Vec<u8> = test.mkbuf(fill_char);
+
+    // Setup peer.
+    let sockqd: QDesc = test
+        .libos
+        .socket(libc::AF_INET, libc::SOCK_STREAM, 0)
+        .unwrap();
```
BrianZill, line 166: I don't think we should ever be using unwrap().  The only exception would be if it is a logic error for the underlying routine to return the None or Err case instead of the Some or Ok case.  For test code running outside of a test framework, I think using expect() might be okay (if the error message was good).  But inside a test framework that is expecting explicit failures, we should check for error returns and explicitly fail so the test framework does the right thing.  And of course in production code we shouldn't use unwrap() or expect() as we need to always handle errors gracefully.

This comment applies to each instance of unwrap in both of these test files.

ppenna: Sure. These passed unoticed.

tests/pingpong.rs
```diff
@@ -13,29 +11,37 @@ use self::common::Test;
...
+    let fill_char: u8 = 'a' as u8;
     let local_addr: Ipv4Endpoint = test.local_addr();
     let remote_addr: Ipv4Endpoint = test.remote_addr();
 
     // Setup peer.
-    let sockfd = test
+    let sockfd: QDesc = test
         .libos
         .socket(libc::AF_INET, libc::SOCK_DGRAM, 0)
         .unwrap();
```
BrianZill, line 47: Still 4 of these left.

tests/pingpong.rs
```diff
@@ -136,3 +145,128 @@ fn udp_ping_pong() {
...
+            let recvbuf = match test.libos.wait2(qtoken) {
+                Ok((_, OperationResult::Pop(_, buf))) => buf,
+                Err(e) => panic!("operation failed: {:?}", e.cause),
+                _ => unreachable!(),
+            };
+
+            // Sanity check received data.
+            assert!(
+                Test::bufcmp(&expectbuf, recvbuf.clone()),
+                "server expectbuf != recevbuf"
```
BrianZill, line 208: Typo (spelling of recvbuf).  Same thing on line 264.

tests/pingpong.rs
BrianZill: Thanks for fixing the others though, looks much better now.

ppenna: Oh, sure... I didn't meant to touch the UDP testing code in this PR, that's why I haven't fixed. I'll do it.

tests/pingpong.rs
ppenna: It is fixed now.

tests/pingpong.rs
BrianZill: Thanks for fixing these.

tests/pingpong.rs
BrianZill: Thanks for fixing these, looks much better now.

BrianZill (approved):

Issue #71, opened 2022-04-06
[pdpix] `wait()` and `wait_any()` System Calls Should Return an Error


ppenna:
Description
========

We should review the implementation of `wait()`, `wait_any()` and all their existing flavors, and make them return a `Result<..>`.

These functions use the supplied `QToken` as a parameter to `get_handle()` in order to retrieve the associated `ScheduleHandle`. We are currently assuming that the supplied value is valid and we simply `unwrap()` the result.  Unfortunately, it is not guaranteed to never fail because the `QToken` may not be valid.

ppenna:
This bug was fixed in commit bb2811ddfdc6b86f415cb1ea5d226e6eb9dab7e3.

PR #77, opened 2022-04-08
[pdpix] wait() and wait_any() System Calls Should Return an Error

ppenna:
Description
========

In this PR, I fix issue https://github.com/demikernel/demikernel/issues/71.


BrianZill (approved):
Looks good!

Issue #51, opened 2022-03-24
[catnip] Bad DPDK-Managed Buffer Allocation/Release


ppenna:
Description
========

Pointer arithmetic used when freeing DPDK-managed buffers is wrong, causing dpdk memory pool to be corrupted in the very first allocation/deallocation.

This bug is exposed in the [enhancement-tests-memory](https://github.com/demikernel/demikernel/tree/enhancement-tests-memory) branch, in the following test unit:

```
fn test_unit_sga_alloc_free_loop_tight() {
    let size: usize = 4096;
    let libos: LibOS = LibOS::new();

    // Allocate and deallocate several times.
    for _ in 0..1_000_000 {
        let sga: dmtr_sgarray_t = libos.sgaalloc(size).unwrap();
        libos.sgafree(sga).unwrap();
    }
}
```

Even if we allocate and deallocate buffers one after another, we soon run out of `mbufs`:

```
test test_unit_sga_alloc_free_loop_tight ... EAL: Detected 8 lcore(s)
EAL: Detected 1 NUMA nodes
EAL: Auto-detected process type: PRIMARY
EAL: Detected shared linkage of DPDK
EAL: Multi-process socket /var/run/dpdk/rte/mp_socket
EAL: Selected IOVA mode 'PA'
EAL: No available 1048576 kB hugepages reported
EAL: Probing VFIO support...
EAL: VFIO support initialized
EAL: Probe PCI driver: mlx5_pci (15b3:1016) device: 9ec9:00:02.0 (socket 0)
mlx5_pci: Failed to allocate Tx DevX UAR (BF)
mlx5_pci: Failed to allocate Rx DevX UAR (BF)
mlx5_pci: No available register for Sampler.
mlx5_pci: Size 0xFFFF is not power of 2, will be aligned to 0x10000.
net_vdev_netvsc: probably using routed NetVSC interface "eth1" (index 3)
tap_nl_dump_ext_ack(): Cannot delete qdisc with handle of zero
tap_nl_dump_ext_ack(): Failed to find qdisc with specified classid
tap_nl_dump_ext_ack(): Failed to find qdisc with specified classid
tap_nl_dump_ext_ack(): Failed to find qdisc with specified classid
tap_nl_dump_ext_ack(): Failed to find qdisc with specified classid
tap_nl_dump_ext_ack(): Failed to find qdisc with specified classid
tap_nl_dump_ext_ack(): Failed to find qdisc with specified classid
[...]

PR #82, opened 2022-04-11
[catnip] Bad Scatter-Gather Allocation/Free

ppenna:
Description
========

This PR fixes issue https://github.com/demikernel/demikernel/issues/51.

In summary, the pointer arithmetic used for releasing allocated `mbufs` is bad. Instead of relying on this technique, I simply changed current code to use the returned `mbuf` pointer.

Since the management of DPDK buffers is not well design, I've also placed a warning, so that we know when we are relying on this allocation strategy.

Finally, I introduce unit-tests that ensure that the current logic works as expected.

This component will be strongly re-architected in a next PR.

### DIFF
--- a/src/catnip/runtime/memory/manager.rs
+++ b/src/catnip/runtime/memory/manager.rs
@@ -294,15 +294,26 @@ impl MemoryManager {
         })
     }
 
+    /// Releases a scatter-gather array.
     pub fn free_sgarray(&self, sga: dmtr_sgarray_t) -> Result<(), Fail> {
-        assert_eq!(sga.sga_numsegs, 1);
-        let sgaseg = sga.sga_segs[0];
-        let (ptr, len) = (sgaseg.sgaseg_buf, sgaseg.sgaseg_len as usize);
+        // Bad scatter-gather.
+        // TODO: Drop this check once we support scatter-gather arrays with multiple segments.
+        if sga.sga_numsegs != 1 {
+            return Err(Fail::new(
+                libc::EINVAL,
+                "scatter-gather array with invalid size",
+            ));
+        }
 
-        if self.is_body_ptr(ptr) {
-            let mbuf_ptr = self.recover_body_mbuf(ptr).expect("Invalid sga pointer");
+        // Release underlying buffer.
+        if !sga.sga_buf.is_null() {
+            // Release DPDK-managed buffer.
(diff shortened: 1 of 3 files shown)


src/catnip/runtime/memory/manager.rs
```diff
@@ -297,17 +257,36 @@ impl MemoryManager {
...
+        let (ptr, len): (*mut c_void, usize) = (sgaseg.sgaseg_buf, sgaseg.sgaseg_len as usize);
+
+        // Clone underlying buffer.
+        let buf: DPDKBuf = if !sga.sga_buf.is_null() {
+            // Clone DPDK-managed buffer.
+            let mbuf_ptr: *mut rte_mbuf = sga.sga_buf as *mut rte_mbuf;
+            let body_clone: *mut rte_mbuf = self.inner.clone_mbuf(mbuf_ptr);
+            let mut mbuf: Mbuf = Mbuf::new(body_clone, self.clone());
+            // Adjust buffer length.
+            // TODO: Replace the following method for computing the length of a mbuf once we have a propor Mbuf abstraction.
```
BrianZill, line 281: Spelling: propor -> proper

tests/sga.rs
```diff
@@ -0,0 +1,84 @@
...
+
+use ::demikernel::LibOS;
+use ::runtime::types::dmtr_sgarray_t;
+
+//==============================================================================
+// Constants
+//==============================================================================
+
+/// Size of scatter-gather arrays.
+const SGA_MAX_SIZE: usize = 1280;
```
BrianZill, line 16: Might we want to test allocating different sizes, instead of always the same size?

src/catnip/runtime/memory/manager.rs
ppenna: Sure. I've fixed this.

tests/sga.rs
ppenna: Nice! I've added tests for 64-byte SGA arrays, to cover the heap-allocated buffers in catnip. Also, as we've discussed, I'll add even more comprehensive tests after I'm done with the re-architecting of this memory management component.

BrianZill: This is great, thanks!

BrianZill (approved):
Looks good now.

Issue #83, opened 2022-04-11
[catnip] Introduce Abstraction for DPDK Memory Pool


ppenna:
Description
========

In Catnip LibOS we are currently playing with memory pools using raw pointers to DPDK-managed structures and unsafe code everywhere.

This not only make code harder to follow but increase our chances of having bugs.

Proposed Solution
=============

We should create a `MemoryPool` abstraction that encompasses all operations on a DPDK memory pool.

ppenna:
This enhancement was introduced in commit 39346ccbcb14872e1b2492dfc105418203257275.

PR #84, opened 2022-04-12
[catnip] Memory Pool Abstraction

ppenna:
Description
========

In this PR, I mostly introduce the enhancement that is described in https://github.com/demikernel/demikernel/issues/83. 

Here is a digest of the commits:
- In commit 2a31472a5c8bcd6fb8af00cc75be92f6d69ae14a I drop bad checks that were used in the pointer arithmetic for allocating/releasing `mbufs`.
- In commit dc73ffee6c1d4c56d8780a5e4eff2b56c9d0fbc5 I introduce the `MemoryPool` abstraction.
- In commit c31d48bb67b2039986c1790d0012d9c9d785ecd9 and 729f0915277ae2e27082cf763ad1c419e76f1fd2 I refactor the code to rely on the `MemoryPool` abstraction
- In commit 837ed5049129f8678ef2e8a6005e85c5936a8534 I drop the indirect memory pool, because it is not required to provide the shallow copy of `mbufs` ([`rte_pktmbuf_clone`](https://elixir.bootlin.com/dpdk/latest/source/lib/mbuf/rte_mbuf.c#L512) does the job) and it was causing interdependency between several modules.

### DIFF
--- a/src/catnip/runtime/memory/manager.rs
+++ b/src/catnip/runtime/memory/manager.rs
@@ -55,10 +55,6 @@ pub struct Inner {
     // internally within the network stack.
     header_pool: Rc<MemoryPool>,
 
-    // Used by networking stack for cloning buffers passed in from the application. There buffers
-    // are also only used internally within the networking stack.
-    indirect_pool: Rc<MemoryPool>,
-
     // Large body pool for buffers given to the application for zero-copy.
     body_pool: Rc<MemoryPool>,
 }
@@ -75,7 +71,7 @@ pub struct MemoryManager {
 impl MemoryManager {
     pub fn new(max_body_size: usize) -> Result<Self, Error> {
         let memory_config: MemoryConfig =
-            MemoryConfig::new(None, None, None, Some(max_body_size), None, None);
+            MemoryConfig::new(None, None, Some(max_body_size), None, None);
 
         Ok(Self {
             inner: Rc::new(Inner::new(memory_config)?),
@@ -263,16 +259,6 @@ impl Inner {
             config.get_cache_size(),
         )?;
 
-        // Create memory pool for holding indirect buffers.
-        let indirect_pool: MemoryPool = MemoryPool::new(
-            CString::new("indirect_pool")?,
-            // These mbufs have no body -- they're just for indirect mbufs to point to
-            // allocations from the body pool.
-            0,
-            config.get_indirect_pool_size(),
-            config.get_cache_size(),
-        )?;
-
         // Create memory pool for holding packet bodies.
         let body_pool: MemoryPool = MemoryPool::new(
             CString::new("body_pool")?,
@@ -284,7 +270,6 @@ impl Inner {
         Ok(Self {
             config,
             header_pool: Rc::new(header_pool),
-            indirect_pool: Rc::new(indirect_pool),
             body_pool: Rc::new(body_pool),
         })
     }
--- a/src/catnip/runtime/memory/manager.rs
+++ b/src/catnip/runtime/memory/manager.rs
@@ -65,7 +65,7 @@ pub struct Inner {
 
 #[derive(Clone, Debug)]
 pub struct MemoryManager {
-    pub inner: Rc<Inner>,
+    inner: Rc<Inner>,
 }
 
 //==============================================================================
@@ -80,10 +80,6 @@ impl MemoryManager {
         Ok(Self {
             inner: Rc::new(Inner::new(memory_config)?),
         })
-    }
-
-    pub fn make_buffer(&self, packet: *mut rte_mbuf) -> DPDKBuf {
-        DPDKBuf::Managed(Mbuf::new(packet))
     }
 
     /// Converts a runtime buffer into a scatter-gather array.
@@ -291,19 +287,5 @@ impl Inner {
             indirect_pool: Rc::new(indirect_pool),
             body_pool: Rc::new(body_pool),
         })
-    }
-
-    pub fn alloc_indirect_empty(&self) -> *mut rte_mbuf {
-        match self.indirect_pool.alloc_mbuf(None) {
-            Ok(mbuf_ptr) => mbuf_ptr,
-            Err(e) => panic!("failed to allocate mbuf: {:?}", e.cause),
-        }
-    }
-
-    pub fn clone_mbuf(&self, ptr: *mut rte_mbuf) -> *mut rte_mbuf {
-        match MemoryPool::clone_mbuf(ptr) {
-            Ok(mbuf_ptr) => mbuf_ptr,
-            Err(e) => panic!("failed to clone mbuf: {:?}", e.cause),
-        }
     }
 }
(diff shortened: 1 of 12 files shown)


BrianZill:
Is the title meant to say "[canip]" or "[catnip]"?

src/catnip/runtime/memory/mbuf.rs
```diff
@@ -114,9 +116,7 @@ impl Drop for Mbuf {
         if self.ptr.is_null() {
             return;
         }
-        unsafe {
-            rte_pktmbuf_free(self.ptr);
-        }
+        MemoryPool::free_mbuf(self.ptr);
         self.ptr = ptr::null_mut();
```
BrianZill, line 120: This seems very un-Rustian.  Now that the unsafe code is contained inside MemoryPool::free_mbuf(), it seems strange to be nullifying self.ptr here.  I would expect free_mbuf() to be doing this for us, so we don't have access to a bogus pointer upon return.

src/catnip/runtime/memory/mempool.rs
```diff
@@ -0,0 +1,120 @@
...
+        unsafe {
+            // Do it.
+            rte_pktmbuf_free(mbuf_ptr);
+        }
+    }
+
+    /// Clones a mbuf into a memory pool.
+    pub fn clone_mbuf(mbuf_ptr: *mut rte_mbuf) -> Result<*mut rte_mbuf, Fail> {
+        unsafe {
+            // Do it.
```
BrianZill, line 110: A comment like this isn't really helpful.  If you're going to have a comment here, it should provide some information about what is happening.

src/catnip/runtime/memory/mempool.rs
```diff
@@ -0,0 +1,120 @@
...
+        if mbuf_ptr.is_null() {
+            return Err(Fail::new(libc::ENOMEM, "cannot allocate more mbufs"));
+        }
+
+        // Fill out some fields of the underlying mbuf.
+        unsafe {
+            let mut num_bytes: u16 = (*mbuf_ptr).buf_len - (*mbuf_ptr).data_off;
+
+            if let Some(size) = size {
+                // Allocated buffer is not  big enough.
```
BrianZill, line 84: Comments about an if that come before the if should describe what it is checking, not one of the branches.  Descriptions of one of the branches should go inside that branch of the if.

For example:
// Check whether x is big enough.
if x > 5 {
    // x is big enough to do something.
    do_something();
} else {
    // x is too small.
    do_something_else();
}

To me, what you have on line 84 is describing the result of the if on line 83.  But you actually mean to describe the result of the if on line 85.  I'd combine the line 84 comment with the comment on line 86, and say "The allocated buffer is not big enough, rollback the allocation."  If you want a comment on line 84, it should say something like "Check if the allocated buffer is big enough."

BrianZill: Basically, the comment should pertain to the code block in which it resides.

src/catnip/runtime/memory/mbuf.rs
ppenna: Somehow we should make the raw pointer null, because it is not managed by Rust. I did not get what you are suggesting. Could you clarify it further?

src/catnip/runtime/memory/mempool.rs
ppenna: Sure. I'll make it like you suggested.

src/catnip/runtime/memory/mempool.rs
```diff
@@ -0,0 +1,120 @@
...
+use ::std::ffi::CString;
+
+//==============================================================================
+// Structures
+//==============================================================================
+
+/// DPDK Memory Pool
+#[derive(Debug)]
+pub struct MemoryPool {
+    /// Underlying memory pool.
```
BrianZill, line 27: Did you intend to use a [documentation comment](https://doc.rust-lang.org/reference/comments.html) here?

ppenna: Yes. Any suggestion?

src/catnip/runtime/memory/mempool.rs
```diff
@@ -0,0 +1,120 @@
...
+    pub fn into_raw(&self) -> *mut rte_mempool {
+        self.pool
+    }
+
+    /// Allocates a mbuf in the target memory pool.
+    pub fn alloc_mbuf(&self, size: Option<usize>) -> Result<*mut rte_mbuf, Fail> {
+        // TODO: Drop the following warning once DPDK memory management is more stable.
+        warn!("allocating mbuf from DPDK pool");
+
+        // Allocated mbuf.
```
BrianZill, line 73: Just "Allocate mbuf."

src/catnip/runtime/memory/mempool.rs
BrianZill: No, just wondering since we haven't been consistent about using documentation comments throughout the codebase.

src/catnip/runtime/memory/mempool.rs
BrianZill: There's a similar comment on line 102.

src/catnip/runtime/memory/manager.rs
```diff
@@ -294,136 +247,41 @@ impl MemoryManager {
...
-        let mut sz: rte_mempool_objsz = unsafe { mem::zeroed() };
-        let elt_size = mem::size_of::<rte_mbuf>() + priv_size as usize + config.get_max_body_size();
-        let flags = 0;
-        unsafe {
-            rte_mempool_calc_obj_size(elt_size as u32, flags, &mut sz as *mut _);
-        }
-        assert_eq!(sz.header_size, 64);
-        assert_eq!(sz.elt_size as usize, 128 + config.get_max_body_size());
-        assert_eq!(sz.trailer_size, 0);
+        // TODO: The following computation for header size is bad. It should be fixed to maximum possible size.
```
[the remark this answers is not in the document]
ppenna: Sure: https://github.com/demikernel/demikernel/issues/85

src/catnip/runtime/memory/mempool.rs
ppenna: Every file that I touch I somehow try to provide documentation comments for public API and structures. Eventually we get our documentation in a better shape.

src/catnip/runtime/memory/mempool.rs
ppenna: Fixed. Thanks.

src/catnip/runtime/memory/mbuf.rs
BrianZill: I was suggesting that free_mbuf() null out self.ptr itself.  To really do this right, we (or rather, the MemoryPool) would hide the "*mut rte_mbuf" inside a Rust type of our own creation.  We'd pass that type into free_mbuf(), it would take ownership, etc.

But don't worry about it, this is probably good enough for now.

BrianZill (approved):
Okay.

src/catnip/runtime/memory/mbuf.rs
ppenna: Oh, I got it and agree. I'm refactoring `Mbuf` so as to make this interface cleaner too. Thanks.

Issue #86, opened 2022-04-13
[memory] Introduce a Better Abstraction for Runtime Buffers


ppenna:
Description
========

We currently have two abstractions for heap-managed runtime buffers: `Bytes` and `BytesMut`.

We should have a single abstraction and keep the code base simple.

BrianZill:
Regarding the description, this abstraction isn't just for _heap-managed_ buffers, correct?

ppenna:
@BrianZill 

> Regarding the description, this abstraction isn't just for _heap-managed_ buffers, correct?

Both `Bytes` and `BytesMut` concern abstractions for runtime buffers that were allocated in the heap. The new `HeapBuffer` abstraction that I propose in PR #87 is a replacement for them.

ppenna:
This enhancement was introduced in commit 8630b7faebca320f0b6056049f766cd9a0c6cd0b.

PR #87, opened 2022-04-13
[memory] Introduce a Better Abstraction for Runtime Buffers

ppenna:
Description
========

In this PR I fix issue https://github.com/demikernel/demikernel/issues/86.

Here is a digest of the commits:
- In commit aab28885ea2fad42ded151b3bef0a461c3ee7695 I introduce the new `HeapBuffer` abstraction.
- In commits ee0efe081b96594e29036ced6b4ccb0457964d7e, 1b024fbaf5f8a441eacd327f22bdb1b2329170ab,  I patch our LibOSes to rely on this new abstraction.

Also, note that:
- In commit 1c027eb81b24315f83a56ff27713cc2a33f130d8 I fix a bug in our TCP push pop test. It was expecting a packet of a given content and size. This check was not good because this unit tests performns a one-directrion transmission.

### DIFF
--- a/src/catnap/runtime.rs
+++ b/src/catnap/runtime.rs
@@ -5,6 +5,7 @@
 // Imports
 //==============================================================================
 
+use crate::demikernel::dbuf::DataBuffer;
 use ::arrayvec::ArrayVec;
 use ::catwalk::{
     Scheduler,
@@ -19,8 +20,7 @@ use ::rand::{
 use ::runtime::{
     fail::Fail,
     memory::{
-        Bytes,
-        BytesMut,
+        Buffer,
         MemoryRuntime,
     },
     network::{
@@ -89,15 +89,15 @@ impl PosixRuntime {
 /// Memory Runtime Trait Implementation for POSIX Runtime
 impl MemoryRuntime for PosixRuntime {
     /// Memory Buffer
-    type Buf = Bytes;
+    type Buf = DataBuffer;
 
-    /// Creates a [dmtr_sgarray_t] from a memory buffer.
-    fn into_sgarray(&self, buf: Bytes) -> Result<dmtr_sgarray_t, Fail> {
-        let buf_copy: Box<[u8]> = (&buf[..]).into();
-        let ptr: *mut [u8] = Box::into_raw(buf_copy);
+    /// Converts a runtime buffer into a scatter-gather array.
+    fn into_sgarray(&self, dbuf: DataBuffer) -> Result<dmtr_sgarray_t, Fail> {
+        let len: usize = dbuf.len();
+        let dbuf_ptr: *const [u8] = DataBuffer::into_raw(dbuf)?;
         let sgaseg: dmtr_sgaseg_t = dmtr_sgaseg_t {
-            sgaseg_buf: ptr as *mut c_void,
-            sgaseg_len: buf.len() as u32,
+            sgaseg_buf: dbuf_ptr as *mut c_void,
+            sgaseg_len: len as u32,
         };
         Ok(dmtr_sgarray_t {
             sga_buf: ptr::null_mut(),
@@ -107,12 +107,13 @@ impl MemoryRuntime for PosixRuntime {
         })
     }
 
-    /// Allocates a [dmtr_sgarray_t].
+    /// Allocates a scatter-gather array.
     fn alloc_sgarray(&self, size: usize) -> Result<dmtr_sgarray_t, Fail> {
-        let allocation: Box<[u8]> = unsafe { Box::new_uninit_slice(size).assume_init() };
-        let ptr: *mut [u8] = Box::into_raw(allocation);
-        let sgaseg = dmtr_sgaseg_t {
-            sgaseg_buf: ptr as *mut _,
+        // Allocate a heap-managed buffer.
+        let dbuf: DataBuffer = DataBuffer::new(size)?;
+        let dbuf_ptr: *const [u8] = DataBuffer::into_raw(dbuf)?;
+        let sgaseg: dmtr_sgaseg_t = dmtr_sgaseg_t {
+            sgaseg_buf: dbuf_ptr as *mut c_void,
             sgaseg_len: size as u32,
         };
         Ok(dmtr_sgarray_t {
@@ -123,40 +124,46 @@ impl MemoryRuntime for PosixRuntime {
         })
     }
 
-    /// Releases a [dmtr_sgarray_t].
+    /// Releases a scatter-gather array.
     fn free_sgarray(&self, sga: dmtr_sgarray_t) -> Result<(), Fail> {
-        assert_eq!(sga.sga_numsegs, 1);
-        for i in 0..sga.sga_numsegs as usize {
-            let seg: &dmtr_sgaseg_t = &sga.sga_segs[i];
-            let allocation: Box<[u8]> = unsafe {
-                Box::from_raw(slice::from_raw_parts_mut(
-                    seg.sgaseg_buf as *mut _,
-                    seg.sgaseg_len as usize,
-                ))
-            };
-            drop(allocation);
+        // Check arguments.
+        // TODO: Drop this check once we support scatter-gather arrays with multiple segments.
+        if sga.sga_numsegs != 1 {
+            return Err(Fail::new(
+                libc::EINVAL,
+                "scatter-gather array with invalid size",
+            ));
         }
+
+        // Release heap-managed buffer.
+        let sgaseg: dmtr_sgaseg_t = sga.sga_segs[0];
+        let (data_ptr, length): (*mut u8, usize) =
+            (sgaseg.sgaseg_buf as *mut u8, sgaseg.sgaseg_len as usize);
+
+        // Convert back raw slice to a heap buffer and drop allocation.
+        DataBuffer::from_raw_parts(data_ptr, length)?;
 
         Ok(())
     }
 
-    /// Clones a [dmtr_sgarray_t] into a memory buffer.
-    fn clone_sgarray(&self, sga: &dmtr_sgarray_t) -> Result<Bytes, Fail> {
-        let mut len: u32 = 0;
-        for i in 0..sga.sga_numsegs as usize {
-            len += sga.sga_segs[i].sgaseg_len;
+    /// Clones a scatter-gather array.
+    fn clone_sgarray(&self, sga: &dmtr_sgarray_t) -> Result<DataBuffer, Fail> {
+        // Check arguments.
+        // TODO: Drop this check once we support scatter-gather arrays with multiple segments.
+        if sga.sga_numsegs != 1 {
+            return Err(Fail::new(
+                libc::EINVAL,
+                "scatter-gather array with invalid size",
+            ));
         }
-        let mut buf: BytesMut = BytesMut::zeroed(len as usize).unwrap();
-        let mut pos: usize = 0;
-        for i in 0..sga.sga_numsegs as usize {
-            let seg: &dmtr_sgaseg_t = &sga.sga_segs[i];
-            let seg_slice = unsafe {
-                slice::from_raw_parts(seg.sgaseg_buf as *mut u8, seg.sgaseg_len as usize)
-            };
-            buf[pos..(pos + seg_slice.len())].copy_from_slice(seg_slice);
-            pos += seg_slice.len();
-        }
-        Ok(buf.freeze())
+
+        let sgaseg: dmtr_sgaseg_t = sga.sga_segs[0];
+        let (ptr, len): (*mut c_void, usize) = (sgaseg.sgaseg_buf, sgaseg.sgaseg_len as usize);
+
+        // Clone heap-managed buffer.
+        Ok(DataBuffer::from_slice(unsafe {
+            slice::from_raw_parts(ptr as *const u8, len)
+        }))
     }
 }
 
--- a/src/catnip/runtime/memory/manager.rs
+++ b/src/catnip/runtime/memory/manager.rs
@@ -6,6 +6,7 @@
 //==============================================================================
 
 use super::mempool::MemoryPool;
+use crate::demikernel::dbuf::DataBuffer;
 use ::anyhow::Error;
 use ::catnip::protocols::{
     ethernet2::ETHERNET2_HEADER_SIZE,
@@ -19,7 +20,7 @@ use ::dpdk_rs::{
 use ::libc::c_void;
 use ::runtime::{
     fail::Fail,
-    memory::BytesMut,
+    memory::Buffer,
     types::{
         dmtr_sgarray_t,
         dmtr_sgaseg_t,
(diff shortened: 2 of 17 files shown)


BrianZill:
A meta comment: I don't like the name HeapBuffer, because these aren't necessarily _heap_ buffers.  At least not in the way people generally speak of a program's heap.

I think this is supposed to be the equivalent of Window's NET_BUFFERs or Unix's Mbufs, but both of those names are also bad for various reasons.  I would probably call these just "DataBuffer".  Or we can have a separate discussion about what to call them.

ppenna:
@BrianZill 

> A meta comment: I don't like the name HeapBuffer, because these aren't necessarily _heap_ buffers. At least not in the way people generally speak of a program's heap.
> 
> I think this is supposed to be the equivalent of Window's NET_BUFFERs or Unix's Mbufs, but both of those names are also bad for various reasons. I would probably call these just "DataBuffer". Or we can have a separate discussion about what to call them.

Sure! Do you have a suggestion on what we should call this?

BrianZill:
My suggestion would be something along the lines of "DataBuffer", or if that is too generic, maybe "IoBuffer".  I'd avoid using "NetworkBuffer" or "NetBuffer" as we want Demikernel to be used for other types of I/O as well (eventually).  Maybe we could go with "MemoryBuffer" or "MemBuffer", but that seems even more generic than "DataBuffer" and might be too close to Berkeley Mbufs.

ppenna:
> My suggestion would be something along the lines of "DataBuffer", or if that is too generic, maybe "IoBuffer". I'd avoid using "NetworkBuffer" or "NetBuffer" as we want Demikernel to be used for other types of I/O as well (eventually). Maybe we could go with "MemoryBuffer" or "MemBuffer", but that seems even more generic than "DataBuffer" and might be too close to Berkeley Mbufs.

Sure @BrianZill . I did it.

src/catnap/runtime.rs
```diff
@@ -123,40 +124,45 @@ impl MemoryRuntime for PosixRuntime {
...
-            pos += seg_slice.len();
-        }
-        Ok(buf.freeze())
+
+        let sgaseg: dmtr_sgaseg_t = sga.sga_segs[0];
+        let (ptr, len): (*mut c_void, usize) = (sgaseg.sgaseg_buf, sgaseg.sgaseg_len as usize);
+
+        // Clone heap-managed buffer.
+        let seg_slice: &[u8] = unsafe { slice::from_raw_parts(ptr as *const u8, len) };
+        Ok(DataBuffer::from_slice(seg_slice))
```
BrianZill, line 165: Instead of constructing a slice here, and calling DataBuffer::from_slice(), could you just call DataBuffer::from_raw_parts() directly?  Or would that do something different?

src/catnap/runtime.rs
```diff
@@ -123,40 +124,45 @@ impl MemoryRuntime for PosixRuntime {
...
-        for i in 0..sga.sga_numsegs as usize {
-            let seg: &dmtr_sgaseg_t = &sga.sga_segs[i];
-            let allocation: Box<[u8]> = unsafe {
-                Box::from_raw(slice::from_raw_parts_mut(
-                    seg.sgaseg_buf as *mut _,
-                    seg.sgaseg_len as usize,
-                ))
-            };
-            drop(allocation);
+        // Bad scatter-gather.
```
BrianZill, line 129: I don't like this declarative comment.  It's like saying "this is a bad scatter-gather", or scatter-gather is bad in general.  Something like "Check for bad scatter-gather array" would be better.

src/catnip/runtime/memory/manager.rs
```diff
@@ -86,17 +87,14 @@ impl MemoryManager {
     pub fn into_sgarray(&self, buf: DPDKBuf) -> Result<dmtr_sgarray_t, Fail> {
         let (mbuf_ptr, sgaseg): (*mut rte_mbuf, dmtr_sgaseg_t) = match buf {
             // Heap-managed buffer.
-            DPDKBuf::External(bytes) => {
-                // We have to do a copy here since `Bytes` uses an `Arc<[u8]>` internally and has
-                // some additional bookkeeping for an offset and length, but we want to be able to
-                // hand off a raw pointer up the application that they can free later.
```
BrianZill, line 92: Are we changing the contract with the user here?  This comment makes it sound like the user was allowed to call "free" (i.e. the default memory allocators free() routine) on any of the sga_seg_buf pointers.  Would the new code still allow this?  Or must the user now call free_sgarray()?  I'm okay with the latter, just wondering what we're changing here.

We really need a document that defines the application level interface precisely.

src/catnip/runtime/memory/manager.rs
```diff
@@ -198,10 +196,11 @@ impl MemoryManager {
...
             // Release heap-managed buffer.
             let sgaseg: dmtr_sgaseg_t = sga.sga_segs[0];
-            let (ptr, len): (*mut c_void, usize) = (sgaseg.sgaseg_buf, sgaseg.sgaseg_len as usize);
-            let allocation: Box<[u8]> =
-                unsafe { Box::from_raw(slice::from_raw_parts_mut(ptr as *mut _, len)) };
-            drop(allocation);
+            let (data_ptr, length): (*mut u8, usize) =
+                (sgaseg.sgaseg_buf as *mut u8, sgaseg.sgaseg_len as usize);
+
+            // Convert back raw slice to a heap buffer and drop allocation.
```
BrianZill, line 202: This technically isn't a Rust "slice", but rather just a pointer length pair, correct?

src/demikernel/dbuf.rs
```diff
@@ -0,0 +1,189 @@
...
+        self.len -= nbytes;
+    }
+
+    /// Removes bytes from the end of the target data buffer.
+    fn trim(&mut self, nbytes: usize) {
+        if nbytes > self.len {
+            panic!(
+                "trimming past beginning of buffer: {} vs {}",
+                nbytes, self.len
+            );
```
BrianZill, line 185: This comment is more applicable to your formatting PR.

Ideally, this (lines 182 to 185) would all be on one line, like for line 173.  But it looks like this is formatting for 80-character lines.  I haven't used a punch-card machine in over 40 years, or even a VT-100 or VT-220 in over 30-something years.  I think we can reasonably go to 120-character lines now.  But if this was still too long for one line, this formatting is still wrong, as line 184 should be two lines, one each for "nbytes," and "self.len".

src/demikernel/dbuf.rs
```diff
@@ -0,0 +1,189 @@
...
+
+//==============================================================================
+// Associated Functions
+//==============================================================================
+
+/// Associated Functions for Data Buffers
+impl DataBuffer {
+    // Creates a data buffer with a given capacity.
+    pub fn new(capacity: usize) -> Result<Self, Fail> {
+        // Bad capacity.
```
BrianZill, line 48: "Check Arguments." would be a better comment here.

src/demikernel/dbuf.rs
```diff
@@ -0,0 +1,189 @@
...
+        Ok(Self {
+            data: unsafe { Some(Arc::new_zeroed_slice(capacity).assume_init()) },
+            offset: 0,
+            len: capacity,
+        })
+    }
+
+    /// Creates a data buffer from a raw pointer and a length.
+    pub fn from_raw_parts(data: *mut u8, len: usize) -> Result<Self, Fail> {
+        // Bad length.
```
BrianZill, line 63: "Check arguments."

src/demikernel/dbuf.rs
```diff
@@ -0,0 +1,189 @@
...
+        if len == 0 {
+            return Err(Fail::new(libc::EINVAL, "zero-length buffer"));
+        }
+
```
BrianZill, line 67: I wonder if we should attempt some sort of raw pointer verification here.  Sure, it's hard to tell if a pointer is valid (at least without attempting to dereference it), and zero (a.k.a. null) is a valid pointer value on some systems.  But we might want to check against zero anyways, as it most likely would be an error.  In the interest of catching bugs fast.

What do you think?

src/catnap/runtime.rs
ppenna: There is no semantic difference between what you suggested from what I did. It's all about the style that I like to follow - avoid one-liners. Anyways I merged the code and you asked.

src/catnap/runtime.rs
ppenna: Sure. I've changed it to what you suggested.

src/catnip/runtime/memory/manager.rs
ppenna: I don't know who wrote this. This comment was outdated and wrong, that is why I dropped it. The user is not allowed to call free when using memory allocated this way. We don't support use-after free. For this we should provider full integration with C's memory allocator, which is on the roadmap.

src/catnip/runtime/memory/manager.rs
ppenna: This is technically a "raw part", since it was produced by a previous call to `slice::from_raw_parts()`. Please be clear on the action item on this.

src/demikernel/dbuf.rs
ppenna: This is not about an 80-character line configuration, and neither of your suggestions work.

It is imposed by the rust formatting style applied to the whole source tree. It imposes a 100-character maximum width: 

https://github.com/demikernel/demikernel/blob/master/rustfmt.toml#L1

Note that changing this parameter will affect the entire code base.

src/demikernel/dbuf.rs
ppenna: Sure.

src/demikernel/dbuf.rs
ppenna: I've introduced null-pointer checking.

src/catnip/runtime/memory/manager.rs
BrianZill: Okay, thanks.

src/demikernel/dbuf.rs
BrianZill: Great, thanks.

src/catnip/runtime/memory/manager.rs
BrianZill: I was referring to the use of the term "slice" in the comment.  This function has no idea where the pointer length pair came from, just that it was in the scatter-gather array that it's freeing.  There's an assumption here that it came from the same memory allocator that the DataBuffer type uses, as it converts it back into one just to free it.  I was going to comment on that elsewhere.

But for now, I'd just get rid of the term "slice" in the comment.

src/demikernel/dbuf.rs
BrianZill: Ah, that's the file we need to fix!  But that isn't really relevant to this PR.

src/catnap/runtime.rs
```diff
@@ -123,40 +124,46 @@ impl MemoryRuntime for PosixRuntime {
...
     }
 
-    /// Clones a [dmtr_sgarray_t] into a memory buffer.
-    fn clone_sgarray(&self, sga: &dmtr_sgarray_t) -> Result<Bytes, Fail> {
-        let mut len: u32 = 0;
-        for i in 0..sga.sga_numsegs as usize {
-            len += sga.sga_segs[i].sgaseg_len;
+    /// Clones a scatter-gather array.
+    fn clone_sgarray(&self, sga: &dmtr_sgarray_t) -> Result<DataBuffer, Fail> {
+        // Bad scatter-gather.
```
BrianZill, line 151: Check arguments.

src/catpowder/runtime/memory.rs
```diff
@@ -65,39 +67,44 @@ impl MemoryRuntime for LinuxRuntime {
...
-        for i in 0..sga.sga_numsegs as usize {
-            let seg: &dmtr_sgaseg_t = &sga.sga_segs[i];
-            let allocation: Box<[u8]> = unsafe {
-                Box::from_raw(slice::from_raw_parts_mut(
-                    seg.sgaseg_buf as *mut _,
-                    seg.sgaseg_len as usize,
-                ))
-            };
-            drop(allocation);
+        // Bad scatter-gather.
```
BrianZill, line 72: Check arguments.

src/catpowder/runtime/memory.rs
```diff
@@ -65,39 +67,44 @@ impl MemoryRuntime for LinuxRuntime {
...
-        for i in 0..sga.sga_numsegs as usize {
-            let seg: &dmtr_sgaseg_t = &sga.sga_segs[i];
-            let seg_slice = unsafe {
-                slice::from_raw_parts(seg.sgaseg_buf as *mut u8, seg.sgaseg_len as usize)
-            };
-            buf[pos..(pos + seg_slice.len())].copy_from_slice(seg_slice);
-            pos += seg_slice.len();
+    /// Clones a scatter-gather array.
+    fn clone_sgarray(&self, sga: &dmtr_sgarray_t) -> Result<DataBuffer, Fail> {
+        // Bad scatter-gather.
```
BrianZill, line 94: Check arguments.

src/demikernel/dbuf.rs
```diff
@@ -0,0 +1,192 @@
...
+    /// Underlying data.
+    data: Option<Arc<[u8]>>,
+
+    /// Data offset.
+    offset: usize,
+
+    /// Data length
+    len: usize,
+}
+
```
BrianZill, line 39: Just a comment, not something to change now.

We should think about having various additional fields:
- an identifier for the memory allocator the buffer space came from.
- one that "remembers" the original length of the data allocation (i.e. before any adjust or trim), as it might be useful in the future.
- next and prev pointers for putting on lists w/o extra allocations.
- an "extra" field that each LibOs can use for whatever it wants (we should have space for another 64 bit field in this struct before we exceed the x86 cache line size).

BrianZill (changes requested):
This looks good.  Just fix those comments and it should be fine.

src/demikernel/dbuf.rs
ppenna: Sure. I created issue https://github.com/demikernel/demikernel/issues/106 to keep track of this.

src/catpowder/runtime/memory.rs
ppenna: Sure. It is fixed.

src/catpowder/runtime/memory.rs
ppenna: Sure. It is fixed.

src/catnap/runtime.rs
ppenna: Sure. It is fixed.

src/catnip/runtime/memory/manager.rs
ppenna: Sure. It is fixed.

src/demikernel/dbuf.rs
ppenna: Let's discuss this formatting issue on our weekly meeting.

BrianZill (approved):
Looks good now!

Issue #89, opened 2022-04-13
[catnap] Need to Set Nonblocking on New Accepted Sockets


iyzhang:
Description
========

We should set all sockets to nonblocking so that we can poll them without getting stuck in the Linux kernel.

ppenna:
This issue was fixed in commit fa8bc2514d926e51798bc7579b1eaaf5265efbe3.

PR #90, opened 2022-04-14
[catnap] Set Async Opts in New Socket

ppenna:
Description
========

In this PR I fix the bug described in issue https://github.com/demikernel/demikernel/issues/89.


iyzhang (approved):
Looks good to me

Issue #91, opened 2022-04-15
[catnap] `dmtr_accept` Always Returns Zero `QToken`


iyzhang:
Description
========

Redis calls dmtr_accept twice in a row and gets 0 back as the qtoken both times. Specifically the following events happen:
1. Redis calls dmtr_accept on the listening socket -> returns qt = 0
2. Redis calls dmtr_wait_any(qts = {0}) -> returns with a new connection
3. Redis calls dmtr_accept on the listening socket -> returns qt = 0
4. Redis calls dmtr_pop on the new connection socket -> returns qt = 1
5. Redis calls dmtr_wait_any(qts = {0, 1}) -> error "invalid queue token"

I assume something went wrong where the second call to dmtr_accept was not allocated a new queue token (based on the fact that the pop returns qtoken = 1 and it should probably be 2).

ppenna:
@iyzhang 

I cannot confirm this issue.

I've written a test that exercises the scenario that you detailed, and it runs fine.

Here is the server-side log that I got:

```
TRACE [demikernel::catnap] socket() domain=2, type=1, protocol=0
TRACE [demikernel::catnap] bind() qd=QDesc(0), local=Ipv4Endpoint { addr: 10.2.0.6, port: Port16(12345) }
TRACE [demikernel::catnap] listen() qd=QDesc(0), backlog=16
TRACE [demikernel::catnap] accept(): qd=QDesc(0)
TRACE [demikernel::catnap] wait_any2() [QToken(0)]
TRACE [demikernel::catnap::futures::accept] connection accepted (4)
TRACE [demikernel::catnap] accept(): qd=QDesc(0)
TRACE [demikernel::catnap] pop() qd=QDesc(1)
TRACE [demikernel::catnap] wait_any2() [QToken(0), QToken(1)]
TRACE [demikernel::catnap::futures::pop] data received (64 bytes)

[log continues]
```

I've submitted PR https://github.com/demikernel/demikernel/pull/92, to add this test to our regression suite.

Please let me know if it does not reflect the use case scenario that you described.

Finally, could you please provide me a full trace log of server-side? To get it, run set the `RUST_LOG=trace` environment variable in server side, before running Redis.

Thanks!

iyzhang:
That was just a guess on my part. Could it be an issue in the C bindings that your test doesn’t catch?

ppenna:
@iyzhang 

[...]

PR #92, opened 2022-04-18
[test] TCP Ping Pong with Multiple Accepts

ppenna:
Description
========

In this PR, I introduce a system-level test that covers the issue https://github.com/demikernel/demikernel/issues/91.

Additionally in commit 327e0cde4f11539504ed647d794f9e95d1bf3002, I turn off traces on polling path, to ease debugging on Catnap LibOS.


iyzhang (approved):
Yup, this looks correct

Issue #94, opened 2022-04-18
[pdpix] Don't Panic on System Calls


ppenna:
Description
========

Instead of failing gracefully, we are panicking on the following system calls:

- `dmtr_bind()`
- `dmtr_accept()`
- `dmtr_connect()`
- `dmtr_pushto()`
- `dmtr_push()`
- `dmtr_pop()`
- `dmtr_wait_any()`
- `dmtr_sgaalloc()`
- `dmtr_sgafree()`

ppenna:
This bug was fixed in commit 5e04170c5f4327078b4fdebc9bdd43ec2d2c8f63.

Issue #74, opened 2022-04-06
[pdpix] Add `setsockopt()`  to Socket API and Expose C Bindings


ppenna:
Description
========

We have to provide `setsockopt()` in the socket API and expose C bindings for it.

References
=======

- https://pubs.opengroup.org/onlinepubs/9699919799/functions/setsockopt.html

Issue #73, opened 2022-04-06
[pdpix] Add `getsockopt()`  to Socket API and Expose C Bindings


ppenna:
Description
========

We have to provide `getsockopt()` in the socket API and expose C bindings for it.

References
=======

- https://pubs.opengroup.org/onlinepubs/9699919799/functions/getsockopt.html

Issue #72, opened 2022-04-06
[pdpix] Provide an Implementation for `getsockname()`


ppenna:
Description
========

We do expose C bindings for `getsockname()` but we don't provide an implementation for this.

PR #99, opened 2022-04-19
[pdpix] Don't Panic on System Calls

ppenna:
Description
========

In this PR I fix issue https://github.com/demikernel/demikernel/issues/94.

Additionally, note also that I added dummy calls for the `setsockopt()` and `getsockopt()` system calls that shall be introduced in upcoming versions of Demikernel.

Related Issues
==========

- https://github.com/demikernel/demikernel/issues/72
- https://github.com/demikernel/demikernel/issues/73
- https://github.com/demikernel/demikernel/issues/74

### DIFF
--- a/src/demikernel/bindings.rs
+++ b/src/demikernel/bindings.rs
@@ -479,6 +479,23 @@ pub extern "C" fn dmtr_getsockname(qd: c_int, saddr: *mut sockaddr, size: *mut s
 }
 
 //==============================================================================
+// setsockopt
+//==============================================================================
+
+#[allow(unused)]
+#[no_mangle]
+pub extern "C" fn dmtr_setsockopt(
+    qd: c_int,
+    level: c_int,
+    optname: c_int,
+    optval: *const c_void,
+    optlen: socklen_t,
+) -> c_int {
+    // TODO: Implement this system call.
+    libc::ENOSYS
+}
+
+//==============================================================================
 // Standalone Functions
 //==============================================================================
 
--- a/src/demikernel/bindings.rs
+++ b/src/demikernel/bindings.rs
@@ -496,6 +496,23 @@ pub extern "C" fn dmtr_setsockopt(
 }
 
 //==============================================================================
+// getsockopt
+//==============================================================================
+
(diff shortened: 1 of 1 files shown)


src/demikernel/bindings.rs
```diff
@@ -332,20 +451,76 @@ pub extern "C" fn dmtr_sgaalloc(size: libc::size_t) -> dmtr_sgarray_t {
...
+
+//==============================================================================
+// Standalone Functions
+//==============================================================================
+
+/// Converts a [sockaddr] into a [Ipv4Endpoint].
+fn sockaddr_to_ipv4endpoint(saddr: *const sockaddr) -> Result<Ipv4Endpoint, Fail> {
+    let sin: libc::sockaddr_in =
+        unsafe { *mem::transmute::<*const sockaddr, *const libc::sockaddr_in>(saddr) };
+    let addr: Ipv4Addr = { Ipv4Addr::from(u32::from_be_bytes(sin.sin_addr.s_addr.to_le_bytes())) };
```
BrianZill, line 523: What is going on here with byte order?  To the naive reader (i.e. me :-)) this looks like s_addr is being taken as little endian and being read as big endian.

ppenna: I did not author this and I also have no clue why the person intentionally did it.

I just noticed that the same code was everywhere, so I at least created a function to do this copy-and-paste job.

I'm happy to take any improvements on this function afterwards, but considering its name, please let me know what your suggestion is.

BrianZill: Okay, mark it with a ToDo comment to review it later, file an Issue about it, and we'll leave it as is for now.  My guess is that whoever wrote this didn't understand what was going on and just kept putting in byte-order conversions until it worked :-).

BrianZill (approved):
Looks good now.

src/demikernel/bindings.rs
ppenna: Sure. I created issue https://github.com/demikernel/demikernel/issues/101 to keep track of this.

PR #102, opened 2022-04-20
[build] Check Code Formatting Style

ppenna:
Description
========

This PR addresses issue https://github.com/demikernel/demikernel/issues/111.

Note that I had to fix the existing formatting for existing code so as to be consistent with the standards.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,11 @@ all-libs:
 all-tests:
 	$(CARGO) build  --tests $(BUILD) $(CARGO_FEATURES) $(CARGO_FLAGS)
 
+check-fmt: check-fmt-rust
+
+check-fmt-rust:
+	$(CARGO) fmt -- --check
+
 clean:
 	rm -rf target ; \
 	rm -f Cargo.lock ; \
--- a/Makefile
+++ b/Makefile
@@ -39,13 +39,17 @@ endif
 
 all: all-libs all-tests
 
-all-libs:
+all-libs: check-fmt
 	$(CARGO) build $(BUILD) $(CARGO_FEATURES) $(CARGO_FLAGS)
 
-all-tests:
+all-tests: check-fmt
 	$(CARGO) build  --tests $(BUILD) $(CARGO_FEATURES) $(CARGO_FLAGS)
 
-check-fmt: check-fmt-rust
+check-fmt: check-fmt-c check-fmt-rust
+
+check-fmt-c:
+	$(shell find include/ -name "*.h" -name "*.hxx" -name "*.c" -name "*.cpp" -type f -print0 | xargs -0 clang-format --fallback-style=Microsoft --dry-run -Werror )
+	@exit $(.SHELLSTATUS)
 
 check-fmt-rust:
 	$(CARGO) fmt -- --check
--- a/include/dmtr/fail.h
+++ b/include/dmtr/fail.h
@@ -10,43 +10,44 @@
 #include <stdlib.h>
 
 #ifdef __cplusplus
-extern "C" {
+extern "C"
+{
 #endif
 
-typedef void (*dmtr_onfail_t)(int error_arg,
-      const char *expr_arg, const char *funcn_arg, const char *filen_arg,
-      int lineno_arg);
+    typedef void (*dmtr_onfail_t)(int error_arg, const char *expr_arg, const char *funcn_arg, const char *filen_arg,
+                                  int lineno_arg);
 
-#define DMTR_TRUE2(Error, Condition, ErrorCache) \
-    do { \
-        const int ErrorCache = (Error); \
-        if (DMTR_UNLIKELY(!(Condition))) { \
-            dmtr_fail(ErrorCache, #Condition, NULL, __FILE__, __LINE__);  \
-            return ErrorCache; \
-        } \
-   } while (0)
-
-#define DMTR_TRUE(Error, Condition) \
-    DMTR_TRUE2(Error, Condition, DMTR_TRUE_errorCache)
-
-#define DMTR_OK2(Error, ErrorCache) \
-    do { \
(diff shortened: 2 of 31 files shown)


include/dmtr/fail.h
```diff
@@ -10,43 +10,44 @@
...
-    DMTR_TRUE2(Error, Condition, DMTR_TRUE_errorCache)
-
-#define DMTR_OK2(Error, ErrorCache) \
-    do { \
-        const int ErrorCache = (Error); \
-        if (DMTR_UNLIKELY(0 != ErrorCache)) { \
-            dmtr_fail(ErrorCache, #Error, NULL, __FILE__, __LINE__); \
-            return ErrorCache; \
-        } \
+    typedef void (*dmtr_onfail_t)(int error_arg, const char *expr_arg, const char *funcn_arg, const char *filen_arg,
```
BrianZill, line 17: What style is this?  The only one I've consistently seen used inside Microsoft is: either all arguments must fit on the same line, or each argument must be on its own line.  At least for actual function calls, I don't know if this is normally enforced for typedefs of function calls.

Makefile
```diff
@@ -39,12 +39,21 @@ endif
...
 	$(CARGO) build $(BUILD) $(CARGO_FEATURES) $(CARGO_FLAGS)
 
-all-tests:
+all-tests: check-fmt
 	$(CARGO) build  --tests $(BUILD) $(CARGO_FEATURES) $(CARGO_FLAGS)
 
+check-fmt: check-fmt-c check-fmt-rust
+
+check-fmt-c:
+	$(shell find include/ -name "*.h" -type f -print0 | xargs -0 clang-format --fallback-style=Microsoft --dry-run -Werror )
```
BrianZill, line 51: Just .h files?

include/dmtr/fail.h
ppenna: This is Microsoft's style, applied by clang formatting tool:
- https://clang.llvm.org/docs/ClangFormatStyleOptions.html
- https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference

Let me know if you prefer any of the other styles.

Makefile
ppenna: We just have header files in the project right now, but I'm happy to add ".c" files to the rule.

include/dmtr/libos.h
```diff
@@ -8,209 +8,210 @@
...
+     * @param qtok_out Token for waiting for push to complete.
+     * @param qd Queue descriptor for queue to push to.
+     * @param sga Scatter-gather array with pointers to data to push.
+     *
+     * @return On successful completion zero is returned. On failure, an error code
+     * is returned instead.
+     */
+    DMTR_EXPORT int dmtr_push(dmtr_qtoken_t *qtok_out, int qd, const dmtr_sgarray_t *sga);
+
+    DMTR_EXPORT int dmtr_pushto(dmtr_qtoken_t *qtok_out, int qd, const dmtr_sgarray_t *sga,
```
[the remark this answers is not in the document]
ppenna: The code was formatted by the clag-format tool. Maybe it's all about picking up a different style. Let me know if there there any other that you prefer.

Makefile
BrianZill: Yes, I think we should, so they get checked when any are added.  Maybe we should do C++ files as well?

include/dmtr/fail.h
BrianZill: Hmm, not clear where that "Microsoft" style is actually coming from.  That page seems to only cover DotNet languages.  At least for this specific issue, it doesn't conform to what I'm used to seeing used internally.  I'll have to look over our options here.

The specific "Configurable Format Style Options" that appear to pertain to this issue are BinPackArguments and BinPackParameters, both of which should be false.  Although that doesn't completely cover it, as it still allows the first argument or parameter to appear on the same line as the function name.  I think AlignAfterOpenBracket would also need to be set to BAS_AlwaysBreak.

ppenna: I understand your concern. It should be awkward to have something named wrong this way.

Now, I would suggest us to stick to some built-in formatting style of clang, because it is easier for external contributors to match the standard that we choose to embrace. VIM, EMACs and VSCode have facilities on this direction.

For me, any of these built-in would be fine. Which one do you chose?

- LLVM A style complying with the [LLVM coding standards](https://llvm.org/docs/CodingStandards.html)
- Google A style complying with [Google’s C++ style guide](https://google.github.io/styleguide/cppguide.html)

- Chromium A style complying with [Chromium’s style guide](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/styleguide/styleguide.md)
- Mozilla A style complying with [Mozilla’s style guide](https://firefox-source-docs.mozilla.org/code-quality/coding-style/index.html)

- WebKit A style complying with [WebKit’s style guide](https://www.webkit.org/coding/coding-style.html)
- Microsoft A style complying with [Microsoft’s style guide](https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference)
- GNU A style complying with the [GNU coding standards](https://www.gnu.org/prep/standards/standards.html)

Makefile
ppenna: Sure. I'll add `.c`, `.cpp` and `.h` files. Am I missing any file extension?

Anyways, I'll apply these changes once we agree on the formatting style that we are going to use.

include/dmtr/fail.h
ppenna: @BrianZill any thoughts on this?

include/dmtr/libos.h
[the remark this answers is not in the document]
BrianZill: I responded in the other conversation about this.

Issue #95, opened 2022-04-18
[catnap] Use `recv()` Instead of `read()`


iyzhang:
Description
========

Catnap does not handle when the read syscall returns 0. Pedro suggests moving to recv instead of read to get better error messages.

ppenna:
As discussed in our core meeting, error handling will be improved differently: we should encode EOF in the operation result.

PR #103, opened 2022-04-22
[catnap] Use `recv()` and `send()`

ppenna:
Description
========

This PR fixes issue https://github.com/demikernel/demikernel/issues/95.

### DIFF
--- a/src/catnap/futures/pop.rs
+++ b/src/catnap/futures/pop.rs
@@ -7,7 +7,7 @@
 
 use ::nix::{
     errno::Errno,
-    unistd,
+    sys::socket,
 };
 use ::runtime::{
     fail::Fail,
@@ -75,12 +75,17 @@ impl Future for PopFuture {
     fn poll(self: Pin<&mut Self>, ctx: &mut Context<'_>) -> Poll<Self::Output> {
         let self_: &mut PopFuture = self.get_mut();
         let mut bytes: [u8; POP_SIZE] = [0; POP_SIZE];
-        match unistd::read(self_.fd as i32, &mut bytes[..]) {
-            // Operation completed.
-            Ok(nbytes) => {
-                trace!("data received ({:?} bytes)", nbytes);
+        match socket::recv(self_.fd, &mut bytes[..], socket::MsgFlags::empty()) {
+            // Some bytes were received, operation completed.
+            Ok(nbytes) if nbytes > 0 => {
+                trace!("data received ({:?}/{:?} bytes)", nbytes, POP_SIZE);
                 let buf: Bytes = Bytes::from_slice(&bytes[0..nbytes]);
                 Poll::Ready(Ok(buf))
+            },
+            // No bytes were received, let the user know about this.
+            Ok(nbytes) if nbytes == 0 => {
+                warn!("no data received");
+                Poll::Ready(Err(Fail::new(libc::EAGAIN, "no data received")))
             },
             // Operation in progress.
             Err(e) if e == Errno::EWOULDBLOCK || e == Errno::EAGAIN => {
@@ -92,6 +97,8 @@ impl Future for PopFuture {
                 trace!("pop failed ({:?})", e);
                 Poll::Ready(Err(Fail::new(e as i32, "operation failed")))
             },
+            // Should not happen.
+            _ => panic!("unexpected error for recv()"),
         }
     }
 }
--- a/src/catnap/futures/push.rs
+++ b/src/catnap/futures/push.rs
@@ -7,7 +7,7 @@
 
 use ::nix::{
     errno::Errno,
-    unistd,
+    sys::socket,
 };
 use ::runtime::{
     fail::Fail,
@@ -66,11 +66,16 @@ impl Future for PushFuture {
     /// Polls the target [PushFuture].
     fn poll(self: Pin<&mut Self>, ctx: &mut Context<'_>) -> Poll<Self::Output> {
         let self_: &mut PushFuture = self.get_mut();
-        match unistd::write(self_.fd, &self_.buf[..]) {
-            // Operation completed.
-            Ok(_) => {
-                trace!("data pushed ({:?} bytes)", self_.buf.len());
+        match socket::send(self_.fd, &self_.buf[..], socket::MsgFlags::empty()) {
+            // Some bytes were sent, operation completed.
+            Ok(nbytes) if nbytes > 0 => {
+                trace!("data pushed ({:?}/{:?} bytes)", nbytes, self_.buf.len());
                 Poll::Ready(Ok(()))
+            },
+            // No bytes were sent, let the user know about this.
+            Ok(nbytes) if nbytes == 0 => {
+                warn!("no data pushed");
+                Poll::Ready(Err(Fail::new(libc::EAGAIN, "no data pushed")))
             },
             // Operation in progress.
             Err(e) if e == Errno::EWOULDBLOCK || e == Errno::EAGAIN => {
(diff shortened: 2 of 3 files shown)


BrianZill:
I'm confused.  A recv() is equivalent to a read if you don't specify any flags (which this code doesn't).  Also, recv() is traditionally only used on connected sockets, which presumably this isn't (because you're not treating a zero return value as "connection closed").  So why the switch from read() to recv()?

The change here appears to be to return EAGAIN if the read returns 0.  But if the Rustian recv() really means to return EAGAIN, it should directly return an Err of EAGAIN, not 0.  You should only (legitimately) be getting zero back from reads on non-connected sockets (and as a "connection closed" indication on connected sockets).

Is this a connected socket?  If so, how is a connection close being conveyed to the user?  If not, what is going on here?  Is this the "raw socket" LibOS?  Or the "demikernel API over Posix sockets LibOS"?  Anyhow, I'm confused.

ppenna:
@BrianZill 

This issue concerns Catnap, which is the LibOS over POSIX sockets.

You are right about the semantic equivalence between `read()` / `recv()` and `write()` / `send()`.

However, when comes to error handling, the linux manual pages are more clear on the return values of `recv()` and `send()`. For instance, this is what is stated on [read(2)](https://man7.org/linux/man-pages/man2/read.2.html):

> Other errors may occur, depending on the object connected to fd.

By relying on `send()` and `recv()`, we are restricting the object connected to `fd` to a socket one.

Anyways, the bug fix here mostly concerns the scenario where we read zero bytes from the socket. According to the [recv(2)](https://man7.org/linux/man-pages/man2/recv.2.html):

> RETURN VALUE
> These calls return the number of bytes received, or -1 if an
> error occurred.  In the event of an error, errno is set to
> indicate the error.
> 
> When a stream socket peer has performed an orderly shutdown, the
> return value will be 0 (the traditional "end-of-file" return).
> 
> Datagram sockets in various domains (e.g., the UNIX and Internet
> domains) permit zero-length datagrams.  When such a datagram is
> received, the return value is 0.
> 
> The value 0 may also be returned if the requested number of bytes
> to receive from a stream socket was 0.

That is, `recv()` may return zero when either we get disconnected, or we actually read no bytes.

Before this PR, we were returning a zero-length buffer in both cases. This is not good because the application may need to act accordingly, which is the case for Redis.

The workaround that @iyzhang asked for was to return an error code so as to let the application know that we have hit this corner case. 

To this end, one could argue on returning `ENOTCONN`, but this would lead to the same issues of returning a zero-length buffer: we would not be able to distinguish between a legitim `ENOTCONN` return value from `recv()` and a spurious one, caused by a zero-byte return. Actually, having any return value besides `EAGAIN` and `EWOULDBLOCK` would lead to this problem, because these two values are the only ones that we handle internally in Catnap.

If you are interested, I suggest us to discuss this on our weekly engineering meeting.

BrianZill:
I think you may be reading/interpreting that manual page incorrectly.  recv() (or read(), the manual explicitly states that for a socket, the behavior is exactly the same except for the flags argument -- at least for the C API, maybe the Rust library code is doing something different) will only return zero _on a connected socket_ in two cases: 1) the other side closed their end (this is the common case), and 2) the user _explicitly asked to "read" zero bytes_ in their call to recv() (or read()).  The latter case is a normally a pretty rare condition, as the user is essentially requesting a no-op, which isn't very productive.  Anyhow, the user can tell the difference between these two cases because they presumably know if they've asked to read nothing.

I'll ignore non-connected (i.e. datagram) sockets for the moment, as that doesn't appear to be meaningful here, but they can return zero if a zero-length datagram arrives (e.g. a UDP datagram with just headers and no data).  Again, the user presumably knows what kind of socket they have.  And there is no connection to close, so a zero-length read is unambiguous.

Now for the demikernel API, the user doesn't specify how much data to read, or where to put the data that gets read.  The user just asks to read something, and if/when there is data to read, the demikernel API hands over a buffer with some data in it.  The user should only get back a zero length buffer for a connected socket when the connection is closing (i.e. case #1 above).  There should be no other case where the user gets a zero length buffer from the demikernel API because we should never be issuing a read (or recv()) for zero bytes.

If Irene is getting back zero length buffers when the connection hasn't been closed, then we presumably have a bug in the demikernel over Posix code where we're issuing a read (or recv()) call for zero bytes.  That's where the bug lies, not in changing the return value.

iyzhang:
I’m not issuing read or recv from Redis since I’m going through the Demikernel API. I wanted push to return an error code when the connection closes, not 0 since in the Demikernel API 0 generally means ok. Pedro and I discussed and thought it would be easier if we used recv since it does differ in error behavior from read, ie, it explicitly returned a errno for closed connection rather than just returning 0.

BrianZill:
Okay, this is a little long, but I think we're conflating some issues and I'm trying to be as clear as possible.

The read() vs recv() issue:  These system calls **are** semantically the same, _including the return error codes_.  The only difference is that recv() takes an additional flags argument (which we're not using, so they are functionally equivalent).  Both of these return an error code if they are issued on an not-connected stream socket, and both of these return zero when issued on a closing stream socket.  A zero-length "read" is the end-of-connection indicator for stream sockets when using any flavor of the socket API that I'm aware of (the original intent was to mimic the Unix EOF indicator when reading from a file or pipe).  I think there may have been some confusion here with the ENOTCONN error value that recv() (and read()!) can return if issued against a stream socket that has not been connected (yet - as in created, but not yet connected or accepted).  This error code isn't the "connection closing" indicator.

The connection closing indicator in the Demikernel API: I would strongly recommend leaving this as a read of zero bytes for general compatibility with socket applications.  Also, I don't think a closing connection should be thought of as an error condition anymore than reading an EOF on a file or a pipe is.  Return "ok" with a zero length buffer is an unambiguous indication that the other end of the connection is done sending.  Especially if we plan on using the Demikernel API for file I/O in the future.

The return value from wait or wait_any:  The immediate return value from these functions should be 0 ("ok") so long as the wait itself can be issued ok (i.e. it should error return only if the arguments given are improper or incorrect in some way).  The return value from the operation that was waited on should be returned in the dmtr_qresult_t (the qr_value code IIRC).  So the "0 == ok" for the actual wait call is still held even when a "read" returns a zero-length buffer.

Yes, the catnap code that implements the Demikernel API over Posix sockets is the code that issues the read() or recv() call.  So if it is getting back 0 from that read/recv, it means the other end of the connection is done sending (has called close or shutdown on its end), or it means the catnap code has a bug in how it is issuing the call.  It looks like the catnap code uses a non-zero constant for read buffer size, so that should be okay.  The only other probable bug here is in the file descriptor mapping, i.e. if catnap is using the wrong file descriptor when issuing the read.  I suppose the Rust wrapper around the Linux system call could also be doing something wrong, causing a zero-length read to be issued, but that seems unlikely.  Or maybe the result value from the read/recv isn't being put into the dmtr_qresult_t correctly?  You shouldn't be getting zero bytes back unless the connection is closing.

iyzhang:
Ah, ok this is interesting. So, there are actually two opportunities to give the user feedback on push since it is asynchronous unlike POSIX read or write. First, the returned value on the initial push will be 0 if OK, otherwise an error code. If 0, then the queue token is valid and can be used in wait any. Then in wait any, we return a qresult, which gives us the ability to return a variety of error codes but in general if the wait any returns 0, then the qresult is valid and useful and otherwise, the qresult is an error op (and maybe we would want it to return more info). So if either push or wait any return anything other than 0, I want to be able to surface an error to the application (Redis). In the particular case where this happened, the client closed the connection while the server was still trying to write, so it seemed like an error is appropriate.

But it seems like you are suggesting a different API where push returns how many bytes were successfully enqueued? And then wait_any returns how many bytes were successfully sent? Let's chat more about this offline on Wednesday and come up with a more concrete definition for the API.

ppenna:
Based on our core meeting we have decided to approach the problem differently, and encoded EOF in the operation result structure.

Issue #144, opened 2022-07-21
[pdpix] Missing Endianness Conversion for Port Number


ppenna:
Description
--------------

The `sockaddr_to_ipv4endpoint()` function is missing the endianness conversion for the port number.

https://github.com/demikernel/demikernel/blob/eb760f9801db7e7a2073253d909fc8fe5584269b/src/demikernel/bindings.rs#L516-L521

Expected Behavior
============

We should convert port number bytes, likewise we do for the IPv4 address.

PR #145, opened 2022-07-21
[pdpix] Missing Endianness Conversion for Port Number

ppenna:
Description
-------------

This PR fixes https://github.com/demikernel/demikernel/issues/144.

Summary of Changes
-------------------------

- Renamed `sockaddr_to_ipv4endpoint()` to `test_sockaddr_to_socketaddrv4()`
- Dropped useless byte ordering conversion for socket address in `test_sockaddr_to_socketaddrv4`.
- Introduced a test unit for bug https://github.com/demikernel/demikernel/issues/144.
- Fixed missing edianness conversion for port number in `test_sockaddr_to_socketaddrv4()`
- Added check for `AF_INET` in `test_sockaddr_to_socketaddrv4()`

### DIFF
--- a/src/rust/demikernel/bindings.rs
+++ b/src/rust/demikernel/bindings.rs
@@ -514,6 +514,7 @@ pub extern "C" fn demi_getsockopt(
 
 /// Converts a [sockaddr] into a [SocketAddrV4].
 fn sockaddr_to_socketaddrv4(saddr: *const sockaddr) -> Result<SocketAddrV4, Fail> {
+    // TODO: Change the logic bellow and rename this function once we support V6 addresses as well.
     let sin: libc::sockaddr_in = unsafe { *mem::transmute::<*const sockaddr, *const libc::sockaddr_in>(saddr) };
     if sin.sin_family != libc::AF_INET as u16 {
         return Err(Fail::new(libc::ENOTSUP, "communication domain  not supported"));
@@ -525,6 +526,8 @@ fn sockaddr_to_socketaddrv4(saddr: *const sockaddr) -> Result<SocketAddrV4, Fail
 
 #[test]
 fn test_sockaddr_to_socketaddrv4() {
+    // TODO: assign something meaningful to sa_family and check it once we support V6 addresses as well.
+
     // SocketAddrV4: 127.0.0.1:80
     let saddr: libc::sockaddr = {
         sockaddr {
--- a/src/rust/demikernel/bindings.rs
+++ b/src/rust/demikernel/bindings.rs
@@ -512,10 +512,31 @@ pub extern "C" fn demi_getsockopt(
 // Standalone Functions
 //==============================================================================
 
-/// Converts a [sockaddr] into a [Ipv4Endpoint].
+/// Converts a [sockaddr] into a [SocketAddrV4].
 fn sockaddr_to_ipv4endpoint(saddr: *const sockaddr) -> Result<SocketAddrV4, Fail> {
     // TODO: Review why we need byte ordering conversion here.
     let sin: libc::sockaddr_in = unsafe { *mem::transmute::<*const sockaddr, *const libc::sockaddr_in>(saddr) };
+    if sin.sin_family != libc::AF_INET as u16 {
+        return Err(Fail::new(libc::ENOTSUP, "communication domain  not supported"));
+    };
     let addr: Ipv4Addr = { Ipv4Addr::from(u32::from_be_bytes(sin.sin_addr.s_addr.to_le_bytes())) };
     Ok(SocketAddrV4::new(addr, sin.sin_port))
+}
+
+#[test]
+fn test_sockaddr_to_ipv4endpoint() {
+    // SocketAddrV4: 127.0.0.1:80
+    let saddr: libc::sockaddr = {
+        sockaddr {
+            sa_family: libc::AF_INET as u16,
+            sa_data: [0, 80, 127, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0],
+        }
+    };
+    match sockaddr_to_ipv4endpoint(&saddr) {
+        Ok(addr) => {
+            assert_eq!(addr.port(), 80);
+            assert_eq!(addr.ip(), &Ipv4Addr::new(127, 0, 0, 1));
+        },
+        _ => panic!("failed to convert"),
+    }
 }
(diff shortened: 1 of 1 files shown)


src/rust/demikernel/bindings.rs
```diff
@@ -512,10 +512,28 @@ pub extern "C" fn demi_getsockopt(
 // Standalone Functions
 //==============================================================================
 
-/// Converts a [sockaddr] into a [Ipv4Endpoint].
+/// Converts a [sockaddr] into a [SocketAddrV4].
 fn sockaddr_to_ipv4endpoint(saddr: *const sockaddr) -> Result<SocketAddrV4, Fail> {
```
BrianZill, line 516: This function name is left over from when we still had an IPv4Endpoint structure.  Can you change it to something like sockaddr_to_SocketAddrV4?

BrianZill: Or if you want to start putting in multi-protocol support, make it sockaddr_to_SocketAddr (and return a SocketAddr).

src/rust/demikernel/bindings.rs
```diff
@@ -512,10 +512,28 @@ pub extern "C" fn demi_getsockopt(
...
-    // TODO: Review why we need byte ordering conversion here.
     let sin: libc::sockaddr_in = unsafe { *mem::transmute::<*const sockaddr, *const libc::sockaddr_in>(saddr) };
-    let addr: Ipv4Addr = { Ipv4Addr::from(u32::from_be_bytes(sin.sin_addr.s_addr.to_le_bytes())) };
-    Ok(SocketAddrV4::new(addr, sin.sin_port))
+    let addr: Ipv4Addr = Ipv4Addr::from(sin.sin_addr.s_addr.to_be());
+    let port: u16 = sin.sin_port.to_be();
```
BrianZill, line 519: Do you need these "to_be()" things?  I find it jarring to see them there, as they shouldn't do anything.  In any sort of socket address, the address and port numbers are always supposed to be in Network Byte Order, which was decreed by the gods, or at least Jon Postel*, to be big-endian.

* I don't know for sure it was Jon, but he seems like a likely candidate.

BrianZill: And if you do need them for this code to work correctly, then they're wrong.

src/rust/demikernel/bindings.rs
```diff
@@ -512,10 +512,28 @@ pub extern "C" fn demi_getsockopt(
...
+    Ok(SocketAddrV4::new(addr, port))
```
BrianZill, line 520: It's really too bad there isn't a new() function for SocketAddrV4 (and/or SocketAddr) that takes a C-style sockaddr pointer directly.  Then we wouldn't have to be concerned about breaking out the Ipv4Addr and port number ourselves.

src/rust/demikernel/bindings.rs
```diff
@@ -512,10 +512,28 @@ pub extern "C" fn demi_getsockopt(
...
+}
+
+#[test]
+fn test_sockaddr_to_ipv4endpoint() {
+    // SocketAddrV4: 127.0.0.1:80
+    let saddr: libc::sockaddr = {
+        sockaddr {
+            sa_family: 0,
+            sa_data: [0, 80, 127, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0],
```
BrianZill, line 529: For an IPv4 address, sa_family should be AF_INET (2).  Should we check this?  If we implemented sockaddr_to_SocketAddr, which enum it returns should depend upon this field.

src/rust/demikernel/bindings.rs
ppenna: Sure. I renamed this function.

src/rust/demikernel/bindings.rs
ppenna: I understand your perspective.

However, in this PR I aimed for fixing the endianness conversion for IPv4 addresses only.

Anyways, I added TODOs on the source code to note your suggestion.

src/rust/demikernel/bindings.rs
ppenna: Indeed, but that what we have for SocketV4 addresses. Any suggestions here?

src/rust/demikernel/bindings.rs
ppenna: > And if you do need them for this code to work correctly, then they're wrong. Since these values are big-endian by definition, calling "to_be()" on them actually turns them little-endian. Which is host order on x86. So this works until we try to run it on a big-endian machine.

We do need them. I wrote the unit test to precisely illustrate this.

Any of the following alternatives is buggy:
- Drop calls `to_be()`
- Replace `to_be()` for `to_le()`
- Replace `to_be()` for `to_ne()`

Note that before this PR we were not performing a conversion on the port number, and we were operating on bad port numbers. You can check this on unmodified code by turning on rust logging.

BrianZill: I understand all that.  This is a subtle, but important point.

We do need a byte-swap (on x86) for this to work correctly.  But we should do it in a way that is both portable (to big-endian architectures) and makes sense to the reader.

Here's the issue:
- A sockaddr (the C struct used by the Berkeley "sockets" interface) is **always** in network byte order.
> - And network-byte order is **always** big-endian.
- Rust's SockAddrV4 constructor takes an Ipv4Addr (ip address) and an u16 (port number).
> - The u16 is expected to be in host order.
> - The constructor for Rust's Ipv4Addr struct that we're using takes an u32, which is expected to be in host order.

So we're taking something that is always network byte order (big-endian) and converting it to host order so we can create the Rust structure.  A human reading this code would expect whatever conversion function we call to reflect this in its name.  So something named network_to_host_endian() is what we want here.

ppenna: As we discussed I've updated the code to rely on `from_be()` instead of `to_be()`.

src/rust/demikernel/bindings.rs
BrianZill: I don't think there is much we can do about this.

src/rust/demikernel/bindings.rs
BrianZill: Thanks!

BrianZill (approved):
Beautiful, thanks!

Issue #147, opened 2022-07-28
[catnap] Source Address in `OperationResult` for `pop()` Is Set to `None`


ppenna:
Description
-------------

In Catnap, after successfully completing a `pop()` operation, the source address in the `OperationResult` structure is always set to `None`.

As a consequence, a UDP Catnap application must always know the address of the remote peer, which is wrong.

https://github.com/demikernel/demikernel/blob/cbf1243786b79d6e65e1642f3f267cbf41c1657e/src/rust/catnap/futures/mod.rs#L132-L136

Expected Behavior
---------------------

When possible, the source address of the received message should be stored in the `OperationResult` structure.

Related Issues
----------------

- https://github.com/demikernel/demikernel/issues/148

Issue #148, opened 2022-07-28
[catcollar] Source Address in `OperationResult` for `pop()` Is Set to `None`


ppenna:
Description
-------------

In Catcollar, after successfully completing a `pop()` operation, the source address in the `OperationResult` structure is always set to `None`.

As a consequence, a UDP Catnap application must always know the address of the remote peer, which is wrong.

https://github.com/demikernel/demikernel/blob/cbf1243786b79d6e65e1642f3f267cbf41c1657e/src/rust/catcollar/futures/mod.rs#L132-L136

Expected Behavior
---------------------

When possible, the source address of the received message should be stored in the `OperationResult` structure.

Additional Information
--------------------------

- To fix this, we should rely on the `io_uring_prep_recvmsg()` operation and change the following code: https://github.com/demikernel/demikernel/blob/cbf1243786b79d6e65e1642f3f267cbf41c1657e/src/rust/catcollar/iouring.rs#L168
- When completing operations subitted to the underlying IOUring we should change the return of the follow, to enable source addresses to be returned:
https://github.com/demikernel/demikernel/blob/cbf1243786b79d6e65e1642f3f267cbf41c1657e/src/rust/catcollar/iouring.rs#L178

Related Issues
----------------

- https://github.com/demikernel/demikernel/issues/147
- https://github.com/demikernel/liburing-rs/issues/7

PR #149, opened 2022-07-28
[catnap] Set Source Address in `OperationResult` for `pop()`

ppenna:
Description
-------------

This PR fixes https://github.com/demikernel/demikernel/issues/147.

Sumary of Changes
----------------------

- Changed `PopFuture::poll()` to use `recvfrom()` and return the source address of the message
- Changed the `Operation::get_result()` to set the address in `OperationResult` accordingly

Additional Information
--------------------------

- The verbose conversion in `recvfrom()` is required because `nix` does not provide a conversion function between `SockaddrStorage` and `SocketAddrV4` (https://docs.rs/nix/latest/nix/sys/socket/union.SockaddrStorage.html). Once we drop the use of `nix` in favor of `libc` we should get rid of this.
- There is a similar issue with Catcollar, but it is harder to fix. See https://github.com/demikernel/demikernel/issues/148 for more details.

### DIFF
--- a/src/rust/catnap/futures/pop.rs
+++ b/src/rust/catnap/futures/pop.rs
@@ -7,7 +7,10 @@
 
 use ::nix::{
     errno::Errno,
-    sys::socket,
+    sys::{
+        socket,
+        socket::SockaddrStorage,
+    },
 };
 use ::runtime::{
     fail::Fail,
@@ -19,6 +22,10 @@ use ::runtime::{
 };
 use ::std::{
     future::Future,
+    net::{
+        Ipv4Addr,
+        SocketAddrV4,
+    },
     os::unix::prelude::RawFd,
     pin::Pin,
     task::{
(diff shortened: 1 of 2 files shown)


src/rust/catnap/futures/pop.rs
```diff
@@ -69,18 +76,29 @@ impl PopFuture {
...
+            Ok((nbytes, socketaddr)) => {
                 trace!("data received ({:?}/{:?} bytes)", nbytes, POP_SIZE);
                 let buf: Buffer = Buffer::Heap(DataBuffer::from_slice(&bytes[0..nbytes]));
-                Poll::Ready(Ok(buf))
+                let addr: Option<SocketAddrV4> = match socketaddr {
+                    Some(addr) => match addr.as_sockaddr_in() {
+                        Some(sin) => {
+                            let ip: Ipv4Addr = Ipv4Addr::from(sin.ip());
+                            let port: u16 = sin.port();
+                            Some(SocketAddrV4::new(ip, port))
```
BrianZill, line 95: Don't you need to perform network-to-host byte-order conversion here?  nix::sys::socket::recvfrom looks to just be returning the sockaddr in a byte array, which would presumably yield big-endian values.  Unless something funky is happening in "as_sockaddr_in" (which I can't seem to find documentation for, where is that coming from?).

ppenna: @BrianZill, we don't need to because `as_sockaddr_in().ip()` and `as_sockaddr_in().port()` return the IP address and port number in host endianness, respectively. See: https://docs.rs/nix/latest/nix/sys/socket/struct.SockaddrIn.html

BrianZill: Cool, thanks.

BrianZill (approved):
Looks good!

Issue #131, opened 2022-06-08
[test] System Tests for C Interface


ppenna:
Context
---------

Our C regression suite currently misses system tests.

Due to this we recently missed a critical bug to be introduced in the `unstable` branch https://github.com/demikernel/demikernel/issues/130.

Proposed Solution
---------------------

Provide C system tests for the following Rust tests:

- [x] `udp_push_pop` #133 
- [X] `udp_ping_pong` #150
- [x] `tcp_push_pop` #139 
- [x] `tcp_ping_pong` #151

PR #151, opened 2022-08-03
[test] TCP Ping Pong for C API

ppenna:
Description
-------------

This PR is a partial fix for https://github.com/demikernel/demikernel/issues/131.

Summary of Changes
-------------------------

- Introduced C test for TCP ping pong

### DIFF
--- a/tests/c/tcp-ping-pong.c
+++ b/tests/c/tcp-ping-pong.c
@@ -1,0 +1,350 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+/* This should come first. */
+#define _POSIX_C_SOURCE 200809L
+
+/*====================================================================================================================*
+ * Imports                                                                                                            *
+ *====================================================================================================================*/
+
+#include <arpa/inet.h>
+#include <assert.h>
+#include <demi/libos.h>
+#include <demi/sga.h>
+#include <demi/wait.h>
+#include <signal.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/socket.h>
+
+/*====================================================================================================================*
+ * Constants                                                                                                          *
+ *====================================================================================================================*/
+
+/**
+ * @brief Data size.
+ */
+#define DATA_SIZE 64
+
+/**
+ * @brief Maximum number of bytes to transfer.
+ */
+#define MAX_BYTES (DATA_SIZE * 1024)
+
+/*====================================================================================================================*
+ * sighandler()                                                                                                       *
+ *====================================================================================================================*/
+
+/**
+ * @brief Signal handler.
+ *
+ * @param signum Number of received signal.
+ */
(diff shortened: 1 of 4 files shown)


tests/c/tcp-ping-pong.c
```diff
@@ -0,0 +1,346 @@
...
+        memset(sga.sga_segs[0].sgaseg_buf, 1, DATA_SIZE);
+
+        /* Push scatter-gather array. */
+        push_wait(sockqd, &sga, &qr);
+
+        /* Release sent scatter-gather array. */
+        assert(demi_sgafree(&sga) == 0);
+
+        /* Pop data scatter-gather array. */
+        pop_wait(sockqd, &qr);
```
BrianZill, line 256: Shouldn't we check the result for something?  (Beyond what pop_wait() does internally, I mean)

Note that since TCP doesn't guarantee preservation of message boundaries, there is no guarantee here that we'll get back the same number of bytes that we sent in the push_wait above.  But you could turn this into a loop of pop_wait()s that continue until the expected number of bytes come back.

tests/c/tcp-ping-pong.c
```diff
@@ -0,0 +1,346 @@
...
+
+        nbytes += sga.sga_segs[0].sgaseg_len;
```
BrianZill, line 258: I think it would make more sense here to increment nbytes by the number of bytes received in the pop_wait(), rather than the number of bytes sent in the push_wait().  Especially since you're not currently checking that they're the same thing.

tests/c/tcp-ping-pong.c
```diff
@@ -0,0 +1,346 @@
...
+#include <sys/socket.h>
+
+/*====================================================================================================================*
+ * Constants                                                                                                          *
+ *====================================================================================================================*/
+
+/**
+ * @brief Data size.
+ */
+#define DATA_SIZE 64
```
BrianZill, line 29: It would be nice if both DATA_SIZE and MAX_BYTES were just default values for their respective values, with the actual values being overridable via command line arguments.

tests/c/tcp-ping-pong.c
ppenna: This was unintentional. Thanks for noting down.

tests/c/tcp-ping-pong.c
ppenna: Sure, but the three other tests are currently relying on compile-time constants as well. May we create a good first issue for this?

tests/c/tcp-ping-pong.c
ppenna: Sure. I missed this. I introduced a simple payload checking logic, similarly to what we have in the rust test.

tests/c/tcp-ping-pong.c
BrianZill: Okay.

BrianZill (approved):
Looks okay now.

PR #156, opened 2022-08-12
[catcollar] Set Source Address in `OperationResult` for `pop()`

ppenna:
Description
-------------

This PR fixes https://github.com/demikernel/demikernel/issues/148.

Summary of Changes
-------------------------

- Changed `IoUring` to rely on `io_uring_prep_recvmsg()` and thus be able to get the socket address of the source.
- Changed lookip logic in `IoUring` to use the `msghdr` rather than the buffer address
- Changed futures accordinly

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,8 +25,8 @@ yaml-rust = "0.4.5"
 log = "0.4.17"
 nix = "0.24.2"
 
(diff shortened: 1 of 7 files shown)


BrianZill:
This looks okay to me, but I'm not familiar with io_uring.  Maybe @anandbonde would like to take a look?

anandbonde (approved):
Looks good.

Issue #158, opened 2022-09-01
[pdpix] Support Coexistence of Different LibOS Types


ppenna:
Context
======

Currently, Demikernel does not support multiple LibOSes, each of which of a different type, to co-exist.

This may be limiting when an application interacts with multiple kernel-bypass technologies at the same time.

Proposed Solution
=============

In order to enable multiple LibOSes to co-exist, we shall work on the following steps:

- [x] Enable compilation of LibOSes to be composed (d61ae7bdf605d1c032c30e809fcafcf04ffe5798)
- [x] Enable dynamic selection of which LibOSes flavor should be instantiated (d61ae7bdf605d1c032c30e809fcafcf04ffe5798)
- [ ] Enable instantiation of multiple LibOSes.

PR #159, opened 2022-09-01
[pdpix] Build Composition of LibOSes

ppenna:
Description
--------------

This PR is a partial fix for https://github.com/demikernel/demikernel/issues/158.

Summary of Changes
-------------------------

- Enabled LibOS instantiation to fail rather than panicking.
- Enabled build composition of different flavors of LibOSes.
  - Build features `catnap-libos`, `catnip-libos`, `catpowder-libos` and `catcollar-libos` are no longer mutually exclusive.
- Enabled dynamic LibOS flavor selection using environment variables.


BrianZill (approved):
Looks good.

Issue #160, opened 2022-09-06
[pdpix] Manual Pages for System Calls


ppenna:
Description
-------------

We should write manual pages for system calls:

- [x] `demi_accept()` cf1827556d34e80cfe5642b91f9333f7a038cc95
- [x] `demi_bind()` d8a13f568770b5e39a3bc9d706a17d21810c00aa
- [x] `demi_close()` c9792c0431b5e962eb2d42bb0caa28b8c23fd0d2
- [x] `demi_connect()` 67d13322e119dedeb190d27cb006780c8091de5a
- [x] `demi_init()` 234a3c443b8097e94ddf40435c130555cd2ffd5b
- [x] `demi_listen()` f8ddec746003c32a0e23be45352b82d8164dc02a
- [x] `demi_pop()` 0739e64b2cba964a32bd3487361482841e6f29a0
- [x] `demi_push()` a5235493ef8c3b1dbcbfee6bdb70cb380d9fb221
- [x] `demi_pushto()` a5d6a069f56d1f0641687337d4ea8d3e1bad5e46
- [x] `demi_sgaalloc()` 16549cfb47cf8ffa942e745541ad8f770ad99cb2
- [x] `demi_sgafree()` 7429937b529fc42a47d3889a0197ee630168a0a8
- [x] `demi_socket()` 7b6dfde807ce06be11a1bf07a8eb122aef4725a2
- [x] `demi_wait_any()` a8137c8a004a746e52f985c807eae6919dcbb24a
- [x] `demi_wait()` a8137c8a004a746e52f985c807eae6919dcbb24a

PR #161, opened 2022-09-06
[doc] `demi_init()` Manual Page

ppenna:
Description
-------------

This is a partial fix for https://github.com/demikernel/demikernel/issues/160.

Summary of Changes
-------------------------

- Added manual page for `demi_init()` system call.

### DIFF
--- a/man/2/demi_init.md
+++ b/man/2/demi_init.md
@@ -1,0 +1,66 @@
+# `demi_init(2)`
+
+## Name
+
+`demi_init` - Initializes Demikernel.
+
+## Synopsis
+
+```c
+#include <demi/libos.h>
+
+int demi_init(int argc, char *const argv[]);
+```
+
+## Description
+
+`demi_init()` initializes Demikernel. It sets up devices, instantiates LibOSes, and performs general initialization
+tasks.
+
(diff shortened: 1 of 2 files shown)


man/2/demi_init.md
```diff
@@ -0,0 +1,59 @@
...
+|ID       | Argument      | Description             |
+|---------|---------------|-------------------------|
+| `ARG-1` | `--catcollar` | Enables Catcollar LibOS |
+| `ARG-2` | `--catnap`    | Enables Catnap LibOS    |
+| `ARG-3` | `--catnip`    | Enables Catnip LibOS    |
+| `ARG-4` | `--catpowder` | Enables Catpowder LibOS |
+
+ Any constraints/restrictions of arguments are detailed next:
+
+- `ARG-1` takes effect only on Linux hosts. If this is the only parameter passed to Demikernel, `demi_init()` will fail.
```
BrianZill, line 34: What other parameter needs to be passed along with this one (--catcollar) for demi_init() not to fail?  Below it says these four (LibOSes) are mutually exclusive, so presumably none of those, and no other possible arguments/parameters are listed.

PR #162, opened 2022-09-08
[inetstack] Moving In Source tree

ppenna:
Description
--------------

In this PR I move in the [inetstack](https://github.com/demikernel/inetstack) and [runtime](https://github.com/demikernel/runtime) source trees.


src/rust/lib.rs
```diff
@@ -5,6 +5,22 @@
 #![deny(clippy::all)]
 #![feature(maybe_uninit_uninit_array, new_uninit)]
 #![feature(try_blocks)]
+#![cfg_attr(feature = "strict", deny(clippy:all))]
```
BrianZill, line 8: Just out of curiosity (this was also present in the old inetstack/src/lib.rs these were apparently pulled from) does this actually do anything different if feature "strict" is specified?  Since there is a non-qualified "deny(clippy:all)" up on line 5?

Cargo.toml
```diff
@@ -52,11 +76,12 @@ path = "examples/rust/tcp-ping-pong.rs"
...
-profiler = [ "inetstack/profiler" ]
+catcollar-libos = [ "liburing" ]
+catnip-libos = [ "libdpdk" ]
+libdpdk = [ "dpdk-rs" ]
+mlx4 = [ "dpdk-rs/mlx4" ]
+mlx5 = [ "dpdk-rs/mlx5" ]
+profiler = [  ]
 
 
 [profile.release]
```
BrianZill, line 87: Inetstack's Cargo.toml (which appears to have been folded into here) also had a "profile.dev" for debug builds.  Do we not want that here?  Or is it covered by something somewhere else?

BrianZill (approved):
This looks good to me, so long as my comment on Cargo.toml isn't relevant.

And I did exhaustively review every file (you must be exhausted from fixing up all those namespace use statements).  I didn't flag the inconsistency between specifying things via super vs. self vs. crate since we've never been consistent about that before, although at some point we should probably go through everything and fix those.

anandbonde (approved):
Looks good to me (once Brian's comment on the Cargo.toml is addressed).

Once this is merged, I will attempt to build this combined repo on Windows.

src/rust/lib.rs
ppenna: This is redundant. I shall now get the integration cleaner. Thanks for noting this.

Cargo.toml
ppenna: Well noted. I'm working on this.

PR #298, opened 2022-10-13
[runtime] DemiBuffer Abstraction

BrianZill:
This PR introduces the `DemiBuffer` abstraction.  The intent is to replace the `Buffer` enumeration (along with its `DataBuffer` and `DpdkBuffer` subtypes) with this new abstraction.

The motivation behind this is that the existing `DataBuffer` differed greatly from `DpdkBuffer` in its capabilities, making it difficult to present a uniform `Buffer` interface.   In particular, the `DataBuffer` implementation precluded adding some desired functionality to `Buffer` that is either already present in DPDK or is easy to add to `DpdkBuffer`.  Things like splitting buffers without copying any of the data, chaining buffers together to represent a packet, and queueing buffers using internal linkage fields (to avoid allocations).

More explanation of the `DemiBuffer` type is provided in the demibuffer.rs file comments.

Note that this PR just defines the new `DemiBuffer` type, it doesn't replace the current use of `Buffer`/`DataBuffer`/`DpdkBuffer` in the rest of the codebase.  That will be coming in a future PR.

### DIFF
--- a/src/rust/runtime/memory/buffer/demibuffer.rs
+++ b/src/rust/runtime/memory/buffer/demibuffer.rs
@@ -368,7 +368,7 @@ impl DemiBuffer {
     // Returns the length of the data stored in the DemiBuffer.
     // Note that while we return a usize here (for convenience), the value is guaranteed to never exceed u16::MAX.
     pub fn len(&self) -> usize {
-        self.get_metadata().data_len as usize
+        self.as_metadata().data_len as usize
     }
 
     // Removes `nbytes` bytes from the beginning of the DemiBuffer chain.
@@ -379,7 +379,7 @@ impl DemiBuffer {
         // ToDo: Review having this "match", since MetaData and MBuf are laid out the same, these are equivalent cases.
         match self.get_tag() {
             Tag::Heap => {
-                let metadata: &mut MetaData = self.get_metadata();
+                let metadata: &mut MetaData = self.as_metadata();
                 if nbytes > metadata.data_len {
                     return Err(Fail::new(libc::EINVAL, "tried to remove more bytes than are present"));
                 }
@@ -390,7 +390,7 @@ impl DemiBuffer {
             #[cfg(feature = "libdpdk")]
             Tag::Dpdk => {
                 // Safety: rte_pktmbuf_adj is a FFI, which is safe since we call it with an actual MBuf pointer.
-                if unsafe { rte_pktmbuf_adj(self.get_mbuf(), nbytes) } == ptr::null_mut() {
+                if unsafe { rte_pktmbuf_adj(self.as_mbuf(), nbytes) } == ptr::null_mut() {
                     return Err(Fail::new(libc::EINVAL, "tried to remove more bytes than are present"));
                 }
             },
@@ -407,7 +407,7 @@ impl DemiBuffer {
         // ToDo: Review having this "match", since MetaData and MBuf are laid out the same, these are equivalent cases.
         match self.get_tag() {
             Tag::Heap => {
-                let md_first: &mut MetaData = self.get_metadata();
+                let md_first: &mut MetaData = self.as_metadata();
                 let md_last: &mut MetaData = md_first.get_last_segment();
 
                 if nbytes > md_last.data_len {
@@ -419,7 +419,7 @@ impl DemiBuffer {
             #[cfg(feature = "libdpdk")]
             Tag::Dpdk => {
                 // Safety: rte_pktmbuf_trim is a FFI, which is safe since we call it with an actual MBuf pointer.
-                if unsafe { rte_pktmbuf_trim(self.get_mbuf(), nbytes) } != 0 {
+                if unsafe { rte_pktmbuf_trim(self.as_mbuf(), nbytes) } != 0 {
                     return Err(Fail::new(libc::EINVAL, "tried to remove more bytes than are present"));
                 }
             },
@@ -435,7 +435,7 @@ impl DemiBuffer {
         // Check that a split is allowed.
         match self.get_tag() {
             Tag::Heap => {
-                let md_front: &mut MetaData = self.get_metadata();
+                let md_front: &mut MetaData = self.as_metadata();
                 if md_front.nb_segs != 1 {
                     return Err(Fail::new(libc::EINVAL, "attempted to split multi-segment DemiBuffer"));
                 }
@@ -445,7 +445,7 @@ impl DemiBuffer {
             },
             #[cfg(feature = "libdpdk")]
             Tag::Dpdk => {
-                let mbuf: *mut rte_mbuf = self.get_mbuf();
+                let mbuf: *mut rte_mbuf = self.as_mbuf();
                 // Safety: The `mbuf` dereferences in this block are safe, as it is aligned and dereferenceable.
                 unsafe {
                     if (*mbuf).nb_segs != 1 {
@@ -479,7 +479,7 @@ impl DemiBuffer {
     #[cfg(feature = "libdpdk")]
     pub fn into_mbuf(this: Self) -> Option<*mut rte_mbuf> {
         if this.get_tag() == Tag::Dpdk {
-            let mbuf = Self::get_mbuf(&this);
+            let mbuf = Self::as_mbuf(&this);
             // Don't run the DemiBuffer destructor on this.
             mem::forget(this);
             Some(mbuf)
@@ -508,7 +508,8 @@ impl DemiBuffer {
 
     // Gets the DemiBuffer as a mutable MetaData reference.
     // Note: Caller is responsible for enforcing Rust's aliasing rules for the returned MetaData reference.
-    fn get_metadata(&self) -> &mut MetaData {
+    #[inline]
+    fn as_metadata(&self) -> &mut MetaData {
         // Safety: The call to as_mut is safe, as the pointer is aligned and dereferenceable, and the MetaData struct
         // it points to is initialized properly.
         unsafe { self.get_ptr::<MetaData>().as_mut() }
@@ -516,13 +517,14 @@ impl DemiBuffer {
 
     // Gets the DemiBuffer as a mutable MBuf pointer.
     #[cfg(feature = "libdpdk")]
-    fn get_mbuf(&self) -> *mut rte_mbuf {
+    #[inline]
+    fn as_mbuf(&self) -> *mut rte_mbuf {
         self.get_ptr::<rte_mbuf>().as_ptr()
     }
 
     // Gets a raw pointer to the DemiBuffer data.
     fn data_ptr(&self) -> *mut u8 {
-        let metadata: &mut MetaData = self.get_metadata();
+        let metadata: &mut MetaData = self.as_metadata();
         let buf_ptr: *mut u8 = metadata.buf_addr;
         // Safety: The call to offset is safe, as its argument is known to remain within the allocated region.
         unsafe { buf_ptr.offset(metadata.data_off as isize) }
@@ -533,7 +535,7 @@ impl DemiBuffer {
     // function isn't strictly necessary, as it does the exact same thing as data_ptr() does.
     #[cfg(feature = "libdpdk")]
     fn dpdk_data_ptr(&self) -> *mut u8 {
-        let mbuf: *mut rte_mbuf = self.get_mbuf();
+        let mbuf: *mut rte_mbuf = self.as_mbuf();
         unsafe {
             // Safety: It is safe to dereference "mbuf" as it is known to be valid.
             let buf_ptr: *mut u8 = (*mbuf).buf_addr as *mut u8;
@@ -689,7 +691,7 @@ impl Clone for DemiBuffer {
             },
             #[cfg(feature = "libdpdk")]
             Tag::Dpdk => unsafe {
-                let mbuf_ptr: *mut rte_mbuf = self.get_mbuf();
+                let mbuf_ptr: *mut rte_mbuf = self.as_mbuf();
                 // ToDo: This allocates the clone MBuf from the same MBuf pool as the original MBuf.  Since the clone
                 // never has any direct data, we could potentially save memory by allocating these from a special pool.
                 // Safety: it is safe to dereference "mbuf_ptr" as it is known to point to a valid MBuf.
@@ -810,7 +812,7 @@ impl Drop for DemiBuffer {
             },
             #[cfg(feature = "libdpdk")]
             Tag::Dpdk => {
-                let mbuf_ptr: *mut rte_mbuf = self.get_mbuf();
+                let mbuf_ptr: *mut rte_mbuf = self.as_mbuf();
                 // Safety: This is safe, as mbuf_ptr does indeed point to a valid MBuf.
                 unsafe {
                     // Note: This DPDK routine properly handles MBuf chains, as well as indirect, and external MBufs.
--- a/src/rust/runtime/memory/buffer/demibuffer.rs
+++ b/src/rust/runtime/memory/buffer/demibuffer.rs
@@ -156,21 +156,21 @@ impl MetaData {
 
     // Increments the reference count and returns the new value.
     #[inline]
-    pub fn inc_refcnt(&mut self) -> u16 {
+    fn inc_refcnt(&mut self) -> u16 {
         self.refcnt += 1;
         self.refcnt
     }
 
     // Decrements the reference count and returns the new value.
     #[inline]
-    pub fn dec_refcnt(&mut self) -> u16 {
+    fn dec_refcnt(&mut self) -> u16 {
         self.refcnt -= 1;
         self.refcnt
     }
 
     // Gets the MetaData for the last segment in the buffer chain.
     #[inline]
-    pub fn get_last_segment(&mut self) -> &mut MetaData {
+    fn get_last_segment(&mut self) -> &mut MetaData {
         let mut md: &mut MetaData = self;
         while md.next.is_some() {
             // Safety: The call to as_mut is safe, as the pointer is aligned and dereferenceable, and the MetaData
@@ -214,7 +214,11 @@ impl From<usize> for Tag {
     }
 }
 
-// The DemiBuffer.
+/// The `DemiBuffer`.
+///
+/// This buffer type is designed to be a common abstraction defining the behavior of data buffers in Demikernel.
+/// It currently supports two underlying types of buffers: heap-allocated and DPDK-allocated.  It defines the basic
+/// operations on `DemiBuffer`s; these have the same effect on the data regardless of the underlying buffer type.
 pub struct DemiBuffer {
     // Pointer to the buffer metadata.
     // Stored as a NonNull so it can efficiently be packed into an Option.
@@ -229,8 +233,8 @@ impl DemiBuffer {
     // Constructors
     // ------------
 
-    // Create a new (Heap-allocated) DemiBuffer.
-    //
+    /// Creates a new (Heap-allocated) `DemiBuffer`.
+
     // Implementation Note:
     // This function is replacing the new() function of DataBuffer, which could return failure.  However, the only
     // failure it actually reported was if the new DataBuffer request was for zero size.  A seperate empty() function
@@ -242,7 +246,6 @@ impl DemiBuffer {
     // return an error condition.  But since we call the allocator directly in this implementation, we could now
     // propagate actual allocation failures outward, if we determine that would be helpful.  For now, we stick to the
     // status quo, and assume this allocation never fails.
-    //
     pub fn new(capacity: u16) -> Self {
         // Allocate some memory off the heap.
         let mut temp: NonNull<MetaData> = allocate_metadata_data(capacity);
@@ -284,7 +287,7 @@ impl DemiBuffer {
         }
     }
 
-    // Create a new Heap-allocated DemiBuffer from a slice.
+    /// Create a new Heap-allocated `DemiBuffer` from a byte slice.
     pub fn from_slice(slice: &[u8]) -> Result<Self, Fail> {
         // Note: The implementation of the TryFrom trait (see below, under "Trait Implementations") automatically
         // provides us with a TryInto trait implementation (which is where try_into comes from).
@@ -292,7 +295,7 @@ impl DemiBuffer {
     }
 
     #[cfg(feature = "libdpdk")]
-    // Creates a DemiBuffer from a raw MBuf pointer (*mut rte_mbuf).
+    /// Creates a `DemiBuffer` from a raw MBuf pointer (*mut rte_mbuf).
     // The MBuf's internal reference count is left unchanged (a reference is effectively donated to the DemiBuffer).
     // Note: Must be called with a non-null (i.e. actual) MBuf pointer.  The MBuf is expected to be in a valid state.
     // It is the caller's responsibility to guarantee this, which is why this function is marked "unsafe".
@@ -311,22 +314,24 @@ impl DemiBuffer {
     // Public Functions
     // ----------------
 
+    /// Returns `true` if this `DemiBuffer` was allocated off of the heap, and `false` otherwise.
     pub fn is_heap_allocated(&self) -> bool {
         self.get_tag() == Tag::Heap
     }
 
     #[cfg(feature = "libdpdk")]
+    /// Returns `true` if this `DemiBuffer` was allocated by DPDK, and `false` otherwise.
     pub fn is_dpdk_allocated(&self) -> bool {
         self.get_tag() == Tag::Dpdk
     }
 
-    // Returns the length of the data stored in the DemiBuffer.
+    /// Returns the length of the data stored in the `DemiBuffer`.
     // Note that while we return a usize here (for convenience), the value is guaranteed to never exceed u16::MAX.
     pub fn len(&self) -> usize {
         self.as_metadata().data_len as usize
     }
 
-    // Removes `nbytes` bytes from the beginning of the DemiBuffer chain.
+    /// Removes `nbytes` bytes from the beginning of the `DemiBuffer` chain.
     // Note: If `nbytes` is greater than the length of the first segment in the chain, then this function will fail and
     // return an error, rather than remove the remaining bytes from subsequent segments in the chain.  This is to match
     // the behavior of DPDK's rte_pktmbuf_adj() routine.
@@ -354,7 +359,7 @@ impl DemiBuffer {
         Ok(())
     }
 
-    // Removes `nbytes` bytes from the end of the DemiBuffer chain.
+    /// Removes `nbytes` bytes from the end of the `DemiBuffer` chain.
     // Note: If `nbytes` is greater than the length of the last segment in the chain, then this function will fail and
     // return an error, rather than remove the remaining bytes from subsequent segments in the chain.  This is to match
     // the behavior of DPDK's rte_pktmbuf_trim() routine.
@@ -383,7 +388,7 @@ impl DemiBuffer {
         Ok(())
     }
 
-    // Splits off a new DemiBuffer containing a subset of the data in this DemiBuffer, starting at the given offset.
+    /// Splits off a new `DemiBuffer` containing a subset of the data in this one, starting at the given offset.
     // The data contained in the new DemiBuffer is removed from the original DemiBuffer.
     // Note: the DemiBuffer being split must be a single buffer segment (not a chain) large enough to hold `offset`.
     pub fn split(&mut self, offset: u16) -> Result<Self, Fail> {
@@ -429,7 +434,7 @@ impl DemiBuffer {
         Ok(back_half)
     }
 
-    // Consumes the DemiBuffer, returning the contained MBuf pointer.
+    /// Consumes the `DemiBuffer`, returning the contained MBuf pointer.
     // The returned MBuf takes all existing references on the data with it (the DemiBuffer donates its ref to the MBuf).
     #[cfg(feature = "libdpdk")]
     pub fn into_mbuf(this: Self) -> Option<*mut rte_mbuf> {
@@ -563,7 +568,7 @@ fn free_metadata_data(buffer: NonNull<MetaData>) {
 // Trait Implementations
 // ---------------------
 
-// Clone Trait Implementation for DemiBuffer
+/// Clone Trait Implementation for `DemiBuffer`.
 impl Clone for DemiBuffer {
     fn clone(&self) -> Self {
         match self.get_tag() {
@@ -665,7 +670,7 @@ impl Clone for DemiBuffer {
     }
 }
 
-// De-Reference Trait Implementation for DemiBuffer.
+/// De-Reference Trait Implementation for `DemiBuffer`.
 impl Deref for DemiBuffer {
     type Target = [u8];
 
@@ -689,7 +694,7 @@ impl Deref for DemiBuffer {
     }
 }
 
-// Mutable De-Reference Trait Implementation for DemiBuffer.
+/// Mutable De-Reference Trait Implementation for `DemiBuffer`.
 impl DerefMut for DemiBuffer {
     fn deref_mut(&mut self) -> &mut [u8] {
         // ToDo: Review having this "match", since MetaData and MBuf are laid out the same, these are equivalent cases.
@@ -711,7 +716,7 @@ impl DerefMut for DemiBuffer {
     }
 }
 
-// Drop Trait Implementation for DemiBuffer
+/// Drop Trait Implementation for `DemiBuffer`.
 impl Drop for DemiBuffer {
     fn drop(&mut self) {
         match self.get_tag() {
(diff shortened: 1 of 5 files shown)


src/rust/runtime/memory/buffer/demibuffer.rs
```diff
@@ -0,0 +1,820 @@
...
+        null_mut,
+        NonNull,
+    },
+    slice,
+};
+
+// Cache line size, defined here for code clarity.
+// While this is arguably architecture dependent, and thus should be pulled from some architecture-specific information,
+// this value (64 bytes) is used by all architectures we're likely to care about (and is also assumed by DPDK).
+const CACHE_LINE_SIZE: usize = 64;
```
ppenna, line 67: Could you move this to `pal::arch`? This is an architecture constant. Preferably, rely on the `crate` for this as well.

src/rust/runtime/memory/buffer/demibuffer.rs
```diff
@@ -0,0 +1,820 @@
...
+
+// Buffer Metadata.
+// This is defined to match a DPDK MBuf (rte_mbuf) in order to potentially use the same code for some DemiBuffer
+// operations that currently use identical (but separate) implementations for heap vs DPDK allocated buffers.
+// Fields beginning with an underscore are not directly used by the current DemiBuffer implementation.
+// Should be cache-line aligned and consume 2 cache lines (128 bytes).
+#[repr(C)]
+#[repr(align(64))]
+struct MetaData {
```
ppenna, line 76: Could you add an assertion on this size?

ppenna, line 76: For unused field members please use the `unused` attribute, as we agreed in previous revisions.

src/rust/runtime/memory/buffer/demibuffer.rs
```diff
@@ -0,0 +1,820 @@
...
+    // Hint to compiler that this struct "owns" a MetaData (for safety determinations).  Doesn't consume space.
+    _phantom: PhantomData<MetaData>,
+}
+
+// DemiBuffer tag types.
+// Since our MetaData structure is 64-byte aligned, the lower 6 bits of a pointer to it are guaranteed to be zero.
+// We currently only use the lower 2 of those bits to hold the tag.
+const TAG_MASK: usize = 0x3;
+const TAG_HEAP: usize = 0x1;
+const TAG_DPDK: usize = 0x2;
```
ppenna, line 191: Could you please use an Enumerated Type for this? It will guarantee the right semantics and we can provide the `from` methods.

src/rust/runtime/memory/buffer/demibuffer.rs
```diff
@@ -0,0 +1,820 @@
+// Copyright (c) Microsoft Corporation.
```
ppenna, line 1: Please use the Rust's comment style to document the exposed API.

src/rust/runtime/memory/buffer/demibuffer.rs
```diff
@@ -0,0 +1,820 @@
...
+                        return Err(Fail::new(libc::EINVAL, "attempted to split multi-segment DemiBuffer"));
+                    }
+                    if (*mbuf).data_len < offset {
+                        return Err(Fail::new(libc::EINVAL, "split offset is more bytes than are present"));
+                    }
+                }
+            },
+            _ => {
+                panic!("corrupted DemiBuffer");
+            },
```
ppenna, line 435: Please, log and fail. Someone may simply discard this buffer and continue.

ppenna (changes requested):

ppenna:
@BrianZill please rename you commits to match the style: 

`[module-name] <Enhancement | Feature | Bug Fix | Workd Around>: <Small Comment Title>`

src/rust/runtime/memory/buffer/demibuffer.rs
ppenna, line 1: Could you add custom messages to all assert and debug asserts that would dump the arguments?

`assert(a != b, "a={:?} b={:?}", a, b)`

src/rust/runtime/memory/buffer/demibuffer.rs
BrianZill: Do you mean create a pal::arch?  I only see a pal::linux in the current dev branch.

src/rust/runtime/memory/buffer/demibuffer.rs
BrianZill: See Line 532.

BrianZill: I don't recall such previous discussions.  Note this isn't a case where the compiler doesn't understand that a field is actually being used, this is a case where the field really isn't being used (yet, at least).

A precedent would be the scheduler's WakerPage struct, which has an unused field named "_unused" (it's there for padding) and no annotation.

src/rust/runtime/memory/buffer/demibuffer.rs
ppenna: @BrianZill yes, create another module `pal::arch` and place architectural dependent code there ;-)

src/rust/runtime/memory/buffer/demibuffer.rs
ppenna: Thanks for pointing me to this line. I do have additional points:

- This shouldn't be a debug assert, since it breaks functionality
- This check should be done as soon as possible at initialization time (Demikernel initialization should fail)

ppenna: image

src/rust/runtime/memory/buffer/demibuffer.rs
BrianZill: I originally tried that, but it was problematic for bitwise ORs and ANDs, so I switched to constants.  I'll look into this again to see if it can be done cleanly in Rust.  If it was a proper Rust enum, then I wouldn't need the match "catch-all" cases that you complain about below.

src/rust/runtime/memory/buffer/demibuffer.rs
BrianZill: Is that documented somewhere?

src/rust/runtime/memory/buffer/demibuffer.rs
BrianZill: These are "cannot happen" (outside of hardware failure) situations that are placed here because the compiler insists upon having a "catch-all" case for the match operator.  If memory reads are returning bogus values, then panicking is appropriate.

src/rust/runtime/memory/buffer/demibuffer.rs
BrianZill: The only assertions in this file use the debug_assert_eq! macro, the documentation for which states:

> On panic, this macro will print the values of the expressions with their debug representations.

Is that not sufficient?

ppenna: https://rust-lang.github.io/api-guidelines/

src/rust/runtime/memory/buffer/demibuffer.rs
ppenna: - If you were using the enumeration, that would be not possible.
- To make the compiler happy about non-covered cases on conditional compilation use `#[allow(unreachable_patterns)]` and change `panic()` to `unreachable()`

src/rust/runtime/memory/buffer/demibuffer.rs
ppenna: It is possible to do it, but you need to provide some trait implementations.

src/rust/runtime/memory/buffer/demibuffer.rs
BrianZill: That case is different from this, as I explained in the comment.  In that case, the field is actually being used, just not in a way the compiler thought matters.  Using an underscore there would indicate that the field isn't being used, when it is.

This case, like the WakerPage case I cited, uses the underscore to indicate that a field isn't being used.  This is using the underscore in the way that it is intended to be used.

src/rust/runtime/memory/buffer/demibuffer.rs
ppenna: No. In order to see the values the user has to set the `RUSTBACKTRACE=1` environment variable in advance, which may not be the case always, and even so it may not know how to print arguments.

src/rust/runtime/memory/buffer/demibuffer.rs
BrianZill: It actually doesn't break functionality, it just breaks efficiency, so I think a debug assert is fine.

This whole thing is just a work-around for Rust's (hopefully temporary) lack of proper compile-time assertions (which is how I would have done this in C).  Instead, I check the size at first use of the struct, and only in debug builds so we don't take the hit of the assertion check in production on every allocation call.  I do think the check for this should be co-located with the struct definition rather than in some other unrelated part of the code.

I see only three reasonable courses of action here:
1. Leave it as is.
1. Use one of the [hacky](https://www.reddit.com/r/rust/comments/9z8iqa/compile_time_assertions_trick/) [ways](https://crates.io/crates/static_assertions) to fake compile-time assertions.
1. Remove the check entirely.

ppenna: 4. Use this:

```rust
macro_rules! assert_size_eq {
    ($x:ty, $($xs:ty),+ $(,)?) => {
        const _: fn() = || {
            $(let _ = $crate::_core::mem::transmute::<$x, $xs>;)+
        };
    };
}
```

BrianZill: Good news!  I found out that Rust now provides a less hacky way to have compile-time assertions.  So, I'm going to switch to that.

src/rust/runtime/memory/buffer/demibuffer.rs
BrianZill: Okay, I added an arch.rs under pal and put the CPU_DATA_CACHE_LINE_SIZE constant in it.

src/rust/runtime/memory/buffer/demibuffer.rs
BrianZill: Added two compile-time assertions regarding the MetaData struct:
- Check that MetaData is aligned to a cache line.
- Check that MetaData is two cache lines in size.

And removed the old run-time debug assertion.

src/rust/runtime/memory/buffer/demibuffer.rs
BrianZill: I switched the valid tag values to a new Tag enum.  The TAG_MASK const stays, as it is only used for bitwise operations.  This allows match operations on Tags to be exhaustive, so no "catch-all" is needed.  Needed to implement BitOr and From traits for my use cases.

Wondering if I should have implemented TryFrom instead of From, and put the panic on memory corruption into get_tag instead.

src/rust/runtime/memory/buffer/demibuffer.rs
BrianZill: Switching tag values from consts to an enum removed the need for "catch-all" cases on those match operations, so this is now moot.

src/rust/runtime/memory/buffer/demibuffer.rs
ppenna: Awesome! Thanks

src/rust/runtime/memory/buffer/demibuffer.rs
```diff
@@ -0,0 +1,824 @@
...
+// DemiBuffer type tags.
+// Since our MetaData structure is 64-byte aligned, the lower 6 bits of a pointer to it are guaranteed to be zero.
+// We currently only use the lower 2 of those bits to hold the type tag.
+#[derive(PartialEq)]
+enum Tag {
+    Heap = 1,
+    #[cfg(feature = "libdpdk")]
+    Dpdk = 2,
+}
+const TAG_MASK: usize = 0x3;
```
ppenna, line 193: Could you move this to the namespace of the tag?

Ex:

```rust
impl Tag {
    pub const MASK: usize = 0x3;
}
```

src/rust/runtime/memory/buffer/demibuffer.rs
```diff
@@ -0,0 +1,824 @@
...
+            1 => Tag::Heap,
+            #[cfg(feature = "libdpdk")]
+            2 => Tag::Dpdk,
+            _ => panic!("memory corruption in DemiBuffer pointer"),
+        }
+    }
+}
+
+// The DemiBuffer.
+pub struct DemiBuffer {
```
ppenna, line 215: We need an `into_raw_parts()` function in order to convert and consume this data structure and provide a pair of pointers for constructing the scatter-gather array.

ppenna (commented):
This data structure is awesome. Thanks for the contribution @BrianZill.

Do you think that it would be possible to add some unit-tests for it?

src/rust/runtime/memory/buffer/demibuffer.rs
```diff
@@ -0,0 +1,824 @@
...
+        // Return the new DemiBuffer.
+        DemiBuffer {
+            ptr: tagged,
+            _phantom: PhantomData,
+        }
+    }
+
+    // Create a new Heap-allocated DemiBuffer from a slice.
+    // Note: This is implemented as stand-alone function instead of a conversion (From) Trait in order to be able to
+    // handle the error case where the given slice is larger than a single DemiBuffer can hold.  ToDo: Review this?
```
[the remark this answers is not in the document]
BrianZill: Yes, the ToDo comment was to review this, thanks!

One thing I realized after reading the [API Guidelines doc](https://rust-lang.github.io/api-guidelines) you pointed me at, is that while I was right that I couldn't implement this using `From`, I could still implement a `TryFrom` trait (basically move most of this code there) and then have this `from_slice` function just call that.  What do you think, worth doing?

src/rust/runtime/memory/buffer/demibuffer.rs
```diff
@@ -0,0 +1,824 @@
...
+    pub fn trim(&mut self, nbytes: u16) -> Result<(), Fail> {
+        // ToDo: Review having this "match", since MetaData and MBuf are laid out the same, these are equivalent cases.
+        match self.get_tag() {
+            Tag::Heap => {
+                let md_first: &mut MetaData = self.as_metadata();
+                let md_last: &mut MetaData = md_first.get_last_segment();
+
+                if nbytes > md_last.data_len {
+                    return Err(Fail::new(libc::EINVAL, "tried to remove more bytes than are present"));
+                }
```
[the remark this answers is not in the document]
BrianZill: This check (and similar ones in `adjust` and `split`) either needs to be done separately for each case, or we need to unequivocally state (i.e., rely upon) that our `MetaData` struct and DPDK's `rte_mbuf` struct are laid out the same way (i.e., that the `data_len` value is at the same offset and has the same size for both).

I don't mind going "all in" on `MetaData`/`rte_mbuf` but was hesitant to do that originally because of the lack of compile-time assertions to check that these stay the same (roughly).  Now that I know how to do that reasonably, I think we can do this.  If we do though, we might as well remove all the special case code for DPDK in `adjust`/`trim` and just use our Rust implementations as I wrote them to be exactly equivalent to `rte_pktmbuf_adj` and `rte_pktmbuf_trim` and making the FFI calls out to DPDK code is undoubtedly less performant.

If we leave them separate, then I think it's okay to rely on DPDK's check because it literally does the same thing.

src/rust/runtime/memory/buffer/demibuffer.rs
```diff
@@ -0,0 +1,824 @@
...
+        // Check that a split is allowed.
+        match self.get_tag() {
+            Tag::Heap => {
+                let md_front: &mut MetaData = self.as_metadata();
+                if md_front.nb_segs != 1 {
+                    return Err(Fail::new(libc::EINVAL, "attempted to split multi-segment DemiBuffer"));
+                }
+                if md_front.data_len < offset {
+                    return Err(Fail::new(libc::EINVAL, "split offset is more bytes than are present"));
+                }
```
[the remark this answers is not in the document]
BrianZill: See above comment on the `trim` version of this.

src/rust/runtime/memory/buffer/demibuffer.rs
```diff
@@ -0,0 +1,824 @@
...
+
+    // Create a new Heap-allocated DemiBuffer from a slice.
+    // Note: This is implemented as stand-alone function instead of a conversion (From) Trait in order to be able to
+    // handle the error case where the given slice is larger than a single DemiBuffer can hold.  ToDo: Review this?
+    pub fn from_slice(slice: &[u8]) -> Result<Self, Fail> {
+        // Check size of the slice to ensure a single DemiBuffer can hold it.
+        if slice.len() > u16::MAX as usize {
+            return Err(Fail::new(libc::EINVAL, "slice is larger than a DemiBuffer can hold"));
+        }
+        let amount: u16 = slice.len() as u16;
```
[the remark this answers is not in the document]
BrianZill: Sure, I suppose.

src/rust/runtime/memory/buffer/demibuffer.rs
```diff
@@ -0,0 +1,824 @@
...
+
+            // Point buf_addr at the newly allocated data space (if any).
+            if amount == 0 {
+                // No direct data, so don't point buf_addr at anything.
+                metadata.buf_addr = null_mut();
+            } else {
+                // The direct data immediately follows the MetaData struct.
+                let address: *mut u8 = temp.cast::<u8>().as_ptr();
+                // Safety: The call to offset is safe, as the provided offset is known to be within the allocation.
+                metadata.buf_addr = unsafe { address.offset(size_of::<MetaData>() as isize) };
```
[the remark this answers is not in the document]
BrianZill: Why?  The [source code for `add`](https://doc.rust-lang.org/src/core/ptr/const_ptr.rs.html#871-873) literally just calls `offset` under the covers.  Sure, the compiler will likely optimize away the extra call, but why make it?

src/rust/runtime/memory/buffer/demibuffer.rs
```diff
@@ -0,0 +1,824 @@
...
+            _phantom: PhantomData,
+        }
+    }
+
+    // Create a new Heap-allocated DemiBuffer from a slice.
+    // Note: This is implemented as stand-alone function instead of a conversion (From) Trait in order to be able to
+    // handle the error case where the given slice is larger than a single DemiBuffer can hold.  ToDo: Review this?
+    pub fn from_slice(slice: &[u8]) -> Result<Self, Fail> {
+        // Check size of the slice to ensure a single DemiBuffer can hold it.
+        if slice.len() > u16::MAX as usize {
```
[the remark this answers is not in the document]
BrianZill: Personally, I find it a little ugly to put a return statement inside an assignment, but it seems that Rust will accept this (at least once I add a trailing semicolon), so okay.

src/rust/runtime/memory/buffer/demibuffer.rs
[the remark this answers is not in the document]
ppenna: Nice idea! I would give it a try, since it does not require major changes in your implementation.

src/rust/runtime/memory/buffer/demibuffer.rs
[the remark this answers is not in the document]
ppenna: This is a simple suggestion anyways :)

src/rust/runtime/memory/buffer/demibuffer.rs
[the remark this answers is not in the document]
ppenna: Yeah, it is ugly, but it binds the logic altogether :)

Thanks for noticing the semicolon that I forgot.

src/rust/runtime/memory/buffer/demibuffer.rs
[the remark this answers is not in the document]
ppenna: > If we do though, we might as well remove all the special case code for DPDK in adjust/trim and just use our Rust implementations as I wrote them to be exactly equivalent to rte_pktmbuf_adj and rte_pktmbuf_trim and making the FFI calls out to DPDK code is undoubtedly less performant

Are you suggesting that we wouldn't call `rte_pktmbuf_adj` and `rte_pktmbuf_trim` at all? I wouldn't recommending doing so, because we would be assuming that future versions of DPDK wouldn't break this assumption.

src/rust/runtime/memory/buffer/demibuffer.rs
ppenna: Today, we do need them due to the way that the scatter-gather array code conversion was engineered originally.

I understand that we may not need it in the long run, but then in order to hook-in your major rework on the scatter-gather array layer is required.

Would it be possible that you provide the `into_raw_parts()` now? Then I can work on getting the scatter-gather array layer cleaner, while you use the buffer to get the missing TCP features implemented.

src/rust/runtime/memory/buffer/demibuffer.rs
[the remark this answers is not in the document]
BrianZill: I pushed a commit that changes this.  :-)

src/rust/runtime/memory/buffer/demibuffer.rs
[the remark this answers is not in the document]
BrianZill: Changed and committed.

src/rust/runtime/memory/buffer/demibuffer.rs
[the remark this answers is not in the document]
ppenna: Because you don't need the `isize` casting there.

src/rust/runtime/memory/buffer/demibuffer.rs
BrianZill: I was going to think about this issue when I integrate this to replace Buffer in a near-future PR.

If I understand correctly, you're arguing for an `into_raw_parts()` function that would be specific to the heap allocated DemiBuffer subtype (similar to how `into_mbuf()` is specific to DPDK allocated ones).  I can do this, but would prefer to make that a separate PR.  [Side note: `into_mbuf()` is needed for interfacing with DPDK FFI routines, I believe the sgarray case can be handled without it]

Anyhow, I'd like to make that a separate PR, as I think there may be other things that I need to expose beyond that to interface properly with everything Buffer did (like, does `data_ptr()` need to be public, or is the current `Deref` enough?) and I don't think I'll know that until I actually attempt to replace Buffer.  `DataBuffer`/`DpdkBuffer` didn't expose the same things, and sometimes exposed different ways to do similar things.  I did look at all the places these are used and attempted to provide the necessary functionality, but in case I missed something, I'd prefer to group them in a latter PR.

ppenna: Sure, it is your call, but keep in mind that:

- This is a requirement to plug in your abstraction to the source tree
- The Scatter-Gather Layer will rapidly change in the upcoming weeks to enable LibOS collocation at the PDPIX layer.

src/rust/runtime/memory/buffer/demibuffer.rs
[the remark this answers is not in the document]
BrianZill: Yes, that was what I was suggesting.  In general, I don't think we should attempt to future-proof our code to work with any possible changes to DPDK.  Something like this would be a breaking change for DPDK, so I doubt they would do that.  And if they did, it would likely be coupled with a change to `rte_mbuf`, which would break us similarly.

I propose we leave this as-is for now and revisit it if the DPDK FFI calls turn out to be performance issues.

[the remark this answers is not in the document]
ppenna: I agree to leave as it is. However, I want to note that I am not in favor of dropping DPDK FFI calls for sake or performance, nor assuming that DPDK interface/internals won't work. DPDK is not a standard and it evolves rapidly. There are many features that 3 years ago simply weren't there.

src/rust/runtime/memory/buffer/demibuffer.rs
BrianZill: Yes, I will have this PR done (or at least ready for review) by the time you return from vacation.

As for the Scatter-Gather layer, I'm concerned that the PDPIX layer as specified doesn't provide enough user flexibility and we may need to expose more of the buffer abstraction functionality (like adjust/trim) through to the Scatter-Gather layer than we currently do.  But that's a topic for another discussion.

src/rust/runtime/memory/buffer/demibuffer.rs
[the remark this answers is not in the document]
BrianZill: > * You are adding in the end of the day, so it makes code easier to follow.

I'd argue that the word "offset" is clearer in its intention (that you're staying inside the allocated region -- part of the Safety rules), so I strongly prefer to leave as is.
> * You don't need the casting there, which may hide potential address overflow computation.

The source code for `add` does the very same cast, which I'm surprised about actually, as they're hiding this conversion when it is better to leave it exposed to the programmer.  The [manual page for `add`](https://doc.rust-lang.org/std/primitive.pointer.html#method.add-1) has a Safety section (identical to `offset`'s) making it clear that you really have to think about this before calling the function (which is why it is marked unsafe in the first place).  It also explicitly states that `add` is just a convenience wrapper for `.offset(count as isize)`.  In any event, my exposed cast is clearer than hiding it in a call to `add`, which negates your point.

This discussion has made it pretty clear to me we should use `offset` and not `add`.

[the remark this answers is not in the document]
ppenna: > This discussion has made it pretty clear to me we should use offset and not add.

I disagree.

These two arguments go beyond coding style:

- With `add` you don't need to add a cast in our code, regardless how it is implemented on theirs.
- Casts may introduce bugs silently.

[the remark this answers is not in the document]
BrianZill: > These two arguments go beyond coding style (ie. you you find it or not `offset` expresses better what is going on):
> 
> * With `add` you don't need to add a cast in our code, regardless how it is implemented on theirs.

While true, this is actually an argument for `offset`.  Since `add` is explicitly documented as being a shorthand for `.offset(count as isize)`, what you're saying is that using `add` hides the cast from the programmer (they have to read the documentation to figure out what it is doing).

> * Casts may introduce bugs silently.

Also true.  But again, this is argument for leaving them visible to the programmer, not hiding them in a wrapper call.

It's really clear to me that we should stick with `offset`.  And as a result of this discussion, I'm beginning to think we might want to consider replacing any use of `add` elsewhere in our codebase with `offset` as well.

src/rust/runtime/memory/buffer/demibuffer.rs
BrianZill: Good idea, it's cleaner that way.

Changed and committed.

src/rust/runtime/memory/buffer/demibuffer.rs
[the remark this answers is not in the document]
ppenna: I cannot sign off in favor for casting.

Casting should be avoided, always. If we have an API that does not require casting, we should use it.

Furthermore, good code is the one that does not require one to opening up the API documentation.

src/rust/runtime/memory/buffer/demibuffer.rs
BrianZill: I'm not following.  The documentation for the debug_assert_eq! macro says it does what you're asking for, as I read it.  I don't understand what more you want me to do here.

ppenna: When the assertion failed (regardless of if it is a debug assert or not) the arguments do not get printed, unless the `RUST_BACKTRACE` environment variable is set.

I'm asking you to add these additional messages, anyways. This helps debugging code that is running without the aforementioned variable is set, especially when you hit a bug that is non-deterministic.

src/rust/runtime/memory/buffer/demibuffer.rs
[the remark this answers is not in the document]
anandbonde: In the existing codebase (dev branch), we use "add" more often than "offset":

image

While both methods are not very different and not difficult to understand, I favor simplicity and consistency across the codebase. We should try to follow just 1 chain of thought so that it's the least surprising way of doing things in code. If we have consistency, we can easily search for this pattern and change if we ever decide to go the other way.

Also, add() does not allow negative offsets (not sure if we'd ever need that though).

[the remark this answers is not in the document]
BrianZill: First off, we cast all over the place.  We're expected to know what we're doing when we are casting.  I disagree in general that we should always use a more dangerous API that doesn't require a cast instead of a safer API that does.  Casting is not universally bad; one just needs to know what you're doing when you do it.  This file alone has 50+ direct or indirect casts in it.

Second, the cast involved in this specific case is guaranteed to be safe.  It is casting the result of the `size_of` operator (a `usize`) to an `isize`.  For this to cast to be unsafe, the `size_of` the thing in question (here a `MetaData` struct) would need to exceed 2^31 bytes.  We don't have any structs even close to that, and this one is known to be only 128 bytes.

These two points alone are enough for me to conclude that there is nothing wrong with using `offset` instead of `add`.  For reasons previously stated, I would prefer to use `offset` over `add` even without the below.  But since you mention it...

> You are having these conclusions because you are reading the library implementation, which may change. We should use it as an opaque box.

Third, no, I'm aware of the cast in `add`'s implementation because I'm reading the library _documentation_ which explicitly states that `add` is just a wrapper around `offset` that makes this cast.  It is officially _defined_ as behaving that way.  The opaque box you speak of is around `offset`, not `add`.

> Furthermore, good code is the one that does not require one to opening up the API documentation.

I seriously hope you aren't writing any code using APIs for which you've never read the documentation.  If you're referring to readability of other's code, I would consider this as yet another argument for using `offset` over `add` in this case, as the cast is visible to the reader, and they don't have to open up the API documentation to know about it.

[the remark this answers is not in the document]
BrianZill: My last reply was to Pedro's comment (GitHub doesn't make this easy to follow).

To Anand, I don't think number of lines of use of an API in the existing codebase really matters here.  I also don't think we need to necessarily choose between one or the other.  But if you're concerned about that, then for simplicity/consistency, I think using offset is both more simple and consistent once you realize that both `add` and `sub` are defined (and documented!) as simple wrappers around `offset`.  So you can use `offset` in place of both `add` and `sub` if you want the simplicity/consistency of a single API.

And yes, I have code that uses negative offsets, so we need to either use `offset` or `sub` for that.

src/rust/runtime/memory/buffer/demibuffer.rs
BrianZill: Can you point to an example in the existing codebase where debug_assert_eq! or assert_eq! does what you're requesting?  All I'm seeing is hundreds of cases like mine.

Before we add a new stipulation on our code that will require changing potentially hundreds or thousands of lines of code, we should consider the impact, and whether there might be a better solution.  Such as redefining the assert macros to print the information you want even when RUSTBACKTRACE isn't set.  Or a workaround, like just having all of us set RUSTBACKTRACE=1 in our shell startup scripts (which would suffice for the debug versions of assert, since they're only meant for us anyway).

src/rust/runtime/memory/buffer/demibuffer.rs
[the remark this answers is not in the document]
BrianZill: As I suggested above, I reworked the code that generates a DemiBuffer from an u8 slice into a new implementation of the TryFrom trait.  Implementing TryFrom gives us a TryInto trait implementation "for free" as it is automatically generated from TryFrom.  Then I reimplemented from_slice() as just a call to (TryInfo trait function) try_into().  This seems more in keeping with the Rust way of doing things.

src/rust/runtime/memory/buffer/demibuffer.rs
iyzhang: I think we should separate this decision out in another PR after we discuss it as a group.

src/rust/runtime/memory/buffer/demibuffer.rs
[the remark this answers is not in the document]
iyzhang: I think this is fine. I'm inclined to use the base function, rather than the convenience function, so that reading the code doesn't require following an indirection in the API documentation. While we can treat the API as a black box, the documentation explicitly points to add being a convenience function (and I'm inclined to not use convenience functions but that's just a personal preference that I don't want to impose). My recommendation is that we resolve this PR and make some "best practice" decisions separately and then open a separate housekeeping PR to clean up everything based on the decision.

src/rust/runtime/memory/buffer/demibuffer.rs
BrianZill: We decided in today's group meeting to move forward with this PR and have a group discussion later about using rustdoc to document our code and how we want to do that.

anandbonde (approved):

ppenna (approved):

Issue #274, opened 2022-09-23
[pdpix] Add Support for Windows


anandbonde:
Context
======

Bing is a important collaborator for Demikernel and a path to production. Many Bing services run on Windows. So, having a fully supported Windows port is crucial.

Goal:

- Provide a seamless build mechanism for Linux and Windows (based on Makefiles)
- Build all Demikernel components (including tests and examples) on Windows
- Successfully run all tests Demikernel tests on Windows
- Port all LibOSes to Windows

Proposed Solution
=============

- [x] Support a seamless build mechanism for Linux and Windows
    - [x] Split Makefiles for Linux and Windows (https://github.com/demikernel/demikernel/pull/273)
    - [x] Add Windows Makefiles and ensure that all components build successfully (https://github.com/demikernel/demikernel/pull/309)
- [x] Create a hierarchy of subsystems in the Platform Abstraction Layer (https://github.com/demikernel/demikernel/pull/305)
- [x] Implement a "Catnap"-like solution for Windows (https://github.com/demikernel/demikernel/pull/314)
- [x] Add DPDK support on Windows
- [x] Use PAL in all LibOSes
    - [x] Use PAL in Catnip
- [x] Support Bing team to build and test on Windows

BrianZill:
The hierarchy of subsystems in the Platform Abstraction Layer should be based on subsystem alone.  In other words, there should be a separate file for each subsystem, with organizing directories only for groupings of subsystems.

For example, for the sockets API, everything related to it should go into the same sockets.rs file.  This file should either be directly in the top level (under pal), or in a Networking directory (if we end up with too many files directly under pal).  There should be no "platform" directories.  The sockets.rs file should contain definitions for everything that is platform specific (constants, function names, structures, whatever).  Inside the file, the definitions should be organized by thing, and not by platform.

I.e.,
```
#[cfg(target_os = "windows")]
use windows::Win32::Networking::WinSock;

// Socket Address Types
#[cfg(target_os = "linux")]
pub const AF_INET: i32 = libc::AF_INET;
#[cfg(target_os = "windows")]
pub const AF_INET: i32 = WinSock::AF_INET.0 as i32;

// Socket Types
#[cfg(target_os = "linux")]
pub const SOCK_DGRAM: i32 = libc::SOCK_DGRAM;
#[cfg(target_os = "windows")]
pub const SOCK_DGRAM: i32 = WinSock::SOCK_DGRAM as i32;
#[cfg(target_os = "linux")]
[...]

PR #305, opened 2022-10-14
[winport] Platform Abstraction Layer for Network 

anandbonde:
Partially fixes issue https://github.com/demikernel/demikernel/issues/274.

More PRs will follow:
 (1) Refactor the PAL to make platform/subsystem specific hierarchy in the directory structure.
 (2) Tests and examples on Windows
 (3) Catnap for Windows (makefiles, code, using CatnapW where applicable)

### DIFF
--- a/src/rust/pal/constants.rs
+++ b/src/rust/pal/constants.rs
@@ -1,0 +1,31 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+//==============================================================================
+// Windows constants
+//==============================================================================
+
+#[cfg(target_os = "windows")]
+use windows::Win32::Networking::WinSock;
+
+#[cfg(target_os = "windows")]
+pub const AF_INET: i32 = WinSock::AF_INET.0 as i32;
+
+#[cfg(target_os = "windows")]
+pub const SOCK_STREAM: i32 = WinSock::SOCK_STREAM as i32;
+
+#[cfg(target_os = "windows")]
+pub const SOCK_DGRAM: i32 = WinSock::SOCK_DGRAM as i32;
+
+//==============================================================================
+// Linux constants
+//==============================================================================
+
+#[cfg(target_os = "linux")]
+pub const AF_INET: i32 = libc::AF_INET;
+
+#[cfg(target_os = "linux")]
+pub const SOCK_STREAM: i32 = libc::SOCK_STREAM;
+
+#[cfg(target_os = "linux")]
+pub const SOCK_DGRAM: i32 = libc::SOCK_DGRAM;
--- a/src/rust/pal/data_structures.rs
+++ b/src/rust/pal/data_structures.rs
@@ -1,0 +1,34 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+//==============================================================================
+// Windows data structures
+//==============================================================================
+
+#[cfg(target_os = "windows")]
+use windows::Win32::Networking::WinSock;
+
+#[cfg(target_os = "windows")]
+pub type SockAddrIn = WinSock::SOCKADDR_IN;
+
+#[cfg(target_os = "windows")]
+pub type InAddr = WinSock::IN_ADDR;
+
+#[cfg(target_os = "windows")]
+pub type Socklen = i32;
+
+//==============================================================================
+// Linux data structures
+//==============================================================================
+
+#[cfg(target_os = "linux")]
+pub type SockAddrIn = libc::sockaddr_in;
+
+#[cfg(target_os = "linux")]
+pub type InAddr = libc::in_addr;
+
+#[cfg(target_os = "linux")]
+pub type Socklen = libc::socklen_t;
+
+#[cfg(target_os = "linux")]
+pub type RawFd = ::std::os::unix::prelude::RawFd;
--- a/src/rust/pal/data_structures.rs
+++ b/src/rust/pal/data_structures.rs
@@ -19,6 +19,9 @@ pub type Socklen = i32;
 //==============================================================================
 
 #[cfg(target_os = "linux")]
+pub type SockAddr = libc::sockaddr;
+
+#[cfg(target_os = "linux")]
 pub type SockAddrIn = libc::sockaddr_in;
 
 #[cfg(target_os = "linux")]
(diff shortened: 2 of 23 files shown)


src/rust/pal/constants.rs
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+//==============================================================================
+// Windows constants
+//==============================================================================
```
BrianZill, line 6: I would structure this file differently.  Instead of having a "Windows constants" and "Linux constants" sections, I would group the constants together by type.  So, all of these would go in a "Networking constants" (or even "Sockets constants") section.  And then if this file ever gets too big, we could switch this module's implementation to a directory and have separate files for each section.

src/rust/pal/data_structures.rs
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+//==============================================================================
+// Windows data structures
+//==============================================================================
```
BrianZill, line 6: Same thing here, group the data structures by type, rather than platform.

src/rust/pal/functions.rs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+use crate::pal::data_structures::SockAddrIn;
+
+//======================================================================================================================
+// Windows functions
+//======================================================================================================================
```
BrianZill, line 8: And again here, group functions by type, rather than platform.

src/rust/pal/mod.rs
```diff
@@ -1,4 +1,17 @@
...
+// PAL: Platform Abstraction Layer
+//======================================================================================================================
+
+// This is the platform abstraction layer designed to hide the platform specific implementation details of contants and
+// datastructures. This layer exists because of the differences between the Windows/Linux libc/WinSock implementations
+// of the corresponding Rust libraries.
+
+pub mod constants;
+pub mod data_structures;
+pub mod functions;
```
BrianZill, line 14: I wonder if this is even the right way to do this.  Instead of having constants, data_structures, and functions, I would have a single module for each type of thing, like "networking" or "sockets".  Doing it this way won't scale well.

And the existing "linux" module should be converted to "shared_memory", honestly.

I have an outstanding PR where I'm adding an "arch" module to cover differences in CPU architectures.  It would make no sense for me to go to the "constants" module to add constants for that, and then "data_structures", etc. for example.

BrianZill (changes requested):

src/rust/pal/constants.rs
ppenna: @BrianZill I agree, but I asked @anandbonde to get first this merged in, and then he will re-work this source tree.

src/rust/pal/functions.rs
ppenna: @BrianZill I agree, but I asked @anandbonde to get first this merged in, and then he will re-work this source tree.

src/rust/pal/functions.rs
```diff
@@ -0,0 +1,22 @@
...
+
+#[cfg(target_os = "windows")]
+pub fn get_addr_from_sock_addr_in(sock_addr_in: SockAddrIn) -> u32 {
```
ppenna, line 11: This is copying parameters in. Could you rely on a read-only reference parameter, please?

src/rust/pal/functions.rs
```diff
@@ -0,0 +1,22 @@
...
+    unsafe { sock_addr_in.sin_addr.S_un.S_addr }
+}
+
+//======================================================================================================================
+// Linux functions
+//======================================================================================================================
+
+#[cfg(target_os = "linux")]
+pub fn get_addr_from_sock_addr_in(sock_addr_in: SockAddrIn) -> u32 {
```
ppenna, line 20: This is copying parameters in. Could you rely on a read-only reference parameter, please?

src/rust/pal/mod.rs
ppenna: @BrianZill I agree, but I asked @anandbonde to get first this merged in, and then he will re-work this source tree.

ppenna (changes requested):

src/rust/pal/constants.rs
anandbonde: Agreed, Pedro and I had planned to do this in subsequent PR(s). [As per the next steps listed in the PR description above: (1) Refactor the PAL to make platform specific hierarchy in the directory structure.]

src/rust/pal/data_structures.rs
anandbonde: Agreed, Pedro and I had planned to do this in subsequent PR(s). [As per the next steps listed in the PR description above: (1) Refactor the PAL to make platform/subsystem specific hierarchy in the directory structure.]

src/rust/pal/functions.rs
anandbonde: Agreed, Pedro and I had planned to do this in subsequent PR(s). [As per the next steps listed in the PR description above: (1) Refactor the PAL to make platform/subsystem specific hierarchy in the directory structure.]

src/rust/pal/mod.rs
anandbonde: Agreed, Pedro and I had planned to do this in subsequent PR(s). [As per the next steps listed in the PR description above: (1) Refactor the PAL to make platform/subsystem specific hierarchy in the directory structure.]

src/rust/pal/constants.rs
ppenna: @anandbonde could you create Issues to keep track of this work items?

src/rust/pal/functions.rs
anandbonde: Done

src/rust/pal/functions.rs
anandbonde: Done

ppenna (approved):

src/rust/pal/constants.rs
anandbonde: Done, added this in the issue https://github.com/demikernel/demikernel/issues/274 for tracking.

src/rust/pal/mod.rs
BrianZill: I'll go ahead and agree to this, although it's not at all clear to me why you just can't change this now.  Once badly laid out structure makes it into the source tree, it can be hard to remove.  Using this very pal thing as an example, we currently have this "linux" module that shouldn't be there.  The differentiation by platform should only be in each individual #[cfg(whatever)] line.  I hope you're planning of fixing that too in your subsequent PR.

PR #309, opened 2022-10-25
[winport] Add Windows makefiles

anandbonde:
Partially addresses issue https://github.com/demikernel/demikernel/issues/274.

This PR adds the Makefiles for Windows build. Subsequent PRs will add the CatnapW LibOS for Windows and test/example builds for Windows.

**_Note that the build will NOT work on Windows just with this change. This is a part of a series of PRs to follow._**

### DIFF
--- a/windows.mk
+++ b/windows.mk
@@ -1,3 +1,238 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
+#=======================================================================================================================
+# Default Paths
+#=======================================================================================================================
+
+!ifndef PREFIX
+PREFIX = $(USERPROFILE)
+!endif
+
+!ifndef INSTALL_PREFIX
+INSTALL_PREFIX = $(USERPROFILE)
+!endif
+
+!ifndef LD_LIBRARY_PATH
+LD_LIBRARY_PATH = $(USERPROFILE)/lib
+!endif
+
+#=======================================================================================================================
+# Build Configuration
+#=======================================================================================================================
+
+BUILD = release
+!if "$(DEBUG)" == "yes"
+BUILD = dev
+!endif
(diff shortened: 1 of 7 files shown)


windows.mk
```diff
@@ -1,3 +1,238 @@
...
+
+DEMIKERNEL_LIB = $(BUILD_DIR)/demikernel.dll.lib
+LIBS = $(DEMIKERNEL_LIB)
+
+#=======================================================================================================================
+# Build Parameters
+#=======================================================================================================================
+
+!ifndef LIBOS
+LIBOS = catnapw
```
BrianZill, line 70: Are we having different LibOs names for different platforms?  I would think this would just be "catnap".

anandbonde: I tried to match the source directory name to the LibOs name in this case as the directory structure for Catnap for Linux (_src/rust/catnap_) and Windows (_src/rust/catnapw_) are different. We can try to keep the name same by conditionally compiling different source trees, however I thought this might be a simpler approach.

ppenna (approved):
Looks good for me.

@anandbonde please rebase it with dev and make sure that all regression passes before proceeding with the merge.

windows.mk
BrianZill: I don't have a strong opinion on this one.  We could also have a completely separate catfoo name, I suppose.  This probably all needs to be revisited later when we change things to allow multiple Demikernel I/O stacks to be active at the same time.  Right now, LibOses and I/O stacks are 1:1, which we presumably won't want to be the case in the future.

anandbonde:
> Looks good for me.
> 
> @anandbonde please rebase it with dev and make sure that all regression passes before proceeding with the merge.

Done

Issue #310, opened 2022-10-25
[runtime] Unit Tests for `DemiBuffer` Raise `SIGSEGV`


ppenna:
Description
========

Unit tests for `DemiBuffer` raise `SIGSEGV` and cause the regressions to fail.

How to Reproduce
=============

1. `git checkout 7d29db7e026743bfc328810c17754d736810ff71`
2. `make test-unit-rust LIBOS=catnap UNIT_TEST="--test-threads=1 runtime::memory::buffer::demibuffer"`
3. See error

- Note 1: `LIBOS` may be any of the currently supported ones.
- Note 2: `--test-threads=1` ensures deterministic run for tests
- Note 3: same error can be observed when running full regressions
- Note 4: this bug was actually spotted in our regressions runs:
  - Catnip: https://github.com/demikernel/demikernel/actions/runs/3318536265 
  - Catcollar: https://github.com/demikernel/demikernel/actions/runs/3318536266 
  - Catnap: https://github.com/demikernel/demikernel/actions/runs/3318536267
  - Catpowder: https://github.com/demikernel/demikernel/actions/runs/3318536268

Screenshots
=========

![image](https://user-images.githubusercontent.com/4939789/197777490-ccd22d1b-61e7-4cc1-8a2c-af5fb07c2a89.png)

Expected Behavior
============

These tests should successfully pass.

Updates
======

- Unit tests were disabled in https://github.com/demikernel/demikernel/pull/311.

ppenna:
@BrianZill may we close this issue?

PR #311, opened 2022-10-25
[test] Disable `DemiBuffer` UnitTest

ppenna:
Description
-------------

This is a workaround for https://github.com/demikernel/demikernel/issues/310.

Summary of Changes
-------------------------

- Disabled unit tests for `DemiBuffer`.


anandbonde (approved):
When will this be turned back on? And how will we track it?

ppenna:
> When will this be turned back on? And how will we track it?

@anandbonde we are keeping track of this on issue https://github.com/demikernel/demikernel/issues/310.

PR #312, opened 2022-10-25
[runtime] Bad Deref in `DemiBuffer::drop()` Causes Segmentation Fault

BrianZill:
This PR fixes [Issue#310](https://github.com/demikernel/demikernel/issues/310).

The drop code was referencing `DemiBuffer`'s `ptr` field directly, rather than calling the internal `get_ptr()` method to get the `MetaData` pointer.

One take-away from this experience, is that we can't trust individual unit tests to (indirectly) test `drop` methods.  The DemiBuffer unit tests ran fine when run individually, or even together.  It appears to be non-deterministic as to whether Cargo's test harness keeps the test environment around long enough for destructors to run.

### DIFF
--- a/src/rust/runtime/memory/buffer/demibuffer.rs
+++ b/src/rust/runtime/memory/buffer/demibuffer.rs
@@ -724,7 +724,7 @@ impl Drop for DemiBuffer {
         match self.get_tag() {
             Tag::Heap => {
                 // This might be a chain of buffers.  If so, we'll walk the list.
-                let mut next_entry: Option<NonNull<MetaData>> = Some(self.ptr);
+                let mut next_entry: Option<NonNull<MetaData>> = Some(self.get_ptr());
                 while let Some(mut entry) = next_entry {
                     // Safety: This is safe, as `entry` is aligned, dereferenceable, and the MetaData struct it points
                     // to is initialized.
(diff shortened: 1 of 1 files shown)


src/rust/runtime/memory/buffer/demibuffer.rs
```diff
@@ -724,7 +724,7 @@ impl Drop for DemiBuffer {
         match self.get_tag() {
             Tag::Heap => {
                 // This might be a chain of buffers.  If so, we'll walk the list.
-                let mut next_entry: Option<NonNull<MetaData>> = Some(self.ptr);
+                let mut next_entry: Option<NonNull<MetaData>> = Some(self.get_ptr());
```
anandbonde, line 727: It appears that "ptr" is not supposed to be used directly (only via the get_tag()/get_ptr() functions). It would be good to make it difficult to use it directly (or at least obvious by having a indicative name like **ptr_with_tag or tagged_ptr** etc.).

BrianZill: I was following the lead of the Rust standard library with the naming here.  Several things, such as Rc, Arc, etc., that are implemented using NonNull<Whatever> call that field "ptr".  But then, their pointers aren't tagged.

I think it would be reasonable to rename it tagged_ptr.  Might have even prevented this bug.  I'll do it tomorrow.

ppenna (approved):
Looks good.

Please, consider rebasing your branch and waiting for the regressions to succeed, before proceeding with the merge.

src/rust/runtime/memory/buffer/demibuffer.rs
BrianZill: I renamed it to tagged_ptr.

Issue #306, opened 2022-10-14
[pdpix] Generic `wait()`, `wait_any()` and `timedwait()`


ppenna:
Description
-------------

Currently, each LibOS has its own definition for the following system calls:

- [x] `wait()` 27eeb4063380ac714370a483ebcd976c1156beb4
- [x] `wait_any()` cb5fb1ac8de027ad8bb1c244c11470a9cacc777b
- [x] `timedwait()` cb5fb1ac8de027ad8bb1c244c11470a9cacc777b

The code is replicated, and we don't want this.

Proposed Solution
---------------------

Have a single definition for each one of these system calls at the PDPIX layer.

PR #313, opened 2022-10-26
[pdpix] Generic `wait()`, `wait_any()` and `timedwait()`

ppenna:
Description
-------------

This PR fixes https://github.com/demikernel/demikernel/issues/306.

Summary of Changes
------------------------

- Moved definition of `wait()` to PDPIX layer
- Moved definition of `wait_any()` to PDPIX layer
- Moved definition of `timedwait()` to PDIX layer
- Dropped definition of `wait_any()` from Catnap, Catnip, Catpowder and Catcollar
- Dropped definition of `timedwait()` from Catnap, Catnip, Catpowder and Catcollar

### DIFF
--- a/src/rust/demikernel/libos/mod.rs
+++ b/src/rust/demikernel/libos/mod.rs
@@ -178,8 +178,26 @@ impl LibOS {
 
     /// Waits for an I/O operation to complete or a timeout to expire.
     pub fn timedwait(&mut self, qt: QToken, abstime: Option<SystemTime>) -> Result<demi_qresult_t, Fail> {
-        match self {
-            LibOS::NetworkLibOS(libos) => libos.timedwait(qt, abstime),
+        trace!("timedwait() qt={:?}, timeout={:?}", qt, abstime);
+
+        // Retrieve associated schedule handle.
+        let mut handle: SchedulerHandle = self.schedule(qt)?;
+
+        loop {
+            // Poll first, so as to give pending operations a chance to complete.
+            self.poll();
+
+            // The operation has completed, so extract the result and return.
+            if handle.has_completed() {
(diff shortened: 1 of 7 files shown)


src/rust/demikernel/libos/mod.rs
```diff
@@ -159,22 +160,70 @@ impl LibOS {
...
+
+            // The operation has completed, so extract the result and return.
+            if handle.has_completed() {
+                return Ok(self.pack_result(handle, qt)?);
+            }
         }
     }
 
     /// Waits for an I/O operation to complete or a timeout to expire.
     pub fn timedwait(&mut self, qt: QToken, abstime: Option<SystemTime>) -> Result<demi_qresult_t, Fail> {
```
BrianZill, line 180: Just out of curiosity, do you know what this variable name is supposed to mean?  And is it pronounced "A.B.S. time" or "abs time", or ...?  From its use, I would have thought it would be named "timeout" or "expiration".

BrianZill: Maybe it is short for "absolute timeout" (as opposed to a relative timeout).

BrianZill (approved):
Yea!

src/rust/demikernel/libos/mod.rs
ppenna: @BrianZill when designing the `timedwait()` system call, I sought some inspiration in POSIX [`sem_timedwait()`](https://pubs.opengroup.org/onlinepubs/9699919799/nframe.html), since we did similarly for other calls.

In my understanding, this is an absolute timeout, because the standard has the following sentence:

"The timeout shall expire when the absolute time specified by abstime passes"

What do you think?

Issue #61, opened 2022-03-31
[pdpix] `wait()` and `wait_any()` System Calls Should Take a Time Out


iyzhang:
Description
-------------

We should provide alternative versions for `wait()` and `wait_any()`, which would time out after some time has elapsed and no asynchronous operation has completed. These system calls would enable applications to not block indefinitely.

Proposed Solution
---------------------

Introduce the following alternative versions for `wait()` and `wait_any()`:

- [x] `timedwait()` - Waits for an asynchronous I/O operation to complete or a timeout to expire. (3a2bfcd437aa812c5aeb2f38798e4ea45038bb48)
- [ ] `timedwait_any()` - Waits for the first asynchronous I/O operation in a list to complete or a timeout to expire.

BrianZill:
I'm curious as to why this changed from adding a timeout value to `wait()` and `wait_any()` to having additional `timedwait()` and `timedwait_any()` calls.  Most APIs I'm familiar with have just the call that includes the timeout value (which is optional in some way, depending upon the language could be non-existent or have a special value(s) to mean infinite or immediate).

ppenna:
@BrianZill Linux and POSIX do feature multiple designs for exposing this same mechanism. For instance, the abstime approach that I proposed is used by [`pthread_cond_timedwait()`](https://pubs.opengroup.org/onlinepubs/9699919799/nframe.html).

BrianZill:
These links you've been posting lately to the Open Group don't work properly, at least for me.  They end up opening some landing page, and not the page for the call you're referencing.  I think **[this](https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_cond_timedwait.html)** is the actual link you meant.

Back to the point we're discussing here, the entirety of pthreads is an alternative API.  For our main API, I would prefer that we follow the long-established precedent of `select()`, which doesn't have a separate API call for specifying a timeout.  Instead, `select()` allows one to provide a timeout if one wants one, otherwise you can just not provide that parameter.

PR #315, opened 2022-10-28
[pdpix] `timedwait_any()` System Call

ppenna:
Description
-------------

This PR is a partial fix for https://github.com/demikernel/demikernel/issues/61.

Summary of Changes
------------------------

- Introduced `timedwait_any()`.
- Added bindings for `timedwait_any()`
- Added invalid test for `timedwait_any()`.
- Updated manual pages accordingly.

### DIFF
--- a/man/demi_wait.md
+++ b/man/demi_wait.md
@@ -8,6 +8,8 @@
 
 `demi_wait_any` - Waits for the first asynchronous I/O operation in a list to complete.
 
+`demi_timedwait_any` - Waits for an asynchronous I/O operation to complete or a timeout to expire.
+
 ## Synopsis
 
 ```c
@@ -17,6 +19,8 @@
 int demi_wait(demi_qresult_t *qr_out, demi_qtoken_t qt);
 int demi_timedwait(demi_qresult_t *qr_out, demi_qtoken_t qt, const struct timespec *abstime);
 int demi_wait_any(demi_qresult_t *qr_out, int *ready_offset, demi_qtoken_t qts[], int num_qts);
+int demi_timedwait_any(demi_qresult_t *qr_out, int *ready_offset, const demi_qtoken_t qts[], int num_qts,
+                                  const struct timespec *abstime);
 ```
 
 ## Description
@@ -34,10 +38,17 @@ the calling thread to block (spin) until the timeout `abstime` expires.
 specified by the list of queue tokens `qts` and it has a length of `num_qts`. This system call may cause the calling
 thread to block (spin) indefinitely.
 
+`demi_timedwait_any()` waits for the first asynchronous I/O operation in a set to complete or for the expiration of a
+timeout, whichever happens first. The set of I/O operations is specified by the list of queue tokens `qts` and it has a
+length of `num_qts`.  The `abstime` parameter specifies an absolute timeout in seconds and nanoseconds since the
+Epoch. If the I/O operation has already completed when `demi_timedwait()` is called, then this system call never fails
+with a timeout error, regardless of the value of `abstime`. This system call may cause the calling thread to block
+(spin) until the timeout `abstime` expires.
+
 When `demi_wait()` and `demi_timedwait()` successfully completes, the structure pointed to by `qr_out` is filled in with
-the result value of the I/O operation that has completed. The `demi_wait_any()` system call behaves similarly, but it
-additionally sets `ready_offset` to indicate the index of that I/O operation in the list of queue tokens `qts` that has
-completed.
+the result value of the I/O operation that has completed. The `demi_wait_any()` and `demi_timedwait_any` system calls
+behave similarly, but they additionally set `ready_offset` to indicate the index of that I/O operation in the list of
+queue tokens `qts` that has completed.
 
 The `demi_qresult_t` is defined as follows:
 
--- a/man/demi_wait.md
+++ b/man/demi_wait.md
@@ -15,6 +15,7 @@
 #include <demi/types.h> /* For demi_qresult_t and demi_qtoken_t. */
 
 int demi_wait(demi_qresult_t *qr_out, demi_qtoken_t qt);
+int demi_timedwait(demi_qresult_t *qr_out, demi_qtoken_t qt, const struct timespec *abstime);
 int demi_wait_any(demi_qresult_t *qr_out, int *ready_offset, demi_qtoken_t qts[], int num_qts);
 ```
 
--- a/src/rust/demikernel/bindings.rs
+++ b/src/rust/demikernel/bindings.rs
@@ -523,6 +523,80 @@ pub extern "C" fn demi_wait_any(
 }
 
 //======================================================================================================================
+// timedwait_any
+//======================================================================================================================
+
+#[no_mangle]
+pub extern "C" fn demi_timedwait_any(
+    qr_out: *mut demi_qresult_t,
+    ready_offset: *mut c_int,
+    qts: *mut demi_qtoken_t,
+    num_qts: c_int,
+    abstime: *const libc::timespec,
+) -> c_int {
+    trace!(
+        "demi_timedwait_any() {:?} {:?} {:?} {:?} {:?}",
+        qr_out,
+        ready_offset,
+        qts,
+        num_qts,
+        abstime
+    );
+
+    // Check arguments.
+    if num_qts < 0 {
+        return libc::EINVAL;
+    }
+
+    // Check for invalid timeout.
+    if abstime.is_null() {
+        warn!("abstime is a null pointer");
+        return libc::EINVAL;
+    }
+
+    // Convert timespec to SystemTime.
+    let abstime: Option<SystemTime> = {
+        if abstime.is_null() {
+            None
+        } else {
+            let timeout: Duration = Duration::from_nanos(
+                unsafe { (*abstime).tv_sec } as u64 * 1_000_000_000_ + unsafe { (*abstime).tv_nsec } as u64,
+            );
+            match SystemTime::UNIX_EPOCH.checked_add(timeout) {
+                Some(abstime) => Some(abstime),
+                None => Some(SystemTime::now()),
+            }
+        }
+    };
+
+    // Get queue tokens.
+    let qts: Vec<QToken> = {
+        let raw_qts: &[u64] = unsafe { slice::from_raw_parts(qts, num_qts as usize) };
+        raw_qts.iter().map(|i| QToken::from(*i)).collect()
+    };
+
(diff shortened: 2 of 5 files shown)


man/demi_wait.md
```diff
@@ -33,10 +38,17 @@ the calling thread to block (spin) until the timeout `abstime` expires.
...
+Epoch. If the I/O operation has already completed when `demi_timedwait()` is called, then this system call never fails
+with a timeout error, regardless of the value of `abstime`. This system call may cause the calling thread to block
+(spin) until the timeout `abstime` expires.
+
 When `demi_wait()` and `demi_timedwait()` successfully completes, the structure pointed to by `qr_out` is filled in with
-the result value of the I/O operation that has completed. The `demi_wait_any()` system call behaves similarly, but it
-additionally sets `ready_offset` to indicate the index of that I/O operation in the list of queue tokens `qts` that has
-completed.
+the result value of the I/O operation that has completed. The `demi_wait_any()` and `demi_timedwait_any` system calls
+behave similarly, but it additionally sets `ready_offset` to indicate the index of that I/O operation in the list of
```
BrianZill, line 50: I believe this "it" should be a "they", if I'm understanding the semantics you intend correctly.

src/rust/demikernel/bindings.rs
```diff
@@ -526,6 +526,80 @@ pub extern "C" fn demi_wait_any(
...
+            e.errno
+        },
+    });
+
+    match ret {
+        Ok(ret) => ret,
+        Err(e) => e.errno,
+    }
+}
+
```
BrianZill, line 602: You could avoid a lot of code duplication here in one of two ways:

1.  Have demi_wait_any() just be a wrapper around demi_timedwait_any() that provides a special "infinite" value for the timeout.
2. Don't bother with separate calls at all, just have demi_wait_any() take a timeout value (which can be "infinite", and possibly also "immediate").

src/rust/demikernel/libos/mod.rs
```diff
@@ -201,6 +201,41 @@ impl LibOS {
...
+                handle.take_key();
+            }
+
+            // Check if timeout has expired.
+            if abstime.is_none() || SystemTime::now() >= abstime.unwrap() {
+                return Err(Fail::new(libc::ETIMEDOUT, "timer expired"));
+            }
+        }
+    }
+
```
BrianZill, line 238: Here too, `timedwait_any()` and `wait_any()` share almost all the same code.  `wait_any()` should either be implemented in terms of `timedwait_any`, or the latter should just be renamed to the former and have a single call.

man/demi_wait.md
```diff
@@ -8,14 +8,19 @@
...
 
 ```c
 #include <demi/wait.h>
 #include <demi/types.h> /* For demi_qresult_t and demi_qtoken_t. */
 
 int demi_wait(demi_qresult_t *qr_out, demi_qtoken_t qt);
+int demi_timedwait(demi_qresult_t *qr_out, demi_qtoken_t qt, const struct timespec *abstime);
 int demi_wait_any(demi_qresult_t *qr_out, int *ready_offset, demi_qtoken_t qts[], int num_qts);
+int demi_timedwait_any(demi_qresult_t *qr_out, int *ready_offset, const demi_qtoken_t qts[], int num_qts,
+                                  const struct timespec *abstime);
```
BrianZill, line 23: I still don't like "abstime".  At the very least, it should be "abs_time", although even then it isn't clear what "abs" is short for.

man/demi_wait.md
```diff
@@ -33,10 +38,17 @@ the calling thread to block (spin) until the timeout `abstime` expires.
 specified by the list of queue tokens `qts` and it has a length of `num_qts`. This system call may cause the calling
 thread to block (spin) indefinitely.
 
+`demi_timedwait_any()` waits for the first asynchronous I/O operation in a set to complete or for the expiration of a
+timeout, whichever happens first. The set of I/O operations is specified by the list of queue tokens `qts` and it has a
+length of `num_qts`.  The `abstime` parameter specifies an absolute timeout in seconds and nanoseconds since the
+Epoch. If the I/O operation has already completed when `demi_timedwait()` is called, then this system call never fails
```
BrianZill, line 44: Why is the timeout an absolute rather than a relative time.  The Unix API call our `demi_wait`* calls are most similar to (and will likely be replacing in a least some code we'd like to port over) is `select()`.  And `select` takes an _interval_ timeout, not an absolute one.

I'll also note that `struct timespec` evolved from `struct timeval`, which was initially introduced to the Unix programming world specifically to be the timeout interval for `select`.  So, I think most programmers will expect this to be a time interval, rather than an absolute time.

man/demi_wait.md
ppenna: Thanks for noting this. It is fixed now.

man/demi_wait.md
ppenna: The rationale for using absolute time was to follow other Linux APIs that alternatively to `select()` and `pselect()` rely

The rationale for using an absolute time in the first place:
- Other Linux APIs work with absolute time (ie: `epoll()` and `pthread_cond_timedwait()`) use absolute time.
- `timedwait()` was first implemented using absolute time and I did not aim to diverge from an existing pattern.

In my understanding the approach that you suggested and the one that I proposed work similarly, which of each having their pros and cons.

I happy to take this suggestion to the others and have a discussion. However, I will advocate for handling these semantic changes in a different PR.

man/demi_wait.md
ppenna: Ratione for naming it as "abstime":
- To match POSIX documentation: https://pubs.opengroup.org/onlinepubs/9699919799/nframe.html
- To match Linux API: https://linux.die.net/man/3/pthread_cond_timedwait

src/rust/demikernel/bindings.rs
ppenna: Thanks for noting this.

I had thought about unifying these calls as well. However, I follow a software development approach where I prefer to work with incremental changes in the code base. I believe that this is important specially because some design choices will become clearer with more iterations.

Sme thoughts we still should consolidate:
- What special value should it be, so it matches a natural programming pattern for Demikernel applications?
- Iif we decide to replace `timespec` for an integer that would give us the timeout value in nanoseconds similarly to `nanosleep()`? Then, what would be that speciall value

In any case I agree that this should be reconsidered in the future. I have created an issue to keep track of this on the `wait()` and `timedwait()` system calls. https://github.com/demikernel/demikernel/issues/316

Once `timedwait_any()` is merged in the source tree I create issue for the `timedwait_any()` and `wait_any()` calls as well.

src/rust/demikernel/libos/mod.rs
ppenna: Thanks for noting this.

I had thought about unifying these calls as well. However, I follow a software development approach where I prefer to work with incremental changes in the code base. I believe that this is important specially because some design choices will become clearer with more iterations.

Sme thoughts we still should consolidate:
- What special value should it be, so it matches a natural programming pattern for Demikernel applications?
- Iif we decide to replace `timespec` for an integer that would give us the timeout value in nanoseconds similarly to `nanosleep()`? Then, what would be that speciall value

In any case I agree that this should be reconsidered in the future. I have created an issue to keep track of this on the `wait()` and `timedwait()` system calls. https://github.com/demikernel/demikernel/issues/316

Once `timedwait_any()` is merged in the source tree I create issue for the `timedwait_any()` and `wait_any()` calls as well.

src/rust/demikernel/bindings.rs
BrianZill: As this is the API we expose to the user, and we had previously discussed the importance of getting the API solidified before we had too many users, I think we need to make fixing this a priority.

As for what the special values should be, I suggest we simply follow the well-established precedent of `select()`, where not specifying a timeout (via a null pointer) means "wait forever" and a timeout of zero means "poll and return immediately".

man/demi_wait.md
BrianZill: Um, `timedwait()` made it into the code base via a PR that skirted the proper review process, therefore I don't think we should consider it a precedent, but rather something that we should look at with great suspicion.  My preference would be to reverse that PR and have it go through the proper review process (so we could fix it and this one together).

As to what we should do here, I don't see the point of adding a new API call just to remove it again in a subsequent PR, especially when we can just fix it now by simply adding the timeout value to `demi_wait()` and `demi_wait_any()`.

man/demi_wait.md
BrianZill: That opengroup link is to a generic landing page, so I'm not clear on what it is you're referencing.  And that "Linux API" link is actually describing a POSIX API, as mentioned in the "Prolog" section of that page.  At least that page makes it clear that "abstime" is short for "absolute time".

Should we violate our variable naming rules just to match some obscure POSIX API?  I'm not sure it matters much.    If you prefer it, that's fine.

man/demi_wait.md
BrianZill: An additional argument for interval rather than absolute time is that it inherently more portable.  An absolute time requires an external definition (such as the one we use here, which relies upon the Unix Epoch).  An interval time does not.

Windows expresses absolute time quite differently from traditional Unix and has a different Epoch.  Now there is plenty of originally Unix code that has been ported to Windows that knows how to translate between the two representations, and there is probably even some Windows library support for this (I know at least .NET does).  So, Windows users can certainly make do with a Unix-centric API.  But why tie our API to Unix if we don't need to?

man/demi_wait.md
anandbonde: I like the idea of making the name unambiguous: e.g., absolute_time (it's not too long compared to ready_offset) but is very clear to the reader.

src/rust/demikernel/bindings.rs
anandbonde: I second the idea of prioritizing API exposure in the right way in the first/earliest attempt. But if you have a planned work item that will follow soon, it sounds good to me.

man/demi_wait.md
anandbonde: The simplest API to use is often the best. If using `interval` is going to result in less code/responsibility on the developer (caller's part), then that sounds like a better solution to me. However, we can do it as a separate change (provided that is prioritized to happen soon) and agree upon these conventions for future APIs as well (if applicable). Also, we should document these in the guidelines that we all will agree on going forward.

anandbonde (approved):
Overall, the changes look good to me with action items to follow.

ppenna:
We agreed to address this change differently.

Issue #169, opened 2022-09-05
[runtime] Fail When Attempting to `trim()` Too Many Bytes


ppenna:
Description
--------------

When one attempts to `trim()` more bytes than the target buffer has, the runtime panics.

https://github.com/demikernel/runtime/blob/6817e6ecdb7d7117a59f6d315ed419791762fe36/src/memory/buffer/mod.rs#L52-L58

https://github.com/demikernel/runtime/blob/6817e6ecdb7d7117a59f6d315ed419791762fe36/src/memory/buffer/dpdkbuffer.rs#L53-L58

https://github.com/demikernel/runtime/blob/6817e6ecdb7d7117a59f6d315ed419791762fe36/src/memory/buffer/databuffer.rs#L52-L57

Expected Behavior
---------------------

Instead of panicking, `trim()` should not modify the underlying buffer and return `Err(Fail)`.

Further References
---------------------

- The expected behavior is the same featured by DPDK. For more information refer to: https://doc.dpdk.org/api/rte__mbuf_8h.html#abe1bed439015acec170d1342e8dabb3b.

ppenna:
@BrianZill I believe that this is now fixed, right?

BrianZill:
> @BrianZill I believe that this is now fixed, right?

Yup.

BrianZill:
Fixed by PR #320.

Issue #318, opened 2022-11-07
[test] SharedRingBuffer Tests Can Leave Behind Artifacts That Cause Subsequent Runs to Fail


BrianZill:
Description
========

The `SharedRingBuffer` tests ([`ring_buffer_on_shm_sequential()`](https://github.com/demikernel/demikernel/blob/0ded82656e81cd2eaf3dc874f452569758a4291a/src/rust/collections/shared_ring.rs#L85) and [`ring_buffer_on_shm_concurrent()`](https://github.com/demikernel/demikernel/blob/0ded82656e81cd2eaf3dc874f452569758a4291a/src/rust/collections/shared_ring.rs#L113)) can leave behind named shared memory regions if the test process exits abnormally.  These regions have fixed names of "shm-test-ring-buffer-serial" and "shm-test-ring-buffer-concurrent", respectively.  If the region used by one of these tests is already present at the start of a subsequent test, the test will fail because the `SharedMemory` type calls [shm_open](https://www.man7.org/linux/man-pages/man3/shm_open.3.html) with [both the O_CREAT and O_EXCL flag bits set](https://github.com/demikernel/demikernel/blob/0ded82656e81cd2eaf3dc874f452569758a4291a/src/rust/pal/linux/shm.rs#L86), which returns failure if the requested named region already exists.

Also, when this failure occurs, the test actually panics, rather than fails due to a failed assertion.

How to Reproduce
=============

1. Create one of the name shared memory regions listed above either by running the test and killing it before it can clean up or via some other means.
2. Run the SharedRingBuffer tests again.
3. See error.

Expected Behavior
============

I would expect the tests to be resilient to such situations, via one or more of the following: (a) allowing the test to reuse an existing region of the desired name if already present, (b) explicitly deleting any pre-existing regions with the desired name prior to attempting to recreate them, (c) use a uniquely named region for each test (similar to how some tmp files are named), or other method.

Also, I would expect the test to fail (i.e. have a failed assertion), rather than panic, as the test infrastructure has hooks to properly deal with and report failed assertions, whereas panics mess up the output.

Related Issues
==========

Link related issues here.

Issue #170, opened 2022-09-05
[runtime] Fail When Attempting to `adjust()` Too Many Bytes


ppenna:
Description
--------------

When one attempts to `adjust()` more bytes than the target buffer has, the runtime panics.

https://github.com/demikernel/runtime/blob/6817e6ecdb7d7117a59f6d315ed419791762fe36/src/memory/buffer/mod.rs#L43-L49

https://github.com/demikernel/runtime/blob/6817e6ecdb7d7117a59f6d315ed419791762fe36/src/memory/buffer/dpdkbuffer.rs#L45-L50

https://github.com/demikernel/runtime/blob/6817e6ecdb7d7117a59f6d315ed419791762fe36/src/memory/buffer/databuffer.rs#L43-L49

Expected Behavior
---------------------

Instead of panicking, `adjust()` should not modify the underlying buffer and return `Err(Fail)`.

Further References
---------------------

- The expected behavior is the same featured by DPDK. For more information refer to: https://doc.dpdk.org/api/rte__mbuf_8h.html#a1bbd752194759ce7b419c4998f2e8651.

ppenna:
@BrianZill I believe that this is now fixed, right?

BrianZill:
> @BrianZill I believe that this is now fixed, right?

Yup.

BrianZill:
Fixed by PR #320.

PR #320, opened 2022-11-09
[demikernel] Integrate `DemiBuffer` into the System as `Buffer`/`DataBuffer` Replacement

BrianZill:
This PR replaces the `Buffer` enum (which was backed by the `DataBuffer` and `DPDKBuffer` structures) with the `DemiBuffer` type (which was introduced in PR #298).  It does not yet remove either the `Buffer` or `DataBuffer` structs (that will come in a future PR).  The `DPDKBuffer` also remains, at least for the near future, but is only still used in a small subset of the catnip LibOs-specific code.

This PR (along with #298) also fixes Issues #169 and #170, which were about changing the API of `adjust()` and `trim()` to return an error (instead of panicking) if called with too large a number of bytes to remove from the buffer.

This PR touches a large number of files (57, I believe) but most of the changes are fairly straightforward replacements of one type for the other.  The slightly less straightforward changes are due to handling the new adjust/trim API.  Probably the biggest actual changes in this integration are to the code that handles sgarrays (`demi_sgarray_t`).  Each LibOs has LibOs-specific code to allocate, free, and convert into/out of the `demi_sgarray_t` type (functions `alloc_sgarray()`, `free_sgarray()`, `into_sgarray()`, and `clone_sgarray()`, respectively).  The switch to `DemiBuffer`s has unified this code somewhat (the catnap, catcollar, and catpowder versions are now almost identical), as much of the support for DPDK-allocated buffers is now handled internally to `DemiBuffer`s.  Nevertheless, this is probably the part of this PR that could use the most reviewer eyes.

The other largeish change is to the `DemiBuffer` itself, as I waited until I performed the integration to improve/solidify its API.  The changes to the `DemiBuffer` include:
- `adjust()`, `trim()`, and `split_off()` now take a `usize` (instead of a `u16`) parameter.  This is more in keeping with the spirit of other things in Rust's standard library and fits better with the existing code.
- Added `as_ptr()` for getting a raw `u8` pointer to the data in the `DemiBuffer`.
- Added `into_raw()` and `from_raw()` for converting `DemiBuffer`s into a semi-opaque type we can hand across a C language interface and back.
- Fixed a bug in the `DemiBuffer` implementation where cloning a zero-length `DemiBuffer` resulted in a clone that couldn't later be `drop`ped without crashing.
- Added another unit test to verify the above bug fix, test `into_raw()`/`from_raw()`, and check for other potential drop-related errors.

Finally, this PR adds #[ignore] markers to the `SharedRingBuffer` tests, as I got tired of manually logging into all of the test machines to delete the shared memory regions that these tests tend to leave behind if they or some other test crashes the test infrastructure.  (I filed Issue #318 about this problem) These tests do pass, however, if one runs them after the errant shared memory regions have been deleted.  I'd be happy to remove these markers from this PR if people think they should be added via a separate PR (or not at all).  Once the tests have been fixed, these markers should be removed.

### DIFF
--- a/src/rust/catcollar/mod.rs
+++ b/src/rust/catcollar/mod.rs
@@ -32,8 +32,7 @@ use crate::{
     runtime::{
         fail::Fail,
         memory::{
-            Buffer,
-            DataBuffer,
+            DemiBuffer,
             MemoryRuntime,
         },
         queue::IoQueueTable,
@@ -78,7 +77,7 @@ use ::std::{
 //======================================================================================================================
 
 // Size of receive buffers.
-const CATCOLLAR_RECVBUF_SIZE: usize = 9000;
+const CATCOLLAR_RECVBUF_SIZE: u16 = 9000;
 
 //======================================================================================================================
 // Structures
@@ -241,7 +240,7 @@ impl CatcollarLibOS {
     pub fn push(&mut self, qd: QDesc, sga: &demi_sgarray_t) -> Result<QToken, Fail> {
         trace!("push() qd={:?}", qd);
 
-        let buf: Buffer = self.runtime.clone_sgarray(sga)?;
+        let buf: DemiBuffer = self.runtime.clone_sgarray(sga)?;
 
         if buf.len() == 0 {
             return Err(Fail::new(libc::EINVAL, "zero-length buffer"));
@@ -300,7 +299,7 @@ impl CatcollarLibOS {
     pub fn pop(&mut self, qd: QDesc) -> Result<QToken, Fail> {
         trace!("pop() qd={:?}", qd);
 
-        let buf: Buffer = Buffer::Heap(DataBuffer::new(CATCOLLAR_RECVBUF_SIZE)?);
+        let buf: DemiBuffer = DemiBuffer::new(CATCOLLAR_RECVBUF_SIZE);
 
         // Issue pop operation.
         match self.sockets.get(&qd) {
--- a/src/rust/catcollar/runtime/memory.rs
+++ b/src/rust/catcollar/runtime/memory.rs
@@ -9,8 +9,7 @@ use super::IoUringRuntime;
 use crate::runtime::{
     fail::Fail,
     memory::{
-        Buffer,
-        DataBuffer,
+        DemiBuffer,
         MemoryRuntime,
     },
     types::{
@@ -21,53 +20,60 @@ use crate::runtime::{
 use ::libc::c_void;
 use ::std::{
     mem,
-    slice,
+    ptr::{
+        self,
+        NonNull,
+    },
 };
 
 //==============================================================================
 // Trait Implementations
 //==============================================================================
 
-/// Memory Runtime Trait Implementation for I/O User Ring Runtime
+/// Memory Runtime Trait Implementation for IoUring Runtime
 impl MemoryRuntime for IoUringRuntime {
-    /// Creates a scatter-gather array from a memory buffer.
-    fn into_sgarray(&self, buf: Buffer) -> Result<demi_sgarray_t, Fail> {
-        let len: usize = buf.len();
-        #[allow(unreachable_patterns)]
-        let (dbuf_ptr, sgaseg): (*const u8, demi_sgaseg_t) = match buf {
-            Buffer::Heap(dbuf) => {
-                let (dbuf_ptr, data_ptr): (*const u8, *const u8) = DataBuffer::into_raw_parts(Clone::clone(&dbuf))?;
-                (
-                    dbuf_ptr,
-                    demi_sgaseg_t {
-                        sgaseg_buf: data_ptr as *mut c_void,
-                        sgaseg_len: len as u32,
-                    },
-                )
-            },
-            _ => return Err(Fail::new(libc::EINVAL, "invalid buffer type")),
+    /// Converts a buffer into a scatter-gather array.
+    fn into_sgarray(&self, buf: DemiBuffer) -> Result<demi_sgarray_t, Fail> {
+        // Create a scatter-gather segment to expose the DemiBuffer to the user.
+        let data: *const u8 = buf.as_ptr();
+        let sga_seg: demi_sgaseg_t = demi_sgaseg_t {
+            sgaseg_buf: data as *mut c_void,
+            sgaseg_len: buf.len() as u32,
         };
+
+        // Create and return a new scatter-gather array (which inherits the DemiBuffer's reference).
         Ok(demi_sgarray_t {
-            sga_buf: dbuf_ptr as *mut c_void,
+            sga_buf: buf.into_raw().as_ptr() as *mut c_void,
             sga_numsegs: 1,
-            sga_segs: [sgaseg],
+            sga_segs: [sga_seg],
             sga_addr: unsafe { mem::zeroed() },
         })
     }
 
     /// Allocates a scatter-gather array.
     fn alloc_sgarray(&self, size: usize) -> Result<demi_sgarray_t, Fail> {
-        // Allocate a heap-managed buffer.
-        let dbuf: DataBuffer = DataBuffer::new(size)?;
-        let (dbuf_ptr, data_ptr): (*const u8, *const u8) = DataBuffer::into_raw_parts(dbuf)?;
-        let sgaseg: demi_sgaseg_t = demi_sgaseg_t {
-            sgaseg_buf: data_ptr as *mut c_void,
+        // ToDo: Allocate an array of buffers if requested size is too large for a single buffer.
+
+        // We can't allocate more than a single buffer.
+        if size > u16::MAX as usize {
+            return Err(Fail::new(libc::EINVAL, "size too large for a single demi_sgaseg_t"));
+        }
+
+        // First allocate the underlying DemiBuffer.
+        let buf: DemiBuffer = DemiBuffer::new(size as u16);
+
+        // Create a scatter-gather segment to expose the DemiBuffer to the user.
+        let data: *const u8 = buf.as_ptr();
+        let sga_seg: demi_sgaseg_t = demi_sgaseg_t {
+            sgaseg_buf: data as *mut c_void,
             sgaseg_len: size as u32,
         };
+
+        // Create and return a new scatter-gather array (which inherits the DemiBuffer's reference).
         Ok(demi_sgarray_t {
-            sga_buf: dbuf_ptr as *mut c_void,
+            sga_buf: buf.into_raw().as_ptr() as *mut c_void,
             sga_numsegs: 1,
-            sga_segs: [sgaseg],
+            sga_segs: [sga_seg],
             sga_addr: unsafe { mem::zeroed() },
         })
     }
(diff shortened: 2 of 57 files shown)


src/rust/catcollar/mod.rs
```diff
@@ -78,7 +77,7 @@ use ::std::{
 //======================================================================================================================
 
 // Size of receive buffers.
-const CATCOLLAR_RECVBUF_SIZE: usize = 9000;
+const CATCOLLAR_RECVBUF_SIZE: u16 = 9000;
```
ppenna, line 80: Thanks.

src/rust/catcollar/runtime/memory.rs
```diff
@@ -21,53 +20,60 @@ use crate::runtime::{
...
     }
 
     /// Allocates a scatter-gather array.
     fn alloc_sgarray(&self, size: usize) -> Result<demi_sgarray_t, Fail> {
-        // Allocate a heap-managed buffer.
-        let dbuf: DataBuffer = DataBuffer::new(size)?;
-        let (dbuf_ptr, data_ptr): (*const u8, *const u8) = DataBuffer::into_raw_parts(dbuf)?;
-        let sgaseg: demi_sgaseg_t = demi_sgaseg_t {
-            sgaseg_buf: data_ptr as *mut c_void,
+        // ToDo: Allocate an array of buffers if requested size is too large for a single buffer.
```
ppenna, line 55: I know that we have not agreed on a standard on how to name special comments such as TODOs, FIXMEs and so on. For instance, I've personally been using this latter syntax.

Let us discuss on this with the team, so that we be consistent.

Note that I don't take this this a non-blocker for the PR.

src/rust/catcollar/runtime/memory.rs
```diff
@@ -77,32 +83,80 @@ impl MemoryRuntime for IoUringRuntime {
...
+                ));
+            }
+
+            // Calculate the amount the new starting address is ahead of the old.  And then adjust `clone` to match.
+            let adjustment_amount: usize = sga_data.addr() - clone_data.addr();
+            clone.adjust(adjustment_amount)?;
+
+            // An adjustment above would have reduced clone.len() by the adjustment amount.
+            clone_len -= adjustment_amount;
+            debug_assert_eq!(clone_len, clone.len());
```
ppenna, line 152: This is not on the critical path, why use a debug assert instead of an assert?

src/rust/catnip/runtime/network.rs
```diff
@@ -39,6 +39,10 @@ use crate::timer;
 /// Network Runtime Trait Implementation for DPDK Runtime
 impl NetworkRuntime for DPDKRuntime {
     fn transmit(&self, buf: Box<dyn PacketBuf>) {
+        // ToDo: Consider an important optimization here: If there is data in this packet (i.e. not just headers), and
+        // that data is in a DPDK-owned mbuf, and there is "headroom" in that mbuf to hold the packet headers, just
+        // prepend the headers into that mbuf and save the extra header mbuf allocation that we currently always do.
+
```
ppenna, line 45: Can you open an issue for this?

src/rust/catnip/runtime/network.rs
```diff
@@ -51,7 +55,7 @@ impl NetworkRuntime for DPDKRuntime {
         // Chain body buffer.
 
         // First, allocate a header mbuf and write the header into it.
-        let mut header_mbuf = match self.mm.alloc_header_mbuf() {
+        let mut header_mbuf: DPDKBuffer = match self.mm.alloc_header_mbuf() {
```
ppenna, line 58: I would really enjoy to entirely replace the `HeapBuffer` and `DPDKBuffer` abstractions for just `DemiBuffer`. What are we missing to achieve this? Could you add some TODO for this and create us an issue?

src/rust/catpowder/runtime/network.rs
```diff
@@ -61,16 +59,16 @@ impl NetworkRuntime for LinuxRuntime {
...
         let mut out: [MaybeUninit<u8>; 4096] = [unsafe { MaybeUninit::uninit().assume_init() }; 4096];
         if let Ok((nbytes, _origin_addr)) = self.socket.borrow().recvfrom(&mut out[..]) {
-            let mut ret: ArrayVec<Buffer, RECEIVE_BATCH_SIZE> = ArrayVec::new();
+            let mut ret: ArrayVec<DemiBuffer, RECEIVE_BATCH_SIZE> = ArrayVec::new();
             unsafe {
                 let bytes: [u8; 4096] = mem::transmute::<[MaybeUninit<u8>; 4096], [u8; 4096]>(out);
-                let mut dbuf: Buffer = Buffer::Heap(DataBuffer::from_slice(&bytes));
-                dbuf.trim(4096 - nbytes);
+                let mut dbuf: DemiBuffer = DemiBuffer::from_slice(&bytes).expect("'bytes' should fit");
+                dbuf.trim(4096 - nbytes).expect("'bytes' <= 4096");
```
ppenna, line 71: This is out of the scope of this PR, but If possible, could you create a constant to hold this `4096` bytes?

src/rust/collections/shared_ring.rs
```diff
@@ -82,6 +82,7 @@ mod test {
 
     /// Tests if we succeed to perform sequential accesses to a shared ring buffer.
     #[test]
+    #[ignore]
```
ppenna, line 85: Why are we now ignoring this test?

src/rust/collections/shared_ring.rs
```diff
@@ -110,6 +111,7 @@ mod test {
 
     /// Tests if we succeed to perform concurrent accesses to a shared ring buffer..
     #[test]
+    #[ignore]
```
ppenna, line 114: Why are we now ignoring this test?

src/rust/inetstack/protocols/tcp/segment.rs
```diff
@@ -358,7 +362,8 @@ impl TcpHeader {
             num_options,
             option_list,
         };
-        buf.adjust(data_offset);
+        buf.adjust(data_offset)
+            .expect("buf should contain at least 'data_offset' bytes");
```
ppenna, line 366: Why can't we propagate the error?

src/rust/runtime/memory/buffer/demibuffer.rs
```diff
@@ -617,9 +678,20 @@ impl Clone for DemiBuffer {
...
                         clone.data_off = original.data_off;
                         clone.nb_segs = original.nb_segs;
-                        clone.ol_flags = original.ol_flags | METADATA_F_INDIRECT; // Add indirect flag to clone.
                         clone.pkt_len = original.pkt_len;
                         clone.data_len = original.data_len;
+
+                        // Special case for zero-length buffers.
+                        if original.buf_len == 0 {
+                            debug_assert_eq!(clone.buf_len, 0);
+                            debug_assert_eq!(clone.buf_addr, ptr::null_mut());
```
ppenna, line 687: These are not in the critical-patch, so why should we use a debug_assert?

src/rust/runtime/types/memory.rs
```diff
@@ -33,11 +33,16 @@ pub struct demi_sgaseg_t {
...
 #[derive(Copy, Clone)]
 pub struct demi_sgarray_t {
+    /// Reserved.
     pub sga_buf: *mut c_void,
+    /// Number of segments in this scatter-gather array.
     pub sga_numsegs: u32,
+    /// Scatter-gather array segments.
     pub sga_segs: [demi_sgaseg_t; DEMI_SGARRAY_MAXLEN],
+    /// Source address of the data contained in this scatter-gather array (if present).
     pub sga_addr: SockAddr,
```
ppenna, line 47: Thanks for adding comments here.

ppenna (commented):
Overall, the change looks good.

However, I did leave some minor comments that may be worth discussing with the team.

src/rust/collections/shared_ring.rs
ppenna, line 114: May we schedule a call with the others to discuss this?

src/rust/catnip/runtime/network.rs
BrianZill: I've opened Issue #321 to track this.

src/rust/catcollar/mod.rs
BrianZill: Sure.  :-)

src/rust/catcollar/runtime/memory.rs
BrianZill: Ah yes, this should probably be part of the coding standards document.

src/rust/inetstack/protocols/tcp/segment.rs
BrianZill: This exact error case is checked for and propagated to the caller earlier in the function.  So, this is a "shouldn't happen" case worthy of an assert (which this expect effectively is).

If we get too many of these situations we might want to consider adding adjust_unchecked() and trim_unchecked() calls to the DemiBuffer API to save a few cycles on the fast path.

src/rust/collections/shared_ring.rs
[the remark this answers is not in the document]
BrianZill: Call scheduled.

src/rust/collections/shared_ring.rs
BrianZill: Call scheduled.

src/rust/runtime/memory/buffer/demibuffer.rs
BrianZill: I don't think we know in general when someone calls `clone()` on a buffer whether or not it is in the critical-path.  Offhand, I'd think most of the cases where we could call `clone()` are in the critical-path.  For the specific case here, which only applies to cloning a zero-length buffer, there is only one place (I think) currently in the system where we do that -- when sending a TCP FIN.  I think I'd consider sending a FIN to be more-or-less as critical-path as sending actual data, so I think it makes sense to leave these as `debug_assert`s.  We also don't know if someone might add another clone call or zero-length buffer in the future that would hit the combination of the two in the critical-path.

src/rust/catcollar/runtime/memory.rs
BrianZill: Why isn't this critical path?  Because this block only applies to cases where the user has modified the `sgarray` while they were in control of it?  I believe we decided that such modifications (legitimate ones anyway, that just shrink the valid data space) are considered okay behavior on the part of the user.  So, I'd consider this as critical path as the non-user-modified case, which as part of `clone_sgarray()`, is called on every user send.

src/rust/catpowder/runtime/network.rs
BrianZill: Yes, this is another thing that should be part of the coding standards (no unnamed inline constants that need to match another such value elsewhere).

Did you mean you'd like me to fix this in my PR, or file an Issue about it, or ???

src/rust/catcollar/runtime/memory.rs
ppenna: @anandbonde could you note this down for out next discussion?

src/rust/catpowder/runtime/network.rs
ppenna: > Yes, this is another thing that should be part of the coding standards (no unnamed inline constants that need to match another such value elsewhere).

Agree. @anandbonde could you note this down for out next discussion?

> Did you mean you'd like me to fix this in my PR, or file an Issue about it, or ???

Yes, could you fix this in this PR if possible?

src/rust/catcollar/runtime/memory.rs
ppenna: Ok.

src/rust/runtime/memory/buffer/demibuffer.rs
ppenna: ok.

src/rust/catpowder/runtime/network.rs
BrianZill: This pre-existing problem has been fixed.

src/rust/collections/shared_ring.rs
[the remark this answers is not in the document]
BrianZill: We had the discussion and decided to leave this in this PR.

src/rust/collections/shared_ring.rs
BrianZill: We had the discussion and decided to leave this in this PR.

src/rust/catcollar/runtime/memory.rs
BrianZill: I added a description of this to the "Coding Standards" Channel in Teams and asked Anand to add it to the agenda for the next meeting.

src/rust/catpowder/runtime/network.rs
BrianZill: Also, I added a description of this issue to the "Coding Standards" Channel in Teams and asked Anand to add it to the agenda for the next meeting.

src/rust/collections/shared_ring.rs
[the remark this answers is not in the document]
BrianZill: We apparently didn't all remember the discussion, so I removed this from this PR.

src/rust/collections/shared_ring.rs
BrianZill: We apparently didn't all remember the discussion, so I removed this from this PR.

src/rust/inetstack/protocols/tcp/established/sender.rs
```diff
@@ -314,31 +317,36 @@ impl Sender {
...
         let mut cloned_buf = buf.clone();
         let buf_len: usize = buf.len();
 
         // Pop one byte off the buf still in the queue and all but one of the bytes on our clone.
-        buf.adjust(1);
-        cloned_buf.trim(buf_len - 1);
+        buf.adjust(1).expect("'buf' should contain at least one byte");
+        cloned_buf
+            .trim(buf_len - 1)
+            .expect("'cloned_buf' should contain at least one less than its professed length");
```
anandbonde, line 331: Just curious, did you mean to use "professed" here? Or "processed"? - Both may be valid based on the context, but I may be missing the context here.

src/rust/runtime/memory/buffer/demibuffer.rs
```diff
@@ -872,6 +945,43 @@ mod tests {
         assert!(buf.trim(30).is_err());
     }
 
+    // Test cloning, raw conversion, and zero-size buffers.
+    #[test]
+    fn advanced() {
```
anandbonde, line 950: What is the meaning of "advanced" in this context? Can it be replaced with a more descriptive name?

anandbonde (approved):
Overall, looks good to me. I left a few comments, but they are not blockers in my perspective. Regarding using the actual values, if it is useful, but will result into a big change, we can create an issue and track it separately.

src/rust/runtime/memory/buffer/demibuffer.rs
BrianZill: I meant in the sense that this unit test is testing more "advanced" features of the DemiBuffer than the "basic" test that precedes it.

I thought about naming it after a list of the features being tested, like I do for the "split_off" test, but "split_off" only tests that one new feature (meaning it can have a short name), while this one would need to be called something unwieldly, like "cloning_raw_conversion_zero_sized_buffers_drop".  And as this tests a combination of features, I thought it made sense to write the test this way, instead of breaking it into multiple tests.  I figure if a test fails, the short name is enough to let people find the failing test, which is what really matters.

src/rust/inetstack/protocols/tcp/established/sender.rs
BrianZill: I did mean "professed".  I was just being cheeky, I suppose I could have just used "claimed", but this wording was more fun.

src/rust/inetstack/protocols/tcp/established/sender.rs
```diff
@@ -314,31 +317,36 @@ impl Sender {
...
         let buf_len: usize = buf.len();
 
         if buf_len > max_bytes {
-            let mut cloned_buf = buf.clone();
+            let mut cloned_buf: DemiBuffer = buf.clone();
 
-            buf.adjust(max_bytes);
-            cloned_buf.trim(buf_len - max_bytes);
+            buf.adjust(max_bytes)
+                .expect("'buf' should contain at least 'max_bytes'");
```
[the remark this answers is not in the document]
BrianZill: You mean to convert the actual value to text?  I don't think you can do that trivially with `expect()`, as it just takes a `&str` (a borrowed string slice).  To have it print out the value of max_bytes, you'd have to have code outside of here prepare a String or other buffer, fill it with the message you wanted to print on error, and then pass that buffer as a slice to `expect()`.  Which would be a lot of extra code to execute in the non-error case just to pretty up this error, and also make the code harder to read, I think.

If `expect()` took a format string, then yes, I agree it would be better to print the value of `max_bytes`, rather than the name.

Issue #323, opened 2022-11-10
[pdpix] Multiplex `demi_wait()` on Top of `demi_timedwait()`


ppenna:
Context
---------

The `demi_wait()` system call is a specialization of `demi_timedwait()` - it is equivalent of waiting for an indefinite amount of time.

Proposed Solution
---------------------

We should multiplex `demi_wait()` on top of `demi_timedwait()`.

This may be possible by enforcing the following semantic on the `abstime` parameter on the latter system call:
- If `abstime` has a null value, then the `demi_timedwait()` should block until the concerned I/O queue operation completes.
- If `abstime` has a non-null value it should behave as before.

BrianZill:
I would take this further and add an _interval_ timeout value to both `demi_wait()` and `demi_wait_any()` as we discussed in the meeting we had about this issue.  With the same semantics that a missing _interval_ means "wait forever".  And a zero _interval_ means to poll once and return immediately.

You can still multiplex all of these on a single underlying implementation by adding _now_ (in absolute time) to the _interval_ (if present) in the `demi_wait()` and `demi_wait_any()` cases before calling `demi_timedwait_any()` (which would then be the common case of all the others).

ppenna:
We agreed to address this issue differently. See https://github.com/demikernel/demikernel/pull/333 for more details.

PR #324, opened 2022-11-10
[pdpix] Multiplex `demi_wait()` on Top of `demi_timedwait()`

ppenna:
Description
-------------

This is a fix for https://github.com/demikernel/demikernel/issues/323.

Summary of Changes
-------------------------

- Changed semantics of `demi_timedwait()` to enable indefinite wait
- Multiplex implementation of `demi_wait()` on top of `demi_timedwait()`
- Updated manual pages accordinly

### DIFF
--- a/man/demi_wait.md
+++ b/man/demi_wait.md
@@ -22,7 +22,11 @@ int demi_wait_any(demi_qresult_t *qr_out, int *ready_offset, demi_qtoken_t qts[]
 ## Description
 
 `demi_wait()` waits for the completion of the asynchronous I/O operation associated with the queue token `qt`. This
-system call may cause the calling thread to block (spin) indefinitely.
+system call may cause the calling thread to block (spin) indefinitely. This `demi_wait(qr, qt)` call is equivalent to:
+
+```c
+demi_timedwait(qr, qt, NULL);
+```
 
 `demi_timedwait()` waits for the completion of the asynchronous I/O operation associated with the queue token `qt` or
 for the expiration of a timeout, whichever happens first. If the `abstime` parameter is non null, then it is taken as an
@@ -115,7 +119,6 @@ On error, one of the following positive error codes is returned:
 - `EINVAL` - The `qt` argument refers to an invalid queue token.
 - `EINVAL` - The `num_qts` argument has an invalid size.
 - `EINVAL` - The `qts` argument contains an invalid queue token.
-- `EINVAL` - The `abtime` argument does not point to a valid structure.
 - `ETIMEDOUT` - The system call timed out before an I/O operation was completed.
 
 ## Conforming To
--- a/man/demi_wait.md
+++ b/man/demi_wait.md
@@ -15,6 +15,7 @@
 #include <demi/types.h> /* For demi_qresult_t and demi_qtoken_t. */
 
 int demi_wait(demi_qresult_t *qr_out, demi_qtoken_t qt);
+int demi_timedwait(demi_qresult_t *qr_out, demi_qtoken_t qt, const struct timespec *abstime);
 int demi_wait_any(demi_qresult_t *qr_out, int *ready_offset, demi_qtoken_t qts[], int num_qts);
 ```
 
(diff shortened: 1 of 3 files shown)


BrianZill:
Not to beat a dead horse here, but I thought we had decided in the meeting we had about this issue that the wait time was going to be expressed as an interval, rather than an absolute time?

No timespec -> wait forever.
Zero value timespec -> poll and return immediately.

ppenna:
> Not to beat a dead horse here, but I thought we had decided in the meeting we had about this issue that the wait time was going to be expressed as an interval, rather than an absolute time?
> 
> No timespec -> wait forever. Zero value timespec -> poll and return immediately.
> 
> Or are you proposing to leave demi_timedwait() for those who want an absolute timeout value, and add the interval timeout parameter to demi_wait()? If so, you left out adding the interval timeout parameter to demi_wait().
> 
> I added some more comments on this to Issue #323.

@BrianZill, I do not recall that we reached this conclusion.

The current implementation works as follows:
- `demi_timedwait()` takes an absolute timeout parameter
- A null value for the timeout parameter means "poll indefinetely"
- A zero value for the timeout parameter means "poll once and return"

My argument for using an absolute timeout parameter was:
- This is the semantics used for [`pthread_cond_timedwait()`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_cond_timedwait.html)

Could you please recall me about your arguments?

BrianZill:
> > Not to beat a dead horse here, but I thought we had decided in the meeting we had about this issue that the wait time was going to be expressed as an interval, rather than an absolute time?
> > No timespec -> wait forever. Zero value timespec -> poll and return immediately.
> > Or are you proposing to leave demi_timedwait() for those who want an absolute timeout value, and add the interval timeout parameter to demi_wait()? If so, you left out adding the interval timeout parameter to demi_wait().
> > I added some more comments on this to Issue #323.
> 
> @BrianZill, I do not recall that we reached this conclusion.
> 
> The current implementation works as follows:
> 
> * `demi_timedwait()` takes an absolute timeout parameter
> * A null value for the timeout parameter means "poll indefinetely"
> * A zero value for the timeout parameter means "poll once and return"
> 
> My argument for using an absolute timeout parameter was:
> 
> * This is the semantics used for [`pthread_cond_timedwait()`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_cond_timedwait.html)
> 
> Could you please recall me about your arguments?

In the meeting, we explicitly decided to add an interval timeout parameter to demi_wait() and demi_wait_any().  I don't recall if we explicitly stated to do away with demi_timedwait(), or if that was left implicit.

The arguments for using an interval timeout parameter are multiple:
- It's the established standard used by the select() call.
- Intervals are system-independent, while absolute times are dependent upon some Epoch (which is different for Windows and Unix).  It would be better for our API to remain system-independent.
- Interval timeout values are friendlier to the programmer.  The logic the programmer wants is almost always to wait some interval amount of time.  If the API requires an absolute time, the programmer has to add code to get the current time and add the interval amount of time they want to wait to it before calling the API.

anandbonde:
> > > Not to beat a dead horse here, but I thought we had decided in the meeting we had about this issue that the wait time was going to be expressed as an interval, rather than an absolute time?
> > > No timespec -> wait forever. Zero value timespec -> poll and return immediately.
> > > Or are you proposing to leave demi_timedwait() for those who want an absolute timeout value, and add the interval timeout parameter to demi_wait()? If so, you left out adding the interval timeout parameter to demi_wait().
> > > I added some more comments on this to Issue #323.
> > 
> > 
> > @BrianZill, I do not recall that we reached this conclusion.
> > The current implementation works as follows:
> > 
> > * `demi_timedwait()` takes an absolute timeout parameter
> > * A null value for the timeout parameter means "poll indefinetely"
> > * A zero value for the timeout parameter means "poll once and return"
> > 
> > My argument for using an absolute timeout parameter was:
> > 
> > * This is the semantics used for [`pthread_cond_timedwait()`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_cond_timedwait.html)
> > 
> > Could you please recall me about your arguments?
> 
> In the meeting, we explicitly decided to add an interval timeout parameter to demi_wait() and demi_wait_any(). I don't recall if we explicitly stated to do away with demi_timedwait(), or if that was left implicit.
> 
> The arguments for using an interval timeout parameter are multiple:
> 
> * It's the established standard used by the select() call.
> * Intervals are system-independent, while absolute times are dependent upon some Epoch (which is different for Windows and Unix).  It would be better for our API to remain system-independent.
> * Interval timeout values are friendlier to the programmer.  The logic the programmer wants is almost always to wait some interval amount of time.  If the API requires an absolute time, the programmer has to add code to get the current time and add the interval amount of time they want to wait to it before calling the API.

Here are my thoughts:

**Single/multiple APIs** - I prefer single API

Why?
1. A single API with an optional timeout parameter is better, because it's easier for the programmer to adapt/change if needed.
2. With multiple API, the programmer has to change the function name and add the timeout parameter. With single API, only the parameter changes. So the code is easier to diff/review/grasp in future iterations for the programmer and their team.
3. Having a narrow interface (fewer exposed points) helps us to be flexible within the boundaries of Demikernel and not have too many contracts that we sign up for facing outside.

**Interval/Absolutel time** - I prefer interval

Why?
1. An interval is easier to reason about and programmer needs no computation/logic to handle it.
4. Having the timeout computed in just 1 area of the code is better (localizes complexity).
5. +1 for system-independent implementations.

ppenna:
We agreed to address this change differently.

Issue #326, opened 2022-11-11
[catnip] Missing Check for Return Value of `rte_eal_init()`


ppenna:
Description
========

When initializing DPDK we miss to check for the return value of `rte_eal_init()`.

How to Reproduce
=============

https://github.com/demikernel/demikernel/blob/b503584dc526c858bbe4117378a103d373007d93/src/rust/catnip/runtime/mod.rs#L177-L184

Screenshots
=========

Expected Behavior
============

We should check the return value of `rte_eal_init()`

PR #327, opened 2022-11-11
[catnip] Missing Check for Return Value of `rte_eal_init()`

ppenna:
Description
-------------

This PR fixes https://github.com/demikernel/demikernel/issues/326.

Summary of Changes
------------------------

- Added a check for the return value of `rte_eal_init()`

### DIFF
--- a/src/rust/catnip/runtime/mod.rs
+++ b/src/rust/catnip/runtime/mod.rs
@@ -174,8 +174,10 @@ impl DPDKRuntime {
         std::env::set_var("MLX5_SINGLE_THREADED", "1");
         std::env::set_var("MLX4_SINGLE_THREADED", "1");
         let eal_init_refs = eal_init_args.iter().map(|s| s.as_ptr() as *mut u8).collect::<Vec<_>>();
-        unsafe {
-            rte_eal_init(eal_init_refs.len() as i32, eal_init_refs.as_ptr() as *mut _);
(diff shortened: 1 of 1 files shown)


src/rust/catnip/runtime/mod.rs
```diff
@@ -174,8 +174,9 @@ impl DPDKRuntime {
         std::env::set_var("MLX5_SINGLE_THREADED", "1");
         std::env::set_var("MLX4_SINGLE_THREADED", "1");
         let eal_init_refs = eal_init_args.iter().map(|s| s.as_ptr() as *mut u8).collect::<Vec<_>>();
-        unsafe {
-            rte_eal_init(eal_init_refs.len() as i32, eal_init_refs.as_ptr() as *mut _);
+        let ret: libc::c_int = unsafe { rte_eal_init(eal_init_refs.len() as i32, eal_init_refs.as_ptr() as *mut _) };
+        if ret < 0 {
+            bail!("EAL initialization failed {:?}", ret);
```
BrianZill, line 179: Is there any way we can get access to `rte_errno` here?  And print that instead of the return value?  `rte_eal_init()` is documented as returning -1 on error, and "`rte_errno` is set to a value indicating the cause for failure".

ppenna: Thanks for noting this. I've updated the fix for getting and printing the value of `rte_errno`.

BrianZill: Awesome!  Thanks.

BrianZill (approved):
Looks good now, thanks!

PR #328, opened 2022-11-16
[collections] Mark Flaky `SharedRingBuffer` Unit Tests as `#[ignore]`

BrianZill:
This PR marks the two flaky unit tests for `SharedRingBuffer` as `#[ignore]` until they can be fixed to not corrupt the test infrastructure.

See Issue #318.

### DIFF
--- a/src/rust/collections/shared_ring.rs
+++ b/src/rust/collections/shared_ring.rs
@@ -81,6 +81,7 @@ mod test {
     const RING_BUFFER_CAPACITY: usize = 4096;
 
     /// Tests if we succeed to perform sequential accesses to a shared ring buffer.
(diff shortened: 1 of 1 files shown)


anandbonde (approved):
Looks good.

How are we prioritizing addressing these flaky tests? It would be good to address these as soon as possible so that we don't have the untested functionality (one that these tests were intended to cover).

ppenna:
> Looks good.
> 
> How are we prioritizing addressing these flaky tests? It would be good to address these as soon as possible so that we don't have the untested functionality (one that these tests were intended to cover).

@BrianZill I have assigned you to the issue that you created so it remembers you to share the priority on this.

PR #333, opened 2022-11-21
[pdpix] Timeout Interval Parameter to `demi_wait()` and `demi_wait_any()`

BrianZill:
This PR is a fix for Issue #61.

- Added an interval timeout parameter to `demi_wait()` and `demi_wait_any()`.
- Updated all test and example code in the demikernel repo that uses these call(s) to supply the additional parameter.
- Updated the manual page.
- Changed the internal Rust `wait()` call's implementation to call `wait_any()`.

Note: I originally planned to re-implement `demi_wait()` and `demi_wait_any()` on top of `demi_timedwait()` (i.e. convert the wait interval to an absolute timeout) but had to abort that plan once I noticed in the documentation that `SystemTime` is not guaranteed to be monotonically increasing, and thus isn't suitable for interval timing.  (While the implementation of these is technically system-dependent, Rust expects `SystemTime` to be implemented using the real-time clock, which can be adjusted by NTP or the user, and thus can go backwards.  `Instant` is expected to be implemented using tick-counts or something similar, which is guaranteed to be monotonically increasing)  So I implemented `demi_wait_any()` using `Instant`s, and `demi_wait()` is now implemented using `demi_wait_any()`.

I left `demi_timedwait()` in place until we have time to discuss what to do about it.

### DIFF
--- a/examples/c/tcp-ping-pong.c
+++ b/examples/c/tcp-ping-pong.c
@@ -56,7 +56,7 @@ static int accept_wait(int qd)
     assert(demi_accept(&qt, qd) == 0);
 
     /* Wait for operation to complete. */
-    assert(demi_wait(&qr, qt) == 0);
+    assert(demi_wait(&qr, qt, NULL) == 0);
 
     /* Parse operation result. */
     assert(qr.qr_opcode == DEMI_OPC_ACCEPT);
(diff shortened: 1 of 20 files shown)


ppenna (approved):
Thanks for working on this.

IMO, now we shall drop `timedwait()`

iyzhang (approved):
Looks good to me. Just so I understand, we are now using an interval time out as defined by two Instants. Is that correct?

ppenna:
I am merging this to unblock me.

PR #336, opened 2022-11-22
[pdpix] Use trace Logging Level to Propagate Unexpected Conditions

ppenna:
Description
-------------

This is a partial fix for [pdpix] Use trace Logging Level to Propagate Unexpected Conditions.

Summary of Changes
------------------------
- Changed logging messages on PDPIX to `trace` level.
  - These do not concern the point where the unexpected situation happened.
- Renamed socket queue descriptors in PDPIX layer to `sockqd` to match the used convention in the file.
- Fixed some types in the API docs in the PDPIX layer.
- Added missing types in `wait()`.


BrianZill (approved):
Looks good!

Issue #337, opened 2022-11-22
[examples] Support Multiple Clients


ppenna:
Context
---------

The following examples should be improved to support multiple simultaneous clients:

- [x] `tcp-dump` 4ed45892cd68cbf45c9f02e55ac143c5001d1d1e
- [x] `tcp-echo` f764524dfb5f960bf45599c4dc642999063e70e1

PR #338, opened 2022-11-22
[examples] Support Multiple Clients in `tcp-echo`

ppenna:
Description
-------------

This is a fix for https://github.com/demikernel/demikernel/issues/337.

Summary of Changes
-------------------------

- Changed `tcp-echo` to support multiple connections.

### DIFF
--- a/examples/rust/tcp-echo.rs
+++ b/examples/rust/tcp-echo.rs
@@ -93,7 +93,7 @@ impl ProgramArguments {
                 Arg::new("bufsize")
                     .long("bufsize")
                     .value_parser(clap::value_parser!(String))
-                    .required(true)
+                    .required(false)
                     .value_name("SIZE")
                     .help("Sets buffer size"),
             )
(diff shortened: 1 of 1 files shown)


examples/rust/tcp-echo.rs
```diff
@@ -271,22 +271,28 @@ impl Application {
     /// Runs the target echo server.
     pub fn run_server(&mut self) -> ! {
         let start: Instant = Instant::now();
+        let mut nclients: usize = 0;
```
BrianZill, line 274: Since you also have the `clients` `Vec`, you wouldn't need this, as you could use `clients.len()`, but this is fine.

BrianZill (approved):
This is fine.

There's a lot of boilerplate in these TCP server applications (I'm seeing the same thing with the http-server) that we could possibly encapsulate in one place once we figure out how best to do things.  But that can wait for future work.

PR #339, opened 2022-11-23
[arp] Add Remove Function

iyzhang:
This PR add a remove function for the hashttlcache. Due to the removed value being placed in the graveyard, we do not currently return the removed value as well because that would require us to clone it. I believe this is incorrect behavior but we're unclear on what the graveyard's functionality is or if we will use it going forward so I will just leave this fix as is for now.

### DIFF
--- a/src/rust/inetstack/collections/hashttlcache/mod.rs
+++ b/src/rust/inetstack/collections/hashttlcache/mod.rs
@@ -114,8 +114,19 @@ where
     }
 
     /// Removes an entry from the cache.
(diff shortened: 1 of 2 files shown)


ppenna (approved):
Okay.

If possible, consider working on the following when merging:

- Squash all commits into a single one, since they all refer to the same piece of code and fix and we want to present them as a unit in the commit history.
- Reword your commits to match "... **E**nhancement ..." instead of "... enhancement ..." to ease regular expression grepping.

Issue #194, opened 2022-04-29
[ethernet] Hardware Address Filtering Doesn't Accept Multicast


BrianZill:
Description
========

In the top-level src/lib.rs, the do_receive routine attempts to filter out packets sent to "bad" hardware addresses.  Currently, this only allows for a single local ethernet address, and the broadcast ethernet address.  This should also accept a number of multicast addresses, or all multicast addresses.  For our purposes, all multicast addresses would be fine.

I have currently commented out the "physical destination address mismatch" error return until this is fixed.

Note: the easiest way to fix this would be to add a "is_multicast" method to our MacAddress type, which is defined in the "runtime" repo,  in src/network/types/macaddr.rs, similar to how "is_broadcast" is defined there.

BrianZill:
Fixed in PR #351.

PR #351, opened 2022-11-30
[inetstack] Accept Ethernet Multicast Packets

BrianZill:
This PR fixes Issue #194.

Basically, we have a filter at the Ethernet layer that is intended to drop packets not actually addressed to us (Ethernet MAC address).  In the past, we accepted incoming packets addressed to our actual EUI-48, and also ones sent to the broadcast address (needed for ARP).  But we dropped packets addressed to a multicast address that our NIC was listening on (after printing a warning).

This PR now allows multicast packets to be received by the Ethernet layer of our stack (they still have to pass the IP address checks of course).

Note: There used to be a [linked Issue in a different repo](https://github.com/demikernel/runtime/issues/17) (back when we had these separate) to add the `is_multicast()` functionality.  I added that in a separate PR before the repos were recombined.

### DIFF
--- a/src/rust/inetstack/mod.rs
+++ b/src/rust/inetstack/mod.rs
@@ -565,10 +565,11 @@ impl InetStack {
         timer!("inetstack::engine::receive");
         let (header, payload) = Ethernet2Header::parse(bytes)?;
         debug!("Engine received {:?}", header);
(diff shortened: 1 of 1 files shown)


ppenna (approved):
Good.

PR #352, opened 2022-12-02
[catnip] Use PAL in Catnip

anandbonde:
The Platform Abstraction Layer (PAL) implements the OS specific artifacts for the LibOSes. The aim is to isolate these implementation details at a single place (i.e. the PAL).

Partially addresses issue https://github.com/demikernel/demikernel/issues/274.

### DIFF
--- a/src/rust/pal/functions.rs
+++ b/src/rust/pal/functions.rs
@@ -3,9 +3,41 @@
 
 use crate::pal::data_structures::SockAddrIn;
 
+#[cfg(feature = "catnip-libos")]
+const NUM_OCTETS_IN_IPV4: usize = 4;
+
+#[cfg(feature = "catnip-libos")]
+const NUM_SIN_ZERO_BYTES: usize = 8;
+
+#[cfg(all(feature = "catnip-libos", target_os = "windows"))]
+use windows::Win32::Foundation::CHAR;
+
+#[cfg(all(feature = "catnip-libos", target_os = "windows"))]
+use windows::Win32::Networking::WinSock::IN_ADDR;
+
+#[cfg(all(feature = "catnip-libos", target_os = "windows"))]
+use windows::Win32::Networking::WinSock::IN_ADDR_0;
+
+#[cfg(all(feature = "catnip-libos", target_os = "linux"))]
+use libc::in_addr;
+
 //======================================================================================================================
 // Windows functions
 //======================================================================================================================
+
+#[cfg(all(feature = "catnip-libos", target_os = "windows"))]
+pub fn create_sin_addr(octets: &[u8; NUM_OCTETS_IN_IPV4]) -> IN_ADDR {
+    IN_ADDR {
+        S_un: (IN_ADDR_0 {
+            S_addr: u32::from_le_bytes(*octets),
+        }),
+    }
+}
+
+#[cfg(all(feature = "catnip-libos", target_os = "windows"))]
+pub fn create_sin_zero() -> [CHAR; NUM_SIN_ZERO_BYTES] {
+    [CHAR(0); 8]
+}
 
 #[cfg(target_os = "windows")]
 pub fn get_addr_from_sock_addr_in(sock_addr_in: &SockAddrIn) -> u32 {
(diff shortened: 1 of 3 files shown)


src/rust/pal/functions.rs
```diff
@@ -3,10 +3,36 @@
...
+        S_un: (IN_ADDR_0 {
+            S_addr: u32::from_le_bytes(*octets),
+        }),
+    }
+}
+
+#[cfg(all(feature = "catnip-libos", target_os = "windows"))]
+pub fn create_sin_zero() -> [CHAR; 8] {
+    [CHAR(0); 8]
+}
```
ppenna, line 34: This is not specific to Windows, thus it should go in the PAL module, but not inside the "platform-dependent" submodule.

src/rust/pal/functions.rs
```diff
@@ -16,6 +42,18 @@ pub fn get_addr_from_sock_addr_in(sock_addr_in: &SockAddrIn) -> u32 {
...
+pub fn create_sin_addr(octets: &[u8; 4]) -> in_addr {
+    in_addr {
+        s_addr: u32::from_le_bytes(*octets),
+    }
+}
+
+#[cfg(all(feature = "catnip-libos", target_os = "linux"))]
+pub fn create_sin_zero() -> [u8; 8] {
+    [0; 8]
+}
```
ppenna, line 55: This is not specific to Linux, thus it should go in the PAL module, but not inside the "platform-dependent" submodule.

src/rust/pal/functions.rs
```diff
@@ -3,10 +3,36 @@
...
+pub fn create_sin_addr(octets: &[u8; 4]) -> IN_ADDR {
+    IN_ADDR {
+        S_un: (IN_ADDR_0 {
+            S_addr: u32::from_le_bytes(*octets),
+        }),
+    }
+}
+
+#[cfg(all(feature = "catnip-libos", target_os = "windows"))]
+pub fn create_sin_zero() -> [CHAR; 8] {
```
ppenna, line 32: Could we have a constant for `8`?

src/rust/pal/functions.rs
```diff
@@ -3,10 +3,36 @@
...
+
+#[cfg(all(feature = "catnip-libos", target_os = "linux"))]
+use libc::in_addr;
+
 //======================================================================================================================
 // Windows functions
 //======================================================================================================================
 
+#[cfg(all(feature = "catnip-libos", target_os = "windows"))]
+pub fn create_sin_addr(octets: &[u8; 4]) -> IN_ADDR {
```
ppenna, line 23: Could we have a constant for `4`?

src/rust/pal/functions.rs
```diff
@@ -16,6 +42,18 @@ pub fn get_addr_from_sock_addr_in(sock_addr_in: &SockAddrIn) -> u32 {
 // Linux functions
 //======================================================================================================================
 
+#[cfg(all(feature = "catnip-libos", target_os = "linux"))]
+pub fn create_sin_addr(octets: &[u8; 4]) -> in_addr {
```
ppenna, line 46: Could we have a constant for `4`?

src/rust/pal/functions.rs
```diff
@@ -16,6 +42,18 @@ pub fn get_addr_from_sock_addr_in(sock_addr_in: &SockAddrIn) -> u32 {
...
+    in_addr {
+        s_addr: u32::from_le_bytes(*octets),
+    }
+}
+
+#[cfg(all(feature = "catnip-libos", target_os = "linux"))]
+pub fn create_sin_zero() -> [u8; 8] {
```
ppenna, line 53: Could we have a constant for `8`?

ppenna (changes requested):
Could you reword the commit message and remove the extra ':' from it?

src/rust/pal/functions.rs
anandbonde: Hi @ppenna, this is indeed Linux specific code. For Windows, this code is just CHAR(0). Or maybe I'm not getting your comment.

src/rust/pal/functions.rs
anandbonde: Hi @ppenna, this is indeed Windows specific code (notice the CHAR(0)). For Linux, this code is just 0. Or maybe I'm not getting your comment.

ppenna: @anandbonde, I do understand that this code is only used in Windows. But fundamentally it is not specific to this platform, thus you could place it in a PAL layer that is common to both. Rationale:
- In the PAL windows module we find only platform-specific code.
- If Linux starts requiring this to be defined in the future we don't end up with duplicates
- If we introduce support for other LibOS that does require the same definition, we already have it.

src/rust/pal/functions.rs
ppenna: @anandbonde, I do understand that this code is only used in Windows. But fundamentally it is not specific to this platform, thus you could place it in a PAL layer that is common to both. Rationale:
- In the PAL windows module we find only platform-specific code.
- If Linux starts requiring this to be defined in the future we don't end up with duplicates
- If we introduce support for other LibOS that does require the same definition, we already have it.

src/rust/pal/functions.rs
ppenna: @anandbonde I got your point. I was reading `CHAR(0)` as u8.

src/rust/pal/functions.rs
ppenna: @anandbonde I got your point. I was reading `CHAR(0)` as u8.

src/rust/pal/functions.rs
anandbonde: Done

src/rust/pal/functions.rs
anandbonde: Done

ppenna (approved):

Issue #356, opened 2022-12-06
[doc] Testing Instructions for Demikernel


ppenna:
## Context

We should provide documentation that details how to test Demikernel.

## Proposed Solution

Have a detailed, standalone document, that provides detailed instructions on how to test Demikernel.

This documentation should be linked in the `README.md` file.

## Alternative Solution

We could inline testing instructions in the `README.md` file, but that would make the file long and tedious to read.

PR #358, opened 2022-12-06
[doc] Testing Instructions

ppenna:
Description
-------------

This PR addresses [#356 ](https://github.com/demikernel/demikernel/issues/356)

Summary of Changes
------------------------

- Added a new document that provides detailed instructions on how to test Demikernel
- Changed Demikernel's `README.md` file to link new testing document.
- Added missing entry to `demi_timedwait()` in the manual pages.

### DIFF
--- a/doc/testing.md
+++ b/doc/testing.md
@@ -1,0 +1,186 @@
+# Testing Demikernel
+
+This document contains instructions on how to test Demikernel. What it is covered:
+
+- [What are unit tests](#what-are-unit-tests)
+- [How to run unit tests](#how-to-run-unit-tests)
+- [What are system-level tests](#what-are-system-level-tests)
+- [How to run system-level tests](#how-to-run-system-level-tests)
+
+> The instructions in this file assume that you have at your disposal at least one machine with Demikernel's development
+environment set up. For more information on how to set up Demikernel's development environment, check out instructions
+in the `README.md` file.
+
+## What Are Unit Tests
+
+Unit tests exercise internal components of Demikernel, individually.
+
+Demikernel has several unit tests, and all of them spread across the source tree in `demikernel/src/`. To check their
+implementation, search for `mod tests` and `cfg(test)` in the code base.
+
+## How to Run Unit Tests
+
+You can run unit tests using Demikernel's build system:
+
+```bash
+# Run all unit tests of Demikernel with Catnap LibOS.
+# If you want to run unit tests on a different LibOS, then set the LIBOS flag accordingly.
+make LIBOS=catnap test-unit
+```
+
+## What Are System-Level Tests
+
+System-level tests exercise the user-level API that is exposed by Demikernel.
+
+These tests are located in `demikernel/examples/` and detailed in the following tables.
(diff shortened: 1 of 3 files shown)


doc/testing.md
```diff
@@ -0,0 +1,186 @@
...
+> The instructions in this file assume that you have at your disposal at least one machine with Demikernel's development
+environment set up. For more information on how to set up Demikernel's development environment, check out instructions
+in the `README.md` file.
+
+## What Are Unit Tests
+
+Unit tests exercise internal components of Demikernel, individually.
+
+Demikernel has several unit tests, and all of them spread across the source tree in `demikernel/src/`. In case you want
+to check for their implementation, then search for `mod tests` and `cfg(test)` in the code base.
```
BrianZill, line 19: I'd change this to:

Demikernel has several unit tests spread across the source tree in `demikernel/src/`.  To check their implementation, search for `mod tests` and `cfg(test)` in the code base.

doc/testing.md
```diff
@@ -0,0 +1,186 @@
...
+    --server-addr $SERVER_IPV4_ADDR \
+    --client-addr $CLIENT_IPV4_ADDR
+```
+
+### Running System-Level Tests with Demikernel's Build System (ie. `make`)
+
+```bash
+# Set parameters for Demikernel's TCP/UDP stack.
+export SERVER_IPV4_ADDR=10.9.1.10:12345
+export CLIENT_IPV4_ADDR=10.9.1.11:12345
```
BrianZill, line 103: We shouldn't be using the "12345" port number in published documentation.  According to [the IANA port number list](https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml), this is assigned to the "Italk Chat System".  We should either pick a number from the Ephemeral port range (49152-65535), or if we do want to use something from the User port range (1024-49151), at least use an unassigned one.

If you want something easy to remember in the Ephemeral port range, "56789" or "55555" would work. In the User port range, "7890" is unassigned.

This same comment applies to the "shell" instructions below, and unfortunately also the demikernel-ci.py script as well.

doc/testing.md
```diff
@@ -0,0 +1,186 @@
...
+
```
BrianZill, line 104: For completeness, I'll note that the IPv4 private address ranges specified in RFC 1918 are [specifically allowed to be used in documentation](https://www.rfc-editor.org/rfc/rfc5737), so we're okay using these 10/8 prefix addresses here.

There is also space specifically reserved for documentation, including three /24 prefixes for IPv4, but I don't think we need to bother with those.

BrianZill (changes requested):
Okay once comments are addressed.

BrianZill (approved):

doc/testing.md
ppenna: Done.

doc/testing.md
ppenna: Thanks for noting this. I changed the instructions to use the `192.0.2.0/24` address range.

PR #365, opened 2022-12-12
[ci] Checkout to Dev Branch

ppenna:
Description
-------------

This PR fixes the checkout commands that we have in our CI.

Since `dev` is now the main branch, we should be checking out on setup and clean up steps to it.


ppenna:
Merging this PR because it does not change the mainline code.

Issue #369, opened 2022-12-13
[inetstack] TCP Should Indicate to PDPIX when a Connection is Closing


BrianZill:
## Description

TCP doesn't currently inform PDPIX (and/or the socket layer) when the remote side of a connection is done sending (i.e., when a FIN arrives).

## Expected Behavior

TCP should notify the PDPIX and/or socket layer that a FIN has arrived, so it can notify the user that the remote side of the connection is done sending (if applicable).

## Related Issues

See also Issue #367, which is about the PDPIX part of the user notification system.

Issue #367, opened 2022-12-13
[pdpix] Missing Mechanism to Signal User on Connection Closing


BrianZill:
## Description

Our PDPIX API is currently missing a mechanism to inform the user that a TCP connection has been closed/shutdown by the remote peer.  Instead, a `demi_wait()` on a queue token returned by an outstanding `demi_pop()` call will simply never complete.

In the POSIX API, when a TCP connection (more generically, a SOCK_STREAM) is closed (or shutdown) by the other end, a `recv()` or `read()` call on the connection's socket will return zero to indicate that the other end is done sending.

## Expected Behavior

To achieve similar semantics to POSIX in PDPIX, we could have a pop on a SOCK_STREAM return a zero-length buffer to indicate that the other end is done sending.  Alternatively, we could also have it return an error code (POSIX's use of zero-byte-read to indicate connection closing comes out of a desired compatibility with Unix's zero-byte-read to indicate end-of-file.  It wasn't for that desired similarity, `recv()` could have returned an error code for nothing more to receive).

Note that we also should do something reasonable in all of the following cases:
- The user is waiting on an outstanding pop operation when the remote side's FIN arrives.
  - In this case, I'd expect we'd wake the waiter to pass the "connection closing" indication.  
- The user posts a pop _after_ the remote side's FIN arrives (i.e. no receive is outstanding at the moment of FIN arrival).
  - In this case, I'd expect we'd either return an error from the `demi_pop()` call directly or have a `demi_wait()` on the associated queue token return immediately with the "connection closing" indication.
- The user posts _another_ pop after receiving an indication (whatever we've decided that should be) that the remote side is done sending.
  - Same as above.
- A duplicate FIN arrives after we've already processed the original.
  - In this case, I'd expect that either the TCP stack doesn't bother to notify the PDPIX/socket layer of the additional FIN, or that the PDPIX/socket layer doesn't do anything different as a result of such notification.

## Related Issues

See also Issue #369, which is about the TCP-specific part of the "connection closing" notification mechanism.

Note we may want to have a similar semantics for closing a shared memory pipe.

PR #370, opened 2022-12-13
[inetstack] Push Empty Buffer on FIN

ppenna:
Description
--------------

This is a workaround for https://github.com/demikernel/demikernel/issues/369.

Summary of Changes
-------------------------

- Added code for push a zero-length buffer when a FIN has been received and the concerned TCP connection is closing.


BrianZill (approved):
As we discussed offline, this is good as a quick workaround for Issue #369.  The real fix needs to wait for Issue #367 to also be fixed.

Issue #372, opened 2022-12-14
[build] Missing to Run Some Unit Tests in `test-unit-rust` Target


ppenna:
## Description

The build system for both Linux and Windows is missing to run some unit tests in the `test-unit-rust` rule.

We are passing the `--lib` flag to `cargo test`, which runs all tests in `src/rust`  but skip all in `test/rust`.

https://github.com/demikernel/demikernel/blob/4ae37cd4e3e7890921199cbfccf66729a077462c/linux.mk#L193-L195

https://github.com/demikernel/demikernel/blob/4ae37cd4e3e7890921199cbfccf66729a077462c/windows.mk#L236-L238

Note that this issue affects the coverage of our regressions.

## How to Reproduce

1. Run `make LIBOS=catnap test-unit-rust`.
2. Note down the number of tests that ran.
3. Remove the `--lib` flag in the concerned makefile.
4. Run `make LIBOS=catnap test-unit-rust`.
5. Note that more tests run.

## Expected Behavior

All unit tests should run when triggering the rule `unit-test-rust`.

## Related Issues

Link related issues here.

PR #373, opened 2022-12-14
[test] Missing to Run Some Unit Tests in `test-unit-rust` Target

ppenna:
Description
-------------

This PR fixes https://github.com/demikernel/demikernel/issues/372.

Summary of Changes
------------------------

- Re-enabled unit tests that we were skipping
- Fixed default system test

### DIFF
--- a/linux.mk
+++ b/linux.mk
@@ -173,7 +173,7 @@ export CONFIG_PATH ?= $(HOME)/config.yaml
 export MTU ?= 1500
 export MSS ?= 1500
 export PEER ?= server
-export TEST ?= udp_push_pop
+export TEST ?= udp-push-pop
 export TIMEOUT ?= 30
 
 # Runs system tests.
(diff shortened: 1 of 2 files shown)


BrianZill (commented):
Did you mark the sga tests as "ignore" because you're now running each one specifically from the makefile?  And that is to run those tests single-threaded?

linux.mk
```diff
@@ -173,7 +173,7 @@ export CONFIG_PATH ?= $(HOME)/config.yaml
 export MTU ?= 1500
 export MSS ?= 1500
 export PEER ?= server
-export TEST ?= udp_push_pop
+export TEST ?= udp-push-pop
```
BrianZill, line 176: Does this mean we only run this one example test in the default case?

ppenna: It means that if someone simply invokes `make test-system-rust`, it will attempt to run the `udp-push-pop` example application, in `server` mode with `catnap` LibOS.

Specifically at this point I am just fixing the name of the default program. It is named `udp-push-pop` with dashes, not with underscores.

However, note that this is not the main fix that I aimed at in this PR.

anandbonde (approved):

Issue #203, opened 2022-03-29
[tcp] Implement FIN Retransmission


BrianZill:
Current code handles initial transmission of FIN by representing FIN as a zero-length buffer on the unsent queue and having the send code (both immediate and background) know to set the FIN bit on the segment when sending this buffer.  The zero-length buffer is then   placed on the unacknowledged (a.k.a. retransmission) queue.

The retransmission code needs to recognize the FIN marker on the unacknowledged queue and retransmit it.  Both the retransmission code and the regular send code should also be able to coalesce the FIN bit onto the last data packet.

ppenna:
Hello @BrianZill, may we now close this bug?

BrianZill:
Yes, we can close this.  The main part of this Issue was fixed in PR #374.  The coalescing part mentioned in the second paragraph is also mentioned in Issue #198 and should be fixed before that one is closed.

Issue #202, opened 2022-03-29
[inetstack] Retransmit Timer Issues


BrianZill:
We should review the retransmit timer implementation, specifically for the following issues:

- The code seems overzealous about pushing the retransmit timer into the future.  One case in particular is in the retransmit function, where it resets the expiration deadline regardless of why retransmit was called (it can be called not only on timer expiration, but also because of Fast Retransmit -- where I believe the deadline should remain untouched).
- The timer appears to also be extended upon every arrival of a new ACK, is this really desired behavior?
- The timer doesn't record the sequence number it is tracking for retransmission, instead it does the above, which could make it wait longer than desired for retransmission.  Again, review that this is desired behavior.

Issue #198, opened 2022-03-29
[inetstack] Send Side Buffer Handling Issues


BrianZill:
There are currently two implementations of the basic sender algorithm, one for immediate sending, and one for background sending.  They should do roughly the same thing, although they currently don't, and there should be just a single implementation that is called from both places.

Also, in the immediate send case, nothing checks that the length of the buffer for immediate send actually fits within the MSS.  A common send routine that runs all sends through the unsent queue could fix this effortlessly.

Finally, while splitting of large send requests into multiple segments is supported in the background sending case, coalescing of multiple small buffers into larger segments in not.  This capability should be added.  If there is a FIN, it should also be combined with the last data buffer into a single segment (might affect the representation of a FIN on the unacked queue, see Issue demikernel/demikernel#203).

Issue #195, opened 2022-04-29
[tcp] Retransmit Routine Shouldn't Remove Data from Unacknowledged Queue


BrianZill:
Description
========

The existing retransmit routine has a fatal flaw, it is removing data from the unacknowledged queue.  This breaks our logic that tracks unacknowledged data.  Instead, retransmit() should be retransmitting the data without removing it from the queue.

Since this bug is very serious, I've disabled the calling of retransmit().  This should be reenabled when this bug is fixed.

ppenna:
Hello @BrianZill, may we now close this bug?

BrianZill:
Yes, this Issue was fixed in PR #374.

PR #374, opened 2022-12-15
[tcp] Fix Retransmit and Cleanup Retransmission Timeout Calculator

BrianZill:
This PR addresses several outstanding issues:

- Issue #195:
  - This PR fixes this bug.  We no longer remove data from the unacknowledged queue when retransmitting it.
- Issue #198:
  - However, this issue still stands - In particular, the part about coalescing buffers together to produce a maximum sized packet. We still only retransmit the oldest unacknowledged data buffer (or FIN).  We don't try to resend a full MSS.  That part of the Issue isn't actually a bug though, it's just a performance optimization.
- Issue #202:
  - This PR fixes the main part of this Issue - We no longer modify the RTO value when performing "Fast Retransmit" (or rather we won't, once the rest of fast retransmit is implemented).
- Issue #203:
  - This PR also fixes this bug.  We will now retransmit a FIN, if that is all that is left on the unacknowledged queue upon retransmission timeout (RTO) or fast retransmit invocation.  Again, with the caveat that we don't coalesce a subsequent FIN onto a previous retransmitted data segment (as mentioned in 198).

This PR also cleans up the `RtoCalculator` code to be easier to read and understand, renames some confusingly named functions, and removes the dependency upon the third party "float_duration" crate.  And finally, it restructures the retransmission code in a way that will allow us to implement fast retransmit as a direct call, without needing to use a watched value that is monitored by the "retransmitter" background routine.  There are still some remnants of the latter in "retransmitter", which should come out when we complete the fast retransmit implementation.


src/rust/inetstack/protocols/tcp/established/sender.rs
```diff
@@ -273,6 +269,37 @@ impl Sender {
...
+                    // This buffer is the end-of-send marker.  Retransmit the FIN.
+                    header.fin = true;
+                }
+                cb.emit(header, Some(data), first_hop_link_addr);
+            }
+        } else {
+            // We shouldn't enter the retransmit routine with an empty unacknowledged queue.  So maybe we should assert
+            // here?  But this is relatively benign if it happens, and could be the result of a race-condition or a
+            // mismanaged retransmission timer, so asserting would be over-reacting.
+            warn!("Retransmission with empty unacknowledged queue?");
```
ppenna, line 299: You could log an error. This will not panic but will be pedantic.

ppenna (approved):
This looks good, thanks!

Please remind to reword commits before merging.

Issue #256, opened 2021-11-19
[build] Bump DPDK Version


ppenna:
Description
-------------

We should bump DPDK version to 22.21.

PR #375, opened 2022-12-15
[build] Bumping DPDK Version

ppenna:
Description
-------------

This PR fixes https://github.com/demikernel/demikernel/issues/256

Summary of Changes
------------------------

- Updated `scripts/setup/dpdk.sh` to support DPDK 22.11
- Updated `dpdk-rs` and fixed concerned changes

### DIFF
--- a/src/rust/catnip/runtime/mod.rs
+++ b/src/rust/catnip/runtime/mod.rs
@@ -29,25 +29,24 @@ use crate::runtime::{
         rte_eth_link_get_nowait,
         rte_eth_macaddr_get,
         rte_eth_promiscuous_enable,
-        rte_eth_rx_mq_mode_ETH_MQ_RX_RSS as ETH_MQ_RX_RSS,
+        rte_eth_rss_ip,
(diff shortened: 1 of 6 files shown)


src/rust/catnip/runtime/mod.rs
```diff
@@ -254,31 +253,28 @@ impl DPDKRuntime {
 
         println!("dev_info: {:?}", dev_info);
         let mut port_conf: rte_eth_conf = unsafe { MaybeUninit::zeroed().assume_init() };
-        port_conf.rxmode.max_rx_pkt_len = if use_jumbo_frames {
+        port_conf.rxmode.max_lro_pkt_size = if use_jumbo_frames {
```
anandbonde, line 256: Why are we using a different field now? Or has it just been renamed?

anandbonde (approved):
Looks good.

src/rust/catnip/runtime/mod.rs
ppenna: They changed some variables and constants.

PR #378, opened 2022-12-16
[catnapw] Use `DemiBuffer` Instead of `Buffer`/`DataBuffer`

anandbonde:
Partially addresses issues https://github.com/demikernel/demikernel/pull/320 and https://github.com/demikernel/demikernel/issues/274 for catnapw (Windows)


ppenna (approved):

ppenna:
Merging this to unblock bumping branches.

Issue #388, opened 2022-12-28
[build] Support C Format Checking on Windows Build System


ppenna:
## Description

We are currently missing the C format checking on Windows build system.

https://github.com/demikernel/demikernel/blob/c2c867505a07dc4ac3831fc03d2163202997a0e2/windows.mk#L181-L186

## Expected Behavior

The Windows build system should support C formatting checking as the Linux build system does.

mvp-2003:
Hello @ppenna 
Can I please take up this issue?

Thank you.

Issue #386, opened 2022-12-28
[ci] Integrate Catnapw with Our Regression System


ppenna:
## Context

We have recently introduced Catnap for Windows (Catnapw), thus we should integrate it with our regression system.

## Proposed Solution

- [ ] Integrate Catnapw building pipeline with our regression system.
  - [ ] Format checking.
  - [x] Debug building. https://github.com/demikernel/demikernel/pull/389
  - [x] Release building. https://github.com/demikernel/demikernel/pull/389
- [ ] Integrate Catnapw testing pipeline with our regression system.
  - [ ] Unit testing.
  - [ ] System testing.

PR #389, opened 2022-12-28
[ci] Initial Integration of Catnapw with Regression System

ppenna:
Description
-------------

- This PR is a partial fix for https://github.com/demikernel/demikernel/issues/386.
- This PR is a partial fix for https://github.com/demikernel/demikernel/issues/388.

Summary of Changes
------------------------

- Integrated Catnapw building with our regression system.
  - Building is enabled on dev branch only, because the GitHub Actions plugin that we use does not support passing environment variables to a Windows remote.
- Disabled C format checking on Windows, because it is not implemented.
- Wrote a documentation with Build instructions for Windows.

### DIFF
--- a/.github/workflows/catnapw.yml
+++ b/.github/workflows/catnapw.yml
@@ -1,0 +1,225 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+name: Main Catnapw
+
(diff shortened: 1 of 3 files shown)


anandbonde (approved):
Looks good to me!

Issue #390, opened 2023-01-02
[test] `tcp-ping-pong` Does Not Wait for Full Receive Buffer


ppenna:
## Description

In the `tcp-ping-pong` system-level test, we assume that the scatter-gather array that the testing client receives from the testing server has precisely the length that it expects.

TCP is a stream oriented protocol, thus we cannot make this assumption.

## Evidence

In the code snippet below, in line 214, we have no guarantees that the received scatter-gather array has the expected length.

https://github.com/demikernel/demikernel/blob/536d220a0beae52a279f607c15bb980d28fbc5d5/examples/rust/tcp-ping-pong.rs#L209-L231

## Expected Behavior

The testing client should not make any assumption on the length of the scatter-gather array that it receives from the testing server.

PR #391, opened 2023-01-02
[test] `tcp-ping-pong` Does Not Wait for Full Receive Buffer

ppenna:
## Description

This PR fixes https://github.com/demikernel/demikernel/issues/390.

## Summary of Changes

- Fixed `tcp-ping-pong` to buffer received response on client-side
- Changed `tcp-ping-pong` to send buffers with different payload at each round.

### DIFF
--- a/examples/rust/tcp-ping-pong.rs
+++ b/examples/rust/tcp-ping-pong.rs
@@ -246,10 +246,10 @@ fn client(remote: SocketAddrV4) -> Result<()> {
 
 /// Prints program usage and exits.
 fn usage(program_name: &String) {
-    println!("Usage: {} MODE address\n", program_name);
-    println!("Modes:\n");
-    println!("  --client    Run program in client mode.\n");
-    println!("  --server    Run program in server mode.\n");
+    println!("Usage: {} MODE address", program_name);
+    println!("Modes:");
+    println!("  --client    Run program in client mode.");
+    println!("  --server    Run program in server mode.");
 }
 
 //======================================================================================================================
--- a/examples/rust/tcp-ping-pong.rs
+++ b/examples/rust/tcp-ping-pong.rs
@@ -20,6 +20,7 @@ use ::std::{
     panic,
     slice,
     str::FromStr,
+    u8,
 };
 
 #[cfg(target_os = "windows")]
@@ -42,7 +43,6 @@ use ::demikernel::perftools::profiler;
 //======================================================================================================================
 
 const BUFFER_SIZE: usize = 64;
-const FILL_CHAR: u8 = 0x65;
 
 //======================================================================================================================
 // mksga()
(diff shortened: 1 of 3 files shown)


examples/rust/tcp-ping-pong.rs
```diff
@@ -68,6 +68,94 @@ fn mksga(libos: &mut LibOS, size: usize, value: u8) -> demi_sgarray_t {
...
+            Ok(qt) => qt,
+            Err(e) => panic!("pop failed: {:?}", e.cause),
+        };
+        let sga: demi_sgarray_t = match libos.wait(qt, None) {
+            Ok(qr) if qr.qr_opcode == demi_opcode_t::DEMI_OPC_POP => unsafe { qr.qr_value.sga },
+            Err(e) => panic!("operation failed: {:?}", e.cause),
+            _ => panic!("unexpected result"),
+        };
+
+        // Sanity check received data.
```
BrianZill, line 143: Is this the right comment?  This block appears to be copying the data out to the `recvbuf`, not sanity checking it.

BrianZill (approved):
This change is okay (assuming this test isn't about performance).

But it brings into question what the main purpose of this test is.  If it is for correctness, then it's a reasonable test of the basics, but it doesn't check for data integrity inside each "round".  If it is for performance, then this change is going to hurt, as it introduces an extra data copy - if we care about performance here, we should hang on to the received scatter-gather arrays until we get all the data for one "round" and then sanity check the array data correctly (or perform partial checks on each receive as they arrive).  Ideally, on the server side we'd then reuse that same scatter-gather array to return that data to the client, rather than recreate it on the server side.

Even more ideally, our API would be flexible enough to allow us to modify multiple received scatter-gather arrays that compose a "round" (potentially including overlap with adjacent "rounds") to make the returning push operation take a single scatter-gather array that just describes the current "round" (without requiring any copying of the data).

examples/rust/tcp-ping-pong.rs
ppenna: Oh, this is indeed not the right comment. Thanks for noting, I have fixed it.

ppenna:
> This change is okay (assuming this test isn't about performance).
> 
> But it brings into question what the main purpose of this test is. If it is for correctness, then it's a reasonable test of the basics, but it doesn't check for data integrity inside each "round". If it is for performance, then this change is going to hurt, as it introduces an extra data copy - if we care about performance here, we should hang on to the received scatter-gather arrays until we get all the data for one "round" and then sanity check the array data correctly (or perform partial checks on each receive as they arrive). Ideally, on the server side we'd then reuse that same scatter-gather array to return that data to the client, rather than recreate it on the server side.
> 
> Even more ideally, our API would be flexible enough to allow us to modify multiple received scatter-gather arrays that compose a "round" (potentially including overlap with adjacent "rounds") to make the returning push operation take a single scatter-gather array that just describes the current "round" (without requiring any copying of the data).

@BrianZill this is about fixing the correctness of the test only. Concerning the sanity check of the data in each round, we are doing that: at each round, we sanity check if the received buffer matches the `fill_char`. Are you suggesting us to have something different than that?

Issue #397, opened 2023-01-09
[tools] Support NFS Directories in `demikernel-ci`


ppenna:
## Context

The `demikernel-ci` tool enables one to launch builds and tests on a remote pair of machines.

The tool was initially designed to address the use-case scenario in which each machine has its own local clone of the demikernel repository and thus `demikernel-ci` would use that location to build and launch tests.

Unfortunately, if the target location for the local clone of the `demikernel` project is a NFS directory, the `demikernel-ci` tool will fail because the two remote machines would compete with one another.

## Proposed Solution

Add an optional flag in `demikernel-ci` to enable building only on one machine of the pair.

PR #398, opened 2023-01-09
[tools] Support NFS Directories in `demikernel-ci`

ppenna:
## Description

This PR closes https://github.com/demikernel/demikernel/issues/397

## Summary of Changes

- Added an optional flag in `demikernel-ci` to run build steps only on one machine.


iyzhang (approved):
lgtm

Issue #400, opened 2023-01-09
[catloop] Sockets LibOS Over Shared Memory


ppenna:
## Context

Currently, if two processes running on the same host want to communicate using a sockets API, they would have to rely on a Network LibOS.

Note however that this is not efficient, because processes are running local and they should not require a full-weight network stack.

## Proposed Solution

We should expose a sockets API on top of Catmem LibOS. This would enable sockets semantics over shared memory.

PR #401, opened 2023-01-09
[catloop] Initial LibOS Implementation

ppenna:
## Description

This PR closes https://github.com/demikernel/demikernel/issues/400

## Summary of Changes

- Introduced Catloop LibOS


src/rust/runtime/memory/buffer/databuffer.rs
```diff
@@ -73,10 +73,6 @@ impl DataBuffer {
 
     /// Creates a data buffer from a raw pointer and a length.
     pub fn from_raw_parts(data: *mut u8, len: usize) -> Result<Self, Fail> {
-        // Check if arguments are valid.
```
iyzhang, line 76: Does catloop use zero-length buffers?

iyzhang (changes requested):
This all looks good to me. I would like to see some unit tests before we add this though.

src/rust/runtime/memory/buffer/databuffer.rs
ppenna: No, this was a bug. But note that it does trick the sga length for encoding the magic close.

src/rust/catloop/futures/accept.rs
```diff
@@ -0,0 +1,214 @@
...
+    let len: usize = sga.sga_segs[0].sgaseg_len as usize;
+    let slice: &mut [u8] = unsafe { slice::from_raw_parts_mut(ptr, len) };
+    slice.copy_from_slice(&port.to_ne_bytes());
+    let qt_tx: QToken = control_duplex_pipe.push(&sga)?;
+    catmem.borrow_mut().free_sgarray(sga)?;
+    Ok(qt_tx)
+}
+
+/// Waits for a connection request to arrive.
+fn wait_connect_request(
```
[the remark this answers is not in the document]
ppenna: Sure. It is now fixed.

src/rust/catloop/futures/accept.rs
```diff
@@ -0,0 +1,214 @@
...
+    if !passed {
+        let e: Fail = Fail::new(libc::EPROTO, "invalid request");
+        error!("failed to establish connection ({:?})", e);
+        Err(e)
+    } else {
+        Ok(())
+    }
+}
+
+// Sends the port number to the remote peer.
```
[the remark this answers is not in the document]
ppenna: Sure. It is now fixed.

src/rust/catloop/duplex_pipe.rs
```diff
@@ -0,0 +1,96 @@
...
+        self.tx
+    }
+
+    /// Returns the queue descriptor that is associated with the underlying receiving simplex pipe.
+    pub fn rx(&self) -> QDesc {
+        self.rx
+    }
+
+    /// Creates a duplex pipe.
+    pub fn create_pipe(catmem: Rc<RefCell<CatmemLibOS>>, ipv4: &Ipv4Addr, port: u16) -> Result<Self, Fail> {
```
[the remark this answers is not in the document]
ppenna: Sure. It is now fixed.

ppenna:
> This all looks good to me. I would like to see some unit tests before we add this though.

All tests from our infra-structure currently run on catloop as well. Github Actions is not yet integrated because changes must be in dev branch first.

iyzhang (approved):

src/rust/catloop/futures/accept.rs
```diff
@@ -0,0 +1,214 @@
...
+        // Advance to next state in the connection establishment protocol.
+        self_.state = ServerState::AcceptedConnectRequest(qt_tx, duplex_pipe.clone());
+    }
+
+    // Re-schedule co-routine for later execution.
+    ctx.waker().wake_by_ref();
+    return Poll::Pending;
+}
+
+fn establish_connection(
```
anandbonde, line 200: Can we name this just `connect`? Or would that hide some implied meaning?

src/rust/catloop/futures/accept.rs
```diff
@@ -0,0 +1,214 @@
...
+    let len: usize = sga.sga_segs[0].sgaseg_len as usize;
+    let slice: &mut [u8] = unsafe { slice::from_raw_parts_mut(ptr, len) };
+    slice.copy_from_slice(&port.to_ne_bytes());
+    let qt_tx: QToken = control_duplex_pipe.push(&sga)?;
+    catmem.borrow_mut().free_sgarray(sga)?;
+    Ok(qt_tx)
+}
+
+/// Waits for a connection request to arrive.
+fn listen_and_accept_connections(
```
anandbonde, line 168: Would naming this just `listen_and_accept` be a good idea?

anandbonde (approved):
Overall, looks good to me. It would be good to have some Unit Tests around these components as they are substantially big enough to be tested separately.

src/rust/runtime/memory/buffer/databuffer.rs
BrianZill: I don't understand.  What purpose does moving the `len` check until after the `data` check serve?

But more importantly, we're trying to remove the DataBuffer type from the system, I only left it in place since Anand was in the middle of creating the Windows-compatible version of catnap.  So, the only changes we should be making to this file is to delete it.

src/rust/catloop/futures/accept.rs
```diff
@@ -0,0 +1,214 @@
...
+) -> Poll<Result<(QDesc, SocketAddrV4, Rc<DuplexPipe>), Fail>> {
+    // Check if a connection request arrived.
+    if let Some(handle) = DuplexPipe::poll(&self_.catmem, qt_rx)? {
+        // Check if this is a valid connection request.
+        check_connect_request(&self_.catmem, handle, qt_rx)?;
+
+        // Create underlying pipes before sending the port number through the
+        // control duplex pipe. This prevents us from running into a race
+        // condition were the remote makes progress faster than us and attempts
+        // to open the duplex pipe before that it is created.
```
BrianZill, line 181: I think there is an extra word ("that") in this comment line.

src/rust/catloop/futures/connect.rs
```diff
@@ -0,0 +1,227 @@
...
+        let self_: &mut ConnectFuture = self.get_mut();
+
+        // Poll Catmem LibOS to make progress on ongoing operations.
+        self_.catmem.borrow_mut().poll();
+
+        // Act according to the state in the connection establishment protocol.
+        //
+        //  CLIENT                                                       SERVER
+        //    InitiateConnectRequest   --- ( msg: connect-request ) --->   WaitingConnectRequest
+        //    InitiateConnectRequest                                       WaitingConnectRequest
```
BrianZill, line 108: Is this line correct?  Shouldn't the Client be in "ConnectRequestSent" immediately after sending the connect request?
And the Server go to "AcceptedConnectRequest"?

src/rust/catloop/futures/connect.rs
```diff
@@ -0,0 +1,227 @@
...
+        //     |                                                           WaitingConnectRequest
+        //     V                                                            |
+        //    ConnectRequestSent                                            V
+        //    ConnectRequestSent       <----- ( ack: port-number ) -----  AcceptedConnectRequest
+        //    ConnectRequestSent                                            |
```
BrianZill, line 113: Likewise, shouldn't the Client now be in "ConnectAckReceived"?

src/rust/catloop/futures/accept.rs
```diff
@@ -0,0 +1,214 @@
...
+/// Waits for a connection request to arrive.
+fn listen_and_accept_connections(
+    self_: &mut AcceptFuture,
+    ctx: &mut Context<'_>,
+    qt_rx: QToken,
+) -> Poll<Result<(QDesc, SocketAddrV4, Rc<DuplexPipe>), Fail>> {
+    // Check if a connection request arrived.
+    if let Some(handle) = DuplexPipe::poll(&self_.catmem, qt_rx)? {
+        // Check if this is a valid connection request.
+        check_connect_request(&self_.catmem, handle, qt_rx)?;
```
BrianZill, line 176: What is this check protecting against?

src/rust/catloop/mod.rs
```diff
@@ -0,0 +1,521 @@
...
+    // Cooks a magic connect message.
+    pub fn cook_magic_connect(catmem: &Rc<RefCell<CatmemLibOS>>) -> Result<demi_sgarray_t, Fail> {
+        let sga: demi_sgarray_t = catmem
+            .borrow_mut()
+            .alloc_sgarray(mem::size_of_val(&CatloopLibOS::MAGIC_CONNECT))?;
+
+        let ptr: *mut u8 = sga.sga_segs[0].sgaseg_buf as *mut u8;
+        let len: usize = sga.sga_segs[0].sgaseg_len as usize;
+        let slice: &mut [u8] = unsafe { slice::from_raw_parts_mut(ptr, len) };
+        slice.copy_from_slice(&CatloopLibOS::MAGIC_CONNECT.to_ne_bytes());
```
BrianZill, line 429: Since MAGIC_CONNECT is just a single byte, why all this code in this function (and `is_magic_connect`) to create a slice?

src/rust/catloop/mod.rs
```diff
@@ -0,0 +1,521 @@
...
+    pub fn sgafree(&self, mut sga: demi_sgarray_t) -> Result<(), Fail> {
+        // Sanity check length of scatter-gather array.
+        // ToDo: Drop this check once we support scatter-gather arrays with multiple segments.
+        if sga.sga_numsegs != 1 {
+            return Err(Fail::new(libc::EINVAL, "demi_sgarray_t has invalid segment count"));
+        }
+
+        // Check for magic close message.
+        if sga.sga_segs[0].sgaseg_len == 0 {
+            sga.sga_segs[0].sgaseg_len = mem::size_of_val(&Self::MAGIC_CLOSE) as u32;
```
BrianZill, line 341: This looks dangerous.  Can there be a legitimate zero-length scatter-gather array here?  If so, this could lead to a segmentation fault in `is_magic_close` below.

src/rust/catloop/mod.rs
```diff
@@ -0,0 +1,521 @@
...
+            },
+            Some((demi_opcode_t::DEMI_OPC_PUSH, qd)) | Some((demi_opcode_t::DEMI_OPC_POP, qd)) => {
+                let mut qr: demi_qresult_t = self.catmem.borrow_mut().pack_result(handle, qt)?;
+                qr.qr_qd = qd.into();
+
+                if qr.qr_opcode == demi_opcode_t::DEMI_OPC_POP {
+                    let mut sga: &mut demi_sgarray_t = unsafe { &mut qr.qr_value.sga };
+
+                    // Check for close.
+                    if self.is_magic_close(sga) {
```
BrianZill, line 389: What if someone legitimately sends the MAGIC_CLOSE bytes?  Is that byte sequence not allowed?

ppenna:
@anandbonde All tests from our infra-structure currently run on catloop as well. Github Actions is not yet integrated because changes must be in dev branch first. Like in other libOses, testing individual componentes in the PDPIX layer is not possible, with the current architecture of Demikernel.

src/rust/catloop/futures/accept.rs
ppenna: Sure, I renamed it.

src/rust/catloop/futures/accept.rs
ppenna: Sure, I renamed it.

src/rust/catloop/mod.rs
```diff
@@ -0,0 +1,521 @@
...
+
+//======================================================================================================================
+// Associated Functions
+//======================================================================================================================
+
+impl CatloopLibOS {
+    /// Magic payload used to identify close requests.
+    const MAGIC_CLOSE: u32 = 0xdeadbeef;
+    /// Magic payload used to identify connect requests.
+    /// It must be a single byte to ensure atomicity.
```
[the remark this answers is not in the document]
ppenna: Sure, I added.

src/rust/catloop/futures/accept.rs
ppenna: Sure, fixed.

src/rust/catloop/futures/connect.rs
ppenna: Yes, and that is what I tried to draw using the lines. I intentionally replicated the states, to give some idea of time moving forward using the vertical axis. Note that state transitions are always indicated with vertical and horizontal arrows.

src/rust/catloop/futures/connect.rs
ppenna: No. The client just progresses to `ConneckAckReceived` when it is in the `ConnectRequestSent` and it receives an ack with the port number piggy backed in that reply.

src/rust/catloop/futures/accept.rs
ppenna: It performs two checks:
- The completed I/O queue operation associated to the queue token `qt` concerns a `pop()` operation that has completed.
- The payload received from that `pop()` operation is a valid and legit `MAGIC_CONNECT` message.

I added further comments to make that clear in the code.

src/rust/catloop/mod.rs
ppenna: No, we cannot have a legit zero-length scatter-gather array thus segmentation fault is not possible.

src/rust/catloop/mod.rs
ppenna: It is allowed and will not yield to an erroneous close. The rationale is that a close sequence consists of:

- A scatter-gather array with zero-length. Note that a user cannot allocate this by construction, from the `sga_alloc()` function.
- A payload with the `MAGIC_CLOSE` sequence.

src/rust/catloop/mod.rs
ppenna: Oh, thanks! I applied the optimization that you suggested.

src/rust/catloop/futures/connect.rs
BrianZill: Since line 108 comes after line 107 in time, any transitions resulting from the actions of line 107 should be reflected in the states of the client and server by line 108.

src/rust/catloop/futures/connect.rs
BrianZill: I think I'm reading your diagrams the way similar diagrams in IETF RFCs are drawn.  I see the "ack with the port number piggy backed" arriving on line 112, and thus by line 113 the client should be in "ConnectAckReceived" as it processed the incoming ack at the time of line 112.

For an example, see Figures 7 through 12 in [RFC 793](https://www.rfc-editor.org/rfc/rfc793#3.4).

src/rust/catloop/mod.rs
BrianZill: How does that first point work?  In this code here, `is_magic_close` checks for the magic payload (with length correct for normal data) and adjusts the length to the bogus "magic close" value if the payload is the magic one.  It looks like this code gets called for things other than close, so if someone happens to send the magic payload, won't this turn it into a close?

src/rust/runtime/memory/buffer/databuffer.rs
BrianZill: I unresolved this, as I would still like to know why this file is included in this PR.

src/rust/catloop/mod.rs
ppenna: # What is Implemented

- Legit scatter-Gather arrays are allocated from Catmem LibOS, and have 1 segment:

https://github.com/demikernel/demikernel/blob/5596a9173071634833a6c0f4720f3d70f24fd0d0/src/rust/catmem/mod.rs#L163-L189

- A `close()` operation in Catloop LibOS translates to a `push()` request in Catmem LibOS (note that this permalink may change because it refers to the branch in the PR):

https://github.com/demikernel/demikernel/blob/c3970cc85e5c40f0af39e531051f7fbf2b57e67a/src/rust/catloop/mod.rs#L263-L297

- The magic payload is cooked there:

https://github.com/demikernel/demikernel/blob/c3970cc85e5c40f0af39e531051f7fbf2b57e67a/src/rust/catloop/mod.rs#L454-L467

- When the the message containing the magic payload arrives at the other end, we truncate the length of the buffer to zero when creating the `demi_qresult_t`:

https://github.com/demikernel/demikernel/blob/c3970cc85e5c40f0af39e531051f7fbf2b57e67a/src/rust/catloop/mod.rs#L369-L409

- At the receiver side, double free is prevented:

https://github.com/demikernel/demikernel/blob/c3970cc85e5c40f0af39e531051f7fbf2b57e67a/src/rust/catloop/mod.rs#L336-L352

BrianZill: Yes, that's exactly what I thought was happening.  My question is: what is different about the message that gets sent by `close()` (which calls `cook_magic_close()` to generate a scatter-gather array with the `MAGIC_CLOSE` message, and then pushes that message) from a normal user-generated message that just happens to contain the same sequence of bytes as the `MAGIC_CLOSE` message does?

Or to put it another way, how does the code at "the other end" detect that the `close()` generated message isn't a user-generated message that just happens to have those magic bytes?  The code in pack_result I flagged originally appears to just be looking for that byte sequence and setting the length to zero if it sees it.  That's my question.

ppenna: @BrianZill 

Oh, yes. I understand. Thanks for spotting this. I was focused on the application not crashing due to a segmentation faut on a zero-length buffer.

Currently, in the case where the other end gets a `MAGIC_CLOSE` sequence, it has no way to distinguish a `close()` from an usual `push()` . The alternatives that came up for solving this are:

- The end process that received this `MAGIC_CLOSE` sequence, waits for some timeout to decide whether or not that was a legit close. This does not fix the problem, because we are making assumptions on the "pace of the communication".
- We encode the `MAGIC_CLOSE` sequence as multiple messages, each of each with a different payload. This reduces the chance of an erroneous `close()` but does not fix the problem.
- We implement a four-way handshake like we have in TCP, to reduce even further the chance of having an erroneous `close()`. This also does not fix the problem, because it is a legit communication to have a sequence of `MAGIC_CLOSE`s being passed around. In TCP it works because each payload has an additional header message, where the `FIN` bit is set.

Let me know if you have any additional ideas that would permanently fix the problem and not incur in additional overheads.

src/rust/runtime/memory/buffer/databuffer.rs
ppenna: I reverted this file to its original state. Sorry for touching on it.

src/rust/catloop/mod.rs
BrianZill: Well... (you're going to love this one)...

We could re-architect the system to pass buffers instead of bytes through the `SharedRingBuffer`s.  I.e., Make them `SharedRingBuffer<usize>` rather than u8, change catmem to pass a reference to the `DemiBuffer` instead of looping through the individual bytes and pushing each through the queue individually, and change catloop to allocate all of its `DemiBuffer`s in the shared memory space so the other end can dereference the pointers and get something meaningful.  If the shared memory space could also be used to map the DPDK-backed `DemiBuffer`s, then data received off the network could be passed through catloop to another process with zero-copy.  That would be super-fast.

Oh, and this would make an out-of-band signal for "close" trivial, as we could just use all zeroes instead of a valid `DemiBuffer` address to indicate the channel was closing.

I suppose a less invasive change would be to make the `SharedRingBuffer`s pass a u16, but normally just use half of that to pass a u8 as you do now.  But that would give you lots of extra bits to pass out-of-band information, such as the close indication.

Or, since you are currently passing each u8 individually, you could add your own special multi-byte encoding.  Pick one value to be your special "close" byte (or starting byte of a multi-byte "close" sequence for atomicity reasons) and double it (at the sender, remove extra byte at the receiver) in the normal case but actually send the single byte to "close".  That would be more complicated than passing u16s though but would use up less shared memory space.

src/rust/catloop/futures/connect.rs
ppenna: I have changed the drawing based on what I understood that you want me to do. If the revised version does not meet your expectations, could you pate here what you have in mind?

src/rust/catloop/futures/connect.rs
ppenna: I have changed the drawing based on what I understood that you want me to do. If the revised version does not meet your expectations, could you paste here the drawing that you have in mind?

src/rust/catloop/mod.rs
ppenna: > Well... (you're going to love this one)...

Oh, definitely :)

I have changed Catmem LibOS to use a `SharedRingBuffer<u16>` and use the low byte for data and the high one for control information. Normal pushes, always fill in the high byte with zeros, whereas a "close" push fills in the high byte. With this change all the `MAGIC_CLOSE` logic was removed from Catloop LibOS and now there is no ambiguity on "close()",

src/rust/catmem/futures/push.rs
```diff
@@ -73,8 +73,9 @@ impl Future for PushFuture {
     fn poll(self: Pin<&mut Self>, ctx: &mut Context<'_>) -> Poll<Self::Output> {
         let self_: &mut PushFuture = self.get_mut();
         let mut index: usize = self_.index;
-        for x in &self_.buf[index..] {
-            match self_.ring.try_enqueue(*x) {
+        for low in &self_.buf[index..] {
+            let x: u16 = ((0 & 0xff) << 8) | (low & 0xff) as u16;
```
BrianZill, line 77: Why bother with the `((0 & 0xff) << 8) |` part?  Or really anything other than low as u16?  Not that this will hurt anything though.

src/rust/catmem/mod.rs
```diff
@@ -92,18 +92,36 @@ impl CatmemLibOS {
...
 
         let qd: QDesc = self.qtable.alloc(QType::MemoryQueue.into());
         assert_eq!(self.rings.insert(qd, Rc::new(ring)).is_none(), true);
 
         Ok(qd)
     }
 
+    // Pushes EoF.
+    fn push_eof(&mut self, ring: Rc<SharedRingBuffer<u16>>) -> Result<(), Fail> {
+        let x: u16 = ((1 & 0xff) << 8) | (0 & 0xff) as u16;
```
BrianZill, line 105: Likewise, I'd just do (1 << 8) as u16 here.  Not that it matters, the compiler will probably/hopefully process this whole expression into the equivalent constant at compile time.

src/rust/catloop/futures/connect.rs
BrianZill: I'm still not seeing any changes to the drawings, so I don't know what's going on with GitHub.

src/rust/catloop/futures/accept.rs
BrianZill: Okay.

src/rust/catloop/mod.rs
```diff
@@ -0,0 +1,462 @@
...
+//======================================================================================================================
+// Associated Functions
+//======================================================================================================================
+
+impl CatloopLibOS {
+    /// Magic payload used to identify connect requests.  It must be a single
+    /// byte to ensure atomicity while keeping the connection establishment
+    /// protocol. The rationale for this lies on the fact that a pipe in Catmem
+    /// LibOS operates atomically on bytes. If we used a longer byte sequence,
+    /// we would need to introduce additional logic to make share that
```
BrianZill, line 94: "share" -> "sure"

BrianZill (approved):
Looks okay now.

Thanks for fixing that in-band closing signal problem.

src/rust/catloop/mod.rs
BrianZill: Nice!

src/rust/catloop/futures/connect.rs
BrianZill: I'm still not seeing the changes to the drawings, so I don't know what's going on with GitHub.

Issue #405, opened 2023-01-10
[pal] Bad Level When Setting `SO_REUSEPORT`


ppenna:
## Description

We are currently using the wrong API level for setting the `SO_REUSEPORT`, consequently `set_so_reuseport()` always fails.

## Code Snippet

https://github.com/demikernel/demikernel/blob/f06b7d18533c8cb17d6f469a6854598dd6995b40/src/rust/pal/linux/mod.rs#L38-L49

## Expected Behavior

We should use the `SOL_SOCKET` level to correctly set `SO_REUSEPORT`.

PR #406, opened 2023-01-10
[pal] Bad Level When Setting `SO_REUSEPORT`

ppenna:
## Description

This PR fixes https://github.com/demikernel/demikernel/issues/405

## Summary of Changes

- Fixed the API level for `setsockopt()` when setting `SO_REUSEPORT`.


iyzhang (approved):
looks good to me. Thanks for catching this!

Issue #410, opened 2023-01-11
[tools] Bad Check on Number of Executed Jobs in `demikernel-ci`


ppenna:
## Description

The `demikernel-ci` is performing a bad check on the number of executed jobs. The `len(jobs)` test is wrong because after calling `wait_jobs()` we consume that list. Instead, we should check for `len(status)`, which gets populated after calling `wait_jobs()`

## Code Snippet

https://github.com/demikernel/demikernel/blob/89a4eeaf0189b731974a3881a2aa03b7a833de32/tools/demikernel-ci.py#L48-L61

PR #411, opened 2023-01-11
[tools] Bad Check on Number of Executed Jobs in `demikernel-ci`

ppenna:
## Description

This PR closes https://github.com/demikernel/demikernel/issues/410

## Summary of Changes

- Added missing initialization for `passed` variable in `wait_and_report()`
- Fixed test for number of executed jobs in `wait_and_report()`


iyzhang (approved):
Yup, this fixed the issue for me. Thanks for figuring this out!

Issue #412, opened 2023-01-11
[pal] Helper Function to Convert `std::net::SocketAddrV4` to `libc::sockaddr_in`


ppenna:
## Context

We perform a conversion from `std::net::SocketAddrV4` to `libc::sockaddr_in` in multiple LibOSes:

https://github.com/demikernel/demikernel/blob/89a4eeaf0189b731974a3881a2aa03b7a833de32/src/rust/catcollar/mod.rs#L429-L437

https://github.com/demikernel/demikernel/blob/89a4eeaf0189b731974a3881a2aa03b7a833de32/src/rust/catnap/mod.rs#L416-L424

https://github.com/demikernel/demikernel/blob/89a4eeaf0189b731974a3881a2aa03b7a833de32/src/rust/catnap/mod.rs#L416-L424

The code is replicated and it is critical due to the endianness of the structure.

## Proposed Solution

We should have a helper function that does the conversion between these two structures.

PR #413, opened 2023-01-11
[pal] Helper Function to Convert `std::net::SocketAddrV4` to `libc::sockaddr_in`

ppenna:
## Description

This PR closes https://github.com/demikernel/demikernel/issues/412

## Summary of Changes

- Added a helper function`sockaddrv4_to_sockaddr_in()` to perform the conversion between a `std::net::SocketAddrV4` to `libc::sockaddr_in`
- Changed the concerned code to rely on the helper function.
- Changed the conversion of `sin_port` and `sin_addr.s_addr` to be on network order.

### DIFF
--- a/src/rust/pal/linux/mod.rs
+++ b/src/rust/pal/linux/mod.rs
@@ -13,6 +13,7 @@ pub mod shm;
 
 use ::std::{
     mem,
+    net::SocketAddrV4,
     os::unix::prelude::RawFd,
 };
 
@@ -60,4 +61,21 @@ pub unsafe fn set_nonblock(fd: RawFd) -> i32 {
     // Set file flags.
     flags |= libc::O_NONBLOCK;
     libc::fcntl(fd, libc::F_SETFL, flags, 1)
+}
+
+/// Converts a [std::net::SocketAddrV4] into a [libc::sockaddr_in].
+pub fn socketaddrv4_to_sockaddr_in(addr: &SocketAddrV4) -> libc::sockaddr_in {
+    libc::sockaddr_in {
+        sin_family: libc::AF_INET as libc::sa_family_t,
+        sin_port: u16::to_be(addr.port()),
+        #[cfg(target_endian = "big")]
+        sin_addr: libc::in_addr {
+            s_addr: u32::to_be(u32::from_be_bytes(addr.ip().octets())) as libc::in_addr_t,
+        },
+        #[cfg(target_endian = "little")]
+        sin_addr: libc::in_addr {
+            s_addr: u32::from_le_bytes(addr.ip().octets()),
+        },
+        sin_zero: [0; 8],
+    }
 }
--- a/src/rust/catcollar/mod.rs
+++ b/src/rust/catcollar/mod.rs
@@ -29,6 +29,7 @@ use self::{
 use crate::{
     demikernel::config::Config,
     inetstack::operations::OperationResult,
+    pal::linux,
     runtime::{
         fail::Fail,
         memory::{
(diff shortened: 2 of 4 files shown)


src/rust/pal/linux/mod.rs
```diff
@@ -61,3 +62,15 @@ pub unsafe fn set_nonblock(fd: RawFd) -> i32 {
     flags |= libc::O_NONBLOCK;
     libc::fcntl(fd, libc::F_SETFL, flags, 1)
 }
+
+/// Converts a [std::net::SocketAddrV4] into a [libc::sockaddr_in].
+pub fn socketaddrv4_to_sockaddr_in(addr: &SocketAddrV4) -> libc::sockaddr_in {
+    libc::sockaddr_in {
+        sin_family: libc::AF_INET as u16,
```
BrianZill, line 69: I have no idea why libc defines `AF_INET` as a `libc::c_int` instead of as a `libc::sa_family_t`.  If they had, you wouldn't need this "`as`" clause.  But since you do need it, shouldn't it be "`as libc::sa_family_t`"?

src/rust/pal/linux/mod.rs
```diff
@@ -61,3 +62,15 @@ pub unsafe fn set_nonblock(fd: RawFd) -> i32 {
...
+        sin_port: u16::to_be(addr.port()),
+        sin_addr: libc::in_addr {
+            s_addr: u32::to_be(u32::from_be_bytes(addr.ip().octets())),
+        },
```
BrianZill, line 73: Again, libc uses its own type (`libc::in_addr_t`) as the type for `s_addr`.  So, I think line 72 should technically have "`as libc::in_addr_t`" added to the end of it.

But yes, this conversion is correct.  The version you're replacing was also correct (and avoided an unnecessary byte-swap), but only on little-endian machines.  The ideal thing here would probably be to define our own `pal::in_addr_t` type, with a from_be_bytes() method that creates the equivalent of a big-endian 32-bit unsigned value.

src/rust/pal/linux/mod.rs
```diff
@@ -61,3 +62,15 @@ pub unsafe fn set_nonblock(fd: RawFd) -> i32 {
     flags |= libc::O_NONBLOCK;
     libc::fcntl(fd, libc::F_SETFL, flags, 1)
 }
+
+/// Converts a [std::net::SocketAddrV4] into a [libc::sockaddr_in].
+pub fn socketaddrv4_to_sockaddr_in(addr: &SocketAddrV4) -> libc::sockaddr_in {
```
BrianZill, line 67: Ideally, this function wouldn't be here, it would be in src/rust/pal/sockets.rs (or somesuch), and it would return a value of a platform independent type, like pal::sockaddr_in (or somesuch).  And then there would be separate implementations of it for linux and Windows.  I.e. all platform-dependent code would be under pal, and everything in catcollor, catnap, catpowder, et cetera would be platform-independent.

BrianZill (approved):
This is okay, but we should really clean up the PAL at some point to actually be an abstraction layer and not a dumping ground for platform-specific code (see my comment about where `socketaddrv4_to_sockaddr_in()` should live).

src/rust/pal/linux/mod.rs
ppenna: Thanks for noting this. `u16` was also working, but I agree that moving to the correct type cast is better.

src/rust/pal/linux/mod.rs
ppenna: I agree that we should discuss a better approach for structuring the PAL module. Let us have it and then I handle the necessary reorg.

src/rust/pal/linux/mod.rs
ppenna: Leaving off the casting was working, but I introduced it, so that we are technically correct.

Also, I have introduced conditional compilation to select the right byte conversion technique.

Issue #414, opened 2023-01-11
[tools] Give log files meaningful names


iyzhang:
## Context

The current scripts use the job name to tag the stdout and stderr log files. It would be useful to name the files based on the hostname and branch and compile mode.

## Proposed Solution

Change name of log files in wait_jobs.

PR #415, opened 2023-01-12
[tools] Adding support for test directories and log files with hostnames

iyzhang:
This PR addresses #414. It adds support for creating a separate directory for each run of the demikernel-ci.py script. It keeps one older version from the last run and creates a new directory with the test configuration (libos, branch, compile mode). It then names each of the log files after the test name and whether the host was the server or client and the hostname where it ran. This should make it easier to debug and sort through our testing logs.

### DIFF
--- a/tools/demikernel-ci.py
+++ b/tools/demikernel-ci.py
@@ -6,6 +6,9 @@ import argparse
 import string
 import subprocess
 import time
+from os import mkdir
+from shutil import move,rmtree
+from os.path import isdir
 from typing import List
 
 # ======================================================================================================================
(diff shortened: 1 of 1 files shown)


tools/demikernel-ci.py
```diff
@@ -230,21 +238,32 @@ def run_pipeline(
     step: int = 0
     status: int = 0
 
+    # Create folder for test logs
+    log_directory = "{}-{}-{}".format(libos, branch, "debug" if is_debug else "release")
```
ppenna, line 242: I would change this statement to:

```python
log_directory = "{}-{}-{}".format(libos, branch, "debug" if is_debug else "release").replace("/", "_")
```

This enables us to run the pipeline on detached target branches. This is useful when the target branch was forced pushed.

ppenna (approved):
In additional, could you reword your commit so that the first line description is less than 50 characters long? This way they are properly displayed when navigating the commit history. Finally, you could link your issue in the remainder lines of the commit message.

tools/demikernel-ci.py
iyzhang: Cool, will do

Issue #416, opened 2023-01-13
[tools] Missing Support for Detached Branches


ppenna:
## Context

With the latest changes that we introduced for logging (see https://github.com/demikernel/demikernel/pull/415) we are no longer able to run the CI pipeline on detached branches, because `demikernel-ci` would attempt to create a directory with a slash '/' in its name to place logs:

https://github.com/demikernel/demikernel/blob/c1b68583d5ba84d252b640a19af0301b6103c1b8/tools/demikernel-ci.py#L241-L249

## Proposed Solution

Still, we should be able to run detached branches if we rename the output directory file for logs:

```python
log_directory = "{}-{}-{}".format(libos, branch, "debug" if is_debug else "release").replace("/", "_")
```

Issue #404, opened 2023-01-10
[tools] Missing Type Conversion in `demikernel-ci` for `--delay` Parameter


ppenna:
## Description

We are currently missing a type conversion from `string` to `float` in `demikernel-ci` for the `--delay` parameter.

Consequently, if one specifies this optional parameter, the `demikernel-ci` tool crashes.

## Screenshots

![image](https://user-images.githubusercontent.com/4939789/211570635-26796ad7-2988-4562-96a7-6c2194cf1990.png)

## Expected Behavior

The `demikernel-ci` tool should handle the conversion and don't crash.

PR #418, opened 2023-01-13
[tools] Minor Bug Fixes in `demikernel-ci`

ppenna:
## Description

- This PR fixes https://github.com/demikernel/demikernel/issues/416
- This PR fixes https://github.com/demikernel/demikernel/issues/404

## Summary of Changes

- Added missing support for detached branches.
- Added type conversion for delay parameter.
- Fixed type annotations for `str` types.
- Fixed cleanup step to use `sudo` when required.


iyzhang (approved):
Thanks for fixing these bugs and the ones that I accidentally missed from my last PR. Looks good to me.

Issue #402, opened 2023-01-09
[build] Drop Dependency on `nix`


ppenna:
## Context

We should drop dependency on `nix` crate and directly invoke `libc` functions.

PR #419, opened 2023-01-13
[build] Dropping Dependency on `nix`

ppenna:
## Description

This PR closes https://github.com/demikernel/demikernel/issues/402.

## Summary of Changes

- Dropped dependency on `nix` crate for Catnap LibOS.
- Dropped dependency on `nix` crate for Catcollar LibOS.

### DIFF
--- a/src/rust/catcollar/futures/accept.rs
+++ b/src/rust/catcollar/futures/accept.rs
@@ -12,12 +12,9 @@ use crate::{
         QDesc,
     },
 };
-use ::nix::{
-    errno::Errno,
-    sys::socket,
-};
 use ::std::{
     future::Future,
+    mem,
     os::unix::prelude::RawFd,
     pin::Pin,
     task::{
@@ -38,6 +35,8 @@ pub struct AcceptFuture {
     fd: RawFd,
     /// Queue descriptor of incoming connection.
     new_qd: QDesc,
+    /// Socket address of accept connection.
+    sockaddr: libc::sockaddr_in,
 }
 
 //==============================================================================
@@ -48,7 +47,12 @@ pub struct AcceptFuture {
 impl AcceptFuture {
     /// Creates a descriptor for an accept operation.
     pub fn new(qd: QDesc, fd: RawFd, new_qd: QDesc) -> Self {
-        Self { qd, fd, new_qd }
+        Self {
+            qd,
+            fd,
+            new_qd,
+            sockaddr: unsafe { mem::zeroed() },
+        }
     }
 
     /// Returns the queue descriptor associated to the target accept operation
@@ -74,36 +78,53 @@ impl Future for AcceptFuture {
 
     /// Polls the underlying accept operation.
     fn poll(self: Pin<&mut Self>, ctx: &mut Context<'_>) -> Poll<Self::Output> {
-        let self_: &AcceptFuture = self.get_mut();
-        match socket::accept(self_.fd as i32) {
+        let self_: &mut AcceptFuture = self.get_mut();
+        match unsafe {
+            let mut address_len: libc::socklen_t = mem::size_of::<libc::sockaddr_in>() as u32;
+            libc::accept(
+                self_.fd,
+                (&mut self_.sockaddr as *mut libc::sockaddr_in) as *mut libc::sockaddr,
+                &mut address_len,
+            )
+        } {
             // Operation completed.
-            Ok(new_fd) => {
+            new_fd if new_fd >= 0 => {
                 trace!("connection accepted ({:?})", new_fd);
 
                 // Set socket options.
                 unsafe {
                     if linux::set_tcp_nodelay(new_fd) != 0 {
-                        warn!("cannot set TCP_NONDELAY option");
+                        let errno: libc::c_int = *libc::__errno_location();
+                        warn!("cannot set TCP_NONDELAY option (errno={:?})", errno);
                     }
                     if linux::set_nonblock(new_fd) != 0 {
-                        warn!("cannot set NONBLOCK option");
+                        let errno: libc::c_int = *libc::__errno_location();
+                        warn!("cannot set O_NONBLOCK option (errno={:?})", errno);
                     }
                     if linux::set_so_reuseport(new_fd) != 0 {
-                        warn!("cannot set SO_REUSEPORT option");
+                        let errno: libc::c_int = *libc::__errno_location();
+                        warn!("cannot set SO_REUSEPORT option (errno={:?})", errno);
                     }
                 }
 
                 Poll::Ready(Ok(new_fd))
             },
-            // Operation in progress.
-            Err(e) if e == Errno::EWOULDBLOCK || e == Errno::EAGAIN => {
-                ctx.waker().wake_by_ref();
-                Poll::Pending
-            },
-            // Operation failed.
-            Err(e) => {
-                warn!("failed to accept connection ({:?})", e);
-                Poll::Ready(Err(Fail::new(e as i32, "operation failed")))
+
+            // Operation not completed, thus parse errno to find out what happened.
+            _ => {
+                let errno: libc::c_int = unsafe { *libc::__errno_location() };
+
+                // Operation in progress.
+                if errno == libc::EWOULDBLOCK || errno == libc::EAGAIN {
+                    ctx.waker().wake_by_ref();
+                    return Poll::Pending;
+                }
+                // Operation failed.
+                else {
+                    let message: String = format!("accept(): operation failed (errno={:?})", errno);
+                    error!("{}", message);
+                    return Poll::Ready(Err(Fail::new(errno, &message)));
+                }
             },
         }
     }
--- a/src/rust/catcollar/futures/connect.rs
+++ b/src/rust/catcollar/futures/connect.rs
@@ -5,19 +5,17 @@
 // Imports
 //==============================================================================
 
-use crate::runtime::{
-    fail::Fail,
-    QDesc,
-};
-use ::nix::{
-    errno::Errno,
-    sys::{
-        socket,
-        socket::SockaddrStorage,
+use crate::{
+    pal::linux,
+    runtime::{
+        fail::Fail,
+        QDesc,
     },
 };
 use ::std::{
     future::Future,
+    mem,
+    net::SocketAddrV4,
     os::unix::prelude::RawFd,
     pin::Pin,
     task::{
@@ -36,8 +34,8 @@ pub struct ConnectFuture {
     qd: QDesc,
     // Underlying file descriptor.
     fd: RawFd,
-    /// Destination address.
-    addr: SockaddrStorage,
+    /// Connect address.
+    sockaddr: libc::sockaddr_in,
 }
 
 //==============================================================================
@@ -47,8 +45,12 @@ pub struct ConnectFuture {
 /// Associate Functions for Connect Operation Descriptors
 impl ConnectFuture {
     /// Creates a descriptor for a connect operation.
-    pub fn new(qd: QDesc, fd: RawFd, addr: SockaddrStorage) -> Self {
-        Self { qd, fd, addr }
+    pub fn new(qd: QDesc, fd: RawFd, addr: SocketAddrV4) -> Self {
+        Self {
+            qd,
+            fd,
+            sockaddr: linux::socketaddrv4_to_sockaddr_in(&addr),
+        }
     }
 
     /// Returns the queue descriptor associated to the target connect operation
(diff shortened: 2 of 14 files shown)


src/rust/catcollar/futures/accept.rs
```diff
@@ -64,46 +68,63 @@ impl AcceptFuture {
...
 impl Future for AcceptFuture {
     type Output = Result<RawFd, Fail>;
 
     /// Polls the underlying accept operation.
     fn poll(self: Pin<&mut Self>, ctx: &mut Context<'_>) -> Poll<Self::Output> {
-        let self_: &AcceptFuture = self.get_mut();
-        match socket::accept(self_.fd as i32) {
+        let self_: &mut AcceptFuture = self.get_mut();
+        match unsafe {
+            let mut address_len: libc::socklen_t = 0;
```
BrianZill, line 83: Why is this zero?  Shouldn't this be sizeof sockaddr_in?  Assuming we actually want the address filled in, of course.

src/rust/catcollar/futures/accept.rs
```diff
@@ -64,46 +68,63 @@ impl AcceptFuture {
...
-            Err(e) if e == Errno::EWOULDBLOCK || e == Errno::EAGAIN => {
-                ctx.waker().wake_by_ref();
-                Poll::Pending
-            },
-            // Operation failed.
-            Err(e) => {
-                warn!("failed to accept connection ({:?})", e);
-                Poll::Ready(Err(Fail::new(e as i32, "operation failed")))
+
+            // Operation not completed, thus parse errno to find out what happened.
```
BrianZill, line 113: I would put all these comments inside the block to which they apply, as they describe state that is only true inside the block.

src/rust/catcollar/futures/connect.rs
```diff
@@ -69,22 +71,34 @@ impl Future for ConnectFuture {
     /// Polls the underlying connect operation.
     fn poll(self: Pin<&mut Self>, ctx: &mut Context<'_>) -> Poll<Self::Output> {
         let self_: &mut ConnectFuture = self.get_mut();
-        match socket::connect(self_.fd as i32, &self_.addr) {
+        match unsafe {
+            libc::connect(
+                self_.fd,
+                (&self_.sockaddr as *const libc::sockaddr_in) as *const libc::sockaddr,
+                mem::size_of_val(&self_.sockaddr) as u32,
```
BrianZill, line 78: While this will work, since sizeof sockaddr >= sizeof sockaddr_in, this really should be the latter and not the former.

src/rust/catcollar/futures/connect.rs
```diff
@@ -69,22 +71,34 @@ impl Future for ConnectFuture {
...
-                warn!("failed to establish connection ({:?})", errno);
-                Poll::Ready(Err(Fail::new(errno as i32, "operation failed")))
+
+            // Operation not completed, thus parse errno to find out what happened.
+            _ => {
+                let errno: libc::c_int = unsafe { *libc::__errno_location() };
+
+                // Operation in progress.
+                if errno == libc::EINPROGRESS || errno == libc::EALREADY {
+                    ctx.waker().wake_by_ref();
```
BrianZill, line 93: Keep the trace! message?

src/rust/catnap/futures/accept.rs
```diff
@@ -62,46 +66,63 @@ impl AcceptFuture {
...
 impl Future for AcceptFuture {
     type Output = Result<RawFd, Fail>;
 
     /// Polls the target [AcceptFuture].
     fn poll(self: Pin<&mut Self>, ctx: &mut Context<'_>) -> Poll<Self::Output> {
-        let self_: &AcceptFuture = self.get_mut();
-        match socket::accept(self_.fd as i32) {
+        let self_: &mut AcceptFuture = self.get_mut();
+        match unsafe {
+            let mut address_len: libc::socklen_t = 0;
```
BrianZill, line 81: This one too.

src/rust/catnap/futures/connect.rs
```diff
@@ -68,21 +70,34 @@ impl Future for ConnectFuture {
     /// Polls the target [ConnectFuture].
     fn poll(self: Pin<&mut Self>, ctx: &mut Context<'_>) -> Poll<Self::Output> {
         let self_: &mut ConnectFuture = self.get_mut();
-        match socket::connect(self_.fd as i32, &self_.addr) {
+        match unsafe {
+            libc::connect(
+                self_.fd,
+                (&self_.sockaddr as *const libc::sockaddr_in) as *const libc::sockaddr,
+                mem::size_of_val(&self_.sockaddr) as u32,
```
BrianZill, line 77: Same comment as for the catcollar version.

src/rust/catnap/futures/pop.rs
```diff
@@ -78,34 +81,40 @@ impl Future for PopFuture {
     /// Polls the target [PopFuture].
     fn poll(self: Pin<&mut Self>, ctx: &mut Context<'_>) -> Poll<Self::Output> {
         let self_: &mut PopFuture = self.get_mut();
-        let mut bytes: [u8; POP_SIZE] = [0; POP_SIZE];
-        match socket::recvfrom::<SockaddrStorage>(self_.fd, &mut bytes[..]) {
+        match unsafe {
+            let mut addrlen: libc::socklen_t = 0;
```
BrianZill, line 85: Another one.

src/rust/catnap/futures/pop.rs
```diff
@@ -78,34 +81,40 @@ impl Future for PopFuture {
...
+            libc::recvfrom(
+                self_.fd,
+                (self_.buf.as_mut_ptr() as *mut u8) as *mut libc::c_void,
+                self_.buf.len(),
+                libc::MSG_DONTWAIT,
```
BrianZill, line 90: Why this change?  Shouldn't this fd already be set_nonblock?

src/rust/catnap/futures/push.rs
```diff
@@ -66,21 +63,37 @@ impl Future for PushFuture {
     /// Polls the target [PushFuture].
     fn poll(self: Pin<&mut Self>, ctx: &mut Context<'_>) -> Poll<Self::Output> {
         let self_: &mut PushFuture = self.get_mut();
-        match socket::send(self_.fd, &self_.buf[..], socket::MsgFlags::empty()) {
+        match unsafe {
+            libc::sendto(
```
BrianZill, line 67: Why sendto instead of send here?

src/rust/catcollar/mod.rs
```diff
@@ -162,9 +160,21 @@ impl CatcollarLibOS {
...
             Some(&fd) => {
-                let addr: SockaddrStorage = parse_addr(local);
-                socket::bind(fd, &addr).unwrap();
-                Ok(())
+                let sockaddr: libc::sockaddr_in = linux::socketaddrv4_to_sockaddr_in(&local);
+                match unsafe {
+                    libc::bind(
+                        fd,
+                        (&sockaddr as *const libc::sockaddr_in) as *const libc::sockaddr,
+                        mem::size_of_val(&sockaddr) as u32,
```
BrianZill, line 168: Should be sizeof sockaddr_in.

src/rust/catcollar/futures/connect.rs
BrianZill: Nevermind, I see `sockaddr` is a `libc::sockaddr_in`, so this is ok.

src/rust/catnap/futures/connect.rs
```diff
@@ -68,21 +70,34 @@ impl Future for ConnectFuture {
     /// Polls the target [ConnectFuture].
     fn poll(self: Pin<&mut Self>, ctx: &mut Context<'_>) -> Poll<Self::Output> {
         let self_: &mut ConnectFuture = self.get_mut();
-        match socket::connect(self_.fd as i32, &self_.addr) {
+        match unsafe {
+            libc::connect(
```
iyzhang, line 74: I noticed that every libc invokation requires an unsafe block. Is that because these functions are unsafe? Are the nix ones safe somehow?

src/rust/catnap/futures/pop.rs
```diff
@@ -31,34 +25,43 @@ use ::std::{
...
 
 /// Associate Functions for Pop Operation Descriptors
 impl PopFuture {
     /// Creates a descriptor for a pop operation.
     pub fn new(qd: QDesc, fd: RawFd) -> Self {
-        Self { qd, fd }
+        Self {
+            qd,
+            fd,
+            sockaddr: unsafe { mem::zeroed() },
```
iyzhang, line 62: Is this unsafe block necessary?

src/rust/catnap/futures/mod.rs
```diff
@@ -1,19 +1,19 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-//==============================================================================
+//======================================================================================================================
```
iyzhang, line 4: Not sure if this is intentional, but this file was only changed to extend the comment bars. I don't mind having style fixes in this PR, but it should probably have its own commit.

src/rust/catcollar/futures/mod.rs
```diff
@@ -1,19 +1,19 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-//==============================================================================
+//======================================================================================================================
```
iyzhang, line 4: Same on this file. We should be sure to separate style changes from code changes.

src/rust/catcollar/futures/accept.rs
ppenna: You are correct. I have fixed this. Thanks for checking this.

src/rust/catnap/futures/accept.rs
ppenna: You are correct. I have fixed this. Thanks for checking this.

src/rust/catnap/futures/pop.rs
ppenna: You are correct. I have fixed this. Thanks for checking this.

src/rust/catcollar/futures/accept.rs
ppenna: Sure. I moved them in there.

src/rust/catcollar/futures/connect.rs
ppenna: Thanks

src/rust/catcollar/futures/connect.rs
ppenna: Sure! Thanks for noting.

src/rust/catnap/futures/pop.rs
ppenna: When creating the socket the `set_nonblock()` call may not succeed. In that case the socket is still created, but it would be a blocking one. By supplying the `libc::MSG_DONTWAIT` we are doing our best to make any asynchronous operation in a blocking socket.

src/rust/catnap/futures/push.rs
ppenna: For two reasons:
1. To be able to set the `MSG_DONTWAIT` flag (see my reply from above)
2. To intentionally align up the logic of `PushFuture` and `PopFuture`.

For (2) I do have a change on my queue that unifies the logic of both co-routines.

src/rust/catnap/futures/connect.rs
ppenna: Yes, invoking `libc` functions is unsafe. Functions from nix calls wrap around and provide a "safe" interface.

src/rust/catnap/futures/pop.rs
ppenna: Yes it is: https://doc.rust-lang.org/std/mem/fn.zeroed.html

src/rust/catnap/futures/mod.rs
ppenna: I was following our agreement that if something would be off the "coding style" it could be cleaned up. Anyways, I have reverted these changes.

src/rust/catcollar/futures/mod.rs
ppenna: I was following our agreement that if something would be off the "coding style" it could be cleaned up. Anyways, I have reverted these changes.

src/rust/catnap/futures/mod.rs
iyzhang: Yes, that is fine but I just wanted a separate commit in the PR for the style changes.

src/rust/catcollar/futures/accept.rs
BrianZill: Great, thanks!

src/rust/catcollar/futures/accept.rs
BrianZill: Thanks!

src/rust/catnap/futures/pop.rs
BrianZill: Okay.

src/rust/catnap/futures/push.rs
BrianZill: Okay.

BrianZill (approved):
Looks good now.  Thanks for removing this external dependency!

Issue #426, opened 2023-01-16
[build] Drop Dependency on `derive_more`


ppenna:
## Context

We should drop dependency on `derive_more` crate.

Issue #425, opened 2023-01-16
[build] Drop Dependency on `num-derive`


ppenna:
## Context

We should drop dependency on `num-derive` crate.

Issue #424, opened 2023-01-16
[build] Drop Dependency on `num-traits`


ppenna:
## Context

We should drop dependency on `num-traits` crate.

PR #427, opened 2023-01-16
[build] Drop Dependency on Some Crates

ppenna:
## Description

- This PR closes https://github.com/demikernel/demikernel/issues/424
- This PR closes https://github.com/demikernel/demikernel/issues/425
- This PR closes https://github.com/demikernel/demikernel/issues/426

## Summary of Changes

- Dropped dependency on `num-derive`
- Dropped dependency on `num-traits`
- Dropped dependency on `derive_more`


src/rust/runtime/queue/qdesc.rs
```diff
@@ -6,23 +6,37 @@
...
 
 //==============================================================================
 // Trait Implementations
 //==============================================================================
 
-/// Convert Trait Implementation for Signed 32-bit Integers
 impl From<QDesc> for i32 {
+    /// Converts a [QDesc] to a [i32].
     fn from(val: QDesc) -> Self {
         val.0 as i32
```
BrianZill, line 19: On a 64-bit machine, this conversion will lose half of the QDesc's bits.

src/rust/runtime/queue/qdesc.rs
```diff
@@ -6,23 +6,37 @@
 //==============================================================================
 
 /// IO Queue Descriptor
-#[derive(From, Into, Debug, Eq, PartialEq, Hash, Copy, Clone)]
+#[derive(Debug, Eq, PartialEq, Hash, Copy, Clone)]
 pub struct QDesc(usize);
```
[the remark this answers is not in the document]
ppenna: No there is none and this is something that we inherited from the research prototype. Actually, the C API uses a `int` to refer to Queue Descriptors in the API:

https://github.com/demikernel/demikernel/blob/c39b8f48eb9ddfd558edacf21614ec2d114dca6f/include/demi/types.h#L79-L92

May I open an issue to work on this?

src/rust/runtime/queue/qdesc.rs
ppenna: Yes, you are right, and this existed before as well =( See my previous comment.

src/rust/runtime/queue/qdesc.rs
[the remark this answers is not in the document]
BrianZill: Yes, if fixing this is more involved, then go ahead and punt it to a new Issue to be fixed later.

src/rust/runtime/queue/qdesc.rs
BrianZill: Yes, this is directly related to the usize issue, so if you push that to a new Issue, this should get resolved there as well.

src/rust/runtime/queue/qdesc.rs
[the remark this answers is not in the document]
ppenna: Sure, I have created this issue: https://github.com/demikernel/demikernel/issues/430

src/rust/runtime/queue/qdesc.rs
ppenna: Sure, I have created this issue: https://github.com/demikernel/demikernel/issues/430

src/rust/inetstack/protocols/ethernet2/protocol.rs
```diff
@@ -1,26 +1,39 @@
...
     type Error = Fail;
 
-    fn try_from(n: u16) -> Result<Self, Fail> {
-        match FromPrimitive::from_u16(n) {
-            Some(n) => Ok(n),
-            None => Err(Fail::new(ENOTSUP, "unsupported ETHERTYPE")),
+    /// Attempts to convert a [u16] into an [EtherType2].
+    fn try_from(value: u16) -> Result<Self, Self::Error> {
+        match value {
+            0x806 => Ok(EtherType2::Arp),
```
iyzhang, line 33: I assume that these constants are defined in a RFC somewhere. It might be good to define it as a const and a reference to the RFC, rather than just having these bare numbers.

src/rust/inetstack/protocols/ip/protocol.rs
```diff
@@ -26,18 +24,20 @@ pub enum IpProtocol {
...
 impl TryFrom<u8> for IpProtocol {
     type Error = Fail;
 
-    fn try_from(n: u8) -> Result<Self, Fail> {
-        match FromPrimitive::from_u8(n) {
-            Some(n) => Ok(n),
-            None => Err(Fail::new(ENOTSUP, "unsupported IPv4 protocol")),
+    fn try_from(value: u8) -> Result<Self, Self::Error> {
+        match value {
+            0x01 => Ok(IpProtocol::ICMPv4),
```
iyzhang, line 37: Same here, it would be nice to not just have these bare numbers with no explanation.

src/rust/inetstack/protocols/ethernet2/protocol.rs
ppenna: Tese are defined in the enum itself, but I cannot access them in here. The function is precisely doing the conversion. Let me know if you know a different way of doing that in pure rust. The crates used before provide this "facility" at cost of non-standard 3rd party.

src/rust/inetstack/protocols/ip/protocol.rs
ppenna: Tese are defined in the enum itself, but I cannot access them in here. The function is precisely doing the conversion. Let me know if you know a different way of doing that in pure rust. The crates used before provide this "facility" at cost of non-standard 3rd party.

src/rust/inetstack/protocols/ethernet2/protocol.rs
iyzhang: I just meant do a list of constants at the beginning of file and then use the constants in the code with a link to the enum or RFC.

iyzhang (approved):

src/rust/inetstack/protocols/ethernet2/protocol.rs
BrianZill: I agree.  Defining them as local constants, with a comment describing the authoritative source, would be good.

BrianZill (approved):
Okay.

src/rust/inetstack/protocols/ip/protocol.rs
ppenna: Fixed

Issue #218, opened 2022-01-26
[inetstack] Handle Cancellation of Ping Requests


ppenna:
Description
-------------

When sending ping requests, we spawn a coroutine to wait for the corresponding response and we enqueue this co-routine for later processing.

Unfortunately, however, if the remote peer does not reply the ping or the ping packet gets lost, the queue of pending receive coroutines will grow indefinitely.

https://github.com/demikernel/demikernel/blob/c2c867505a07dc4ac3831fc03d2163202997a0e2/src/rust/inetstack/protocols/icmpv4/peer.rs#L289-L292

Proposed Solution
---------------------

We should drain the queue of pending responses from time to time.

PR #428, opened 2023-01-16
[inetstack] Handle Cancellation of Ping Requests

ppenna:
## Description

This commit fixes https://github.com/demikernel/demikernel/issues/218

# Summary of Changes

- Handled cancellation for ping requests.
- Added missing type annotations.


BrianZill (approved):
Looks fine.

PR #429, opened 2023-01-18
[tools] Initial Azure Tables support

iyzhang:
This change adds support for running a suite of tests and reporting the results in Azure Tables. Summary of changes:
* Add installation of azure-data-tables python package. This is only needed on the machine running the tests, but I thought there was no harm installing it along with other dependencies.
* Demikernel CI python script changed to module. We can now include the CI script functionality in other files, especially the run_pipeline method, which can be used to trigger a specific set of tests. run_pipeline now returns a hashmap of test names and a boolean reflecting whether the test passed, instead of a packed bit string. 
* A new Azure Tables python script. This script takes a similar set of arguments to the CI script but additionally will take an Azure connection string and table name to log the results of the test runs. Unlike the CI script, if no specific test parameters are specified, then all tests will be run. For example, if no libOS is specified, then all libOSes will be tested. If neither release nor debug is specified, then both will be tested.

We may consider later combining this script into the CI script as there is some overlapping code.

### DIFF
--- a/scripts/setup/debian.sh
+++ b/scripts/setup/debian.sh
@@ -10,4 +10,6 @@ PACKAGES="librdmacm-dev libmnl-dev build-essential clang libnuma-dev pkg-config 
 apt-get update
 apt-get -y install $PACKAGES
 
-pip3 install pyelftools
+# This dependency is only needed on machines that drive the Azure Tables script
+# for continuous integration 
+pip3 install azure-data-tables
--- a/tools/azure_tables.py
+++ b/tools/azure_tables.py
@@ -1,0 +1,111 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+import sys
+import argparse
+from azure.data.tables import TableServiceClient
+import demikernel_ci
+
+def read_args() ->argparse.Namespace:
+    description: str = ""
+    description += "Use this utility automate regression testing of Demikernel on a pair of remote host machines and report the results to Azure Tables.\n"
+    description += "Before using this utility, ensure that you have correctly setup the development environment on the remote machines and created the Azure table for storing the results.\n"
+    description += "For more information, check out the README.md file of the project."
+
+    # Initialize parser.
+    parser = argparse.ArgumentParser(prog="azure_tables.py", description=description)
+
+    # Host options.
+    parser.add_argument("--server", required=True, help="set server host name")
+    parser.add_argument("--client", required=True, help="set client host name")
+
+    # Build options.
+    parser.add_argument("--repository", required=True, help="set location of target repository in remote hosts")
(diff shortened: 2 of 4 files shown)


tools/demikernel_ci.py
```diff
@@ -372,9 +351,13 @@ def main():
...
     client_addr: str = args.client_addr if test_system else ""
 
-    status: int = run_pipeline(repository, branch, libos, is_debug, server,
-                               client, test_unit, test_system, server_addr, client_addr, delay, config_path, enable_nfs)
-    sys.exit(status)
+    status: dict = run_pipeline(repository, branch, libos, is_debug, server,
+                                client, test_unit, test_system, server_addr,
+                                client_addr, delay, config_path, enable_nfs)
+    if False in status.values():
+        sys.exit(-1)
```
ppenna, line 358: Could you add a print statement on the actual tests that have failed here? This is important for further integration with GitHub Actions.

tools/azure_tables.py
```diff
@@ -0,0 +1,109 @@
+import os
+import sys
+import argparse
+from azure.data.tables import TableServiceClient
+import demikernel_ci
```
ppenna, line 5: Please, add our copyright notice:

```
# Copyright (c) Microsoft Corporation.
# Licensed under the MIT license.
```

tools/demikernel_ci.py
iyzhang: It should print the usual pass and fail statements since they are produced by the wait_and_report function.

tools/demikernel_ci.py
```diff
@@ -321,7 +300,7 @@ def read_args() -> argparse.Namespace:
     description += "For more information, check out the README.md file of the project."
 
     # Initialize parser.
-    parser = argparse.ArgumentParser(prog="demikernel-ci.py", description=description)
+    parser = argparse.ArgumentParser(prog="demikernel_ci.py", description=description)
```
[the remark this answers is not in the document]
iyzhang: done

scripts/setup/debian.sh
```diff
@@ -11,3 +11,4 @@ apt-get update
 apt-get -y install $PACKAGES
 
 pip3 install pyelftools
```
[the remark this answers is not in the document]
iyzhang: I don't know! I can remove it.

tools/azure_tables.py
iyzhang: done

scripts/setup/debian.sh
```diff
@@ -11,3 +11,4 @@ apt-get update
...
+pip3 install azure-data-tables
```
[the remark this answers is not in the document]
iyzhang: done

PR #431, opened 2023-01-19
[tools] Fixing flag that fails to run any tests in Azure tables script

iyzhang:
This PR fixes a bug in the azure-tables scripts that cause no tests to run when no make flags are given. I also moved catnip to the end of the test sequence per our discussion today, in case DPDK has any issues that wedge the machine.


ppenna (approved):

ppenna:
Thanks!

BrianZill (approved):
I don't know enough about Python to understand how these two if statement combinations differ, but it looks okay to me.

Issue #432, opened 2023-01-19
[catcollar] io_uring fails on bare metal


iyzhang:
## Description

All system tests for catcollar on bare metal fail right now.

## How to Reproduce

1. Run this command on the server side: `make test-system-rust LIBOS=catcollar TEST=udp-ping-pong RUST_BACKTRACE=full CONFIG_PATH=/demikernel/config.yaml ARGS='--server 198.19.200.1:12345 198.19.200.2:23456'`
2. Run this command on the client side: `make test-system-rust LIBOS=catcollar TEST=udp-ping-pong RUST_BACKTRACE=full CONFIG_PATH=/demikernel/config.yaml ARGS='--client 198.19.200.2:23456 198.19.200.1:12345'`

Same issue occurs with this run of the azure tables script:
`python3 ../tools/azure_tables.py --server prometheus1 --client prometheus2 --server-addr 198.19.200.1 --client-addr 198.19.200.2 --test-system --libos catcollar --branch dev --repository /demikernel/testing/demikernel --config-path /demikernel/config.yaml --connection-string XXX --table-name prometheus`

## Logs

Server side output
```
timeout 30 /demikernel/testing/demikernel/bin/examples/rust/udp-ping-pong.elf --server 198.19.200.1:12345 198.19.200.2:23456
thread 'main' panicked at 'unexpected result', examples/rust/udp-ping-pong.rs:108:18
stack backtrace:
   0:     0x55e7bfbc38ac - std::backtrace_rs::backtrace::libunwind::trace::h58c649c53097142b
                               at /rustc/b70baa4f922a1809d79caeaeb902800c3be283b9/library/std/src/../../backtrace/src/backtrace/libunwind.rs:93:5
   1:     0x55e7bfbc38ac - std::backtrace_rs::backtrace::trace_unsynchronized::h8a2218341115550a
                               at /rustc/b70baa4f922a1809d79caeaeb902800c3be283b9/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x55e7bfbc38ac - std::sys_common::backtrace::_print_fmt::hea4b2f7f6d570ca4
                               at /rustc/b70baa4f922a1809d79caeaeb902800c3be283b9/library/std/src/sys_common/backtrace.rs:65:5
   3:     0x55e7bfbc38ac - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h8d8d34f4e85c6a60
[...]

PR #436, opened 2023-01-19
[catcollar] Bad Logic for Async Ongoing Operations

ppenna:
## Description

This PR fixes https://github.com/demikernel/demikernel/issues/432

## Summary of Changes

- Fixed memory leak for `iovec` structure in `msghdr` - we were not retrieving the correct heap-allocated pointer.
- Fixed logic for handling ongoing async operations - before https://github.com/demikernel/demikernel/pull/419 the socket was operating on sync mode.


iyzhang (approved):
I'm not sure I understand enough about catcollar to understand all of the changes, but it looks fine to me and fixes the failing tests

Issue #364, opened 2022-12-11
[pdpix] Remote Address Is Not Set In `demi_accept_result_t`


ArchangelSDY:
Description
========

The `addr` in `demi_accept_result_t` seems always to be zero in an accept result. Is this a bug or am I doing this wrong? What's the correct way to obtain remote addr?

How to Reproduce
=============

1. `listen` and `accept`.
2. `wait`
3. Check `qr.qr_value.ares.addr`.

Screenshots
=========

Expected Behavior
============

Should be able to get a non-null remote addr after `accept`.

Related Issues
==========

- https://github.com/demikernel/demikernel/issues/402

## Updates

All network LibOSes fill in the `demikernel_qresult_t` with zeros:

https://github.com/demikernel/demikernel/blob/f1e518e1b8400b7934a27ba2cb595137954d34a6/src/rust/catcollar/mod.rs#L404-L418

https://github.com/demikernel/demikernel/blob/f1e518e1b8400b7934a27ba2cb595137954d34a6/src/rust/catnap/mod.rs#L391-L405

https://github.com/demikernel/demikernel/blob/f1e518e1b8400b7934a27ba2cb595137954d34a6/src/rust/catnap/mod.rs#L391-L405

https://github.com/demikernel/demikernel/blob/f1e518e1b8400b7934a27ba2cb595137954d34a6/src/rust/catpowder/interop.rs#L31-L45

ppenna:
@ArchangelSDY could you please share what LibOS you are using?

ArchangelSDY:
@ppenna I'm using catnip.

PR #438, opened 2023-01-20
[pdpix] Remote Address Is Not Set In `demi_accept_result_t`

ppenna:
## Description

This commit fixes https://github.com/demikernel/demikernel/issues/364

## Summary of Changes

- Changed Futures system in all LibOSes to output the accepted IPv4 address.
- Fixed method for crafting a `demi_accept_result_t` in all LibOses.
- Fixed concerned unit tests.

### DIFF
--- a/src/rust/catnip/interop.rs
+++ b/src/rust/catnip/interop.rs
@@ -39,12 +39,19 @@ pub fn pack_result(rt: Rc<DPDKRuntime>, result: OperationResult, qd: QDesc, qt: 
             qr_qt: qt,
             qr_value: unsafe { mem::zeroed() },
         },
-        OperationResult::Accept(new_qd) => {
-            let sin = unsafe { mem::zeroed() };
-            let qr_value = demi_qr_value_t {
+        OperationResult::Accept((new_qd, addr)) => {
+            let saddr: SockAddrIn = {
+                SockAddrIn {
+                    sin_family: AF_INET as u16,
+                    sin_port: addr.port().into(),
+                    sin_addr: create_sin_addr(&addr.ip().octets()),
+                    sin_zero: create_sin_zero(),
+                }
+            };
+            let qr_value: demi_qr_value_t = demi_qr_value_t {
                 ares: demi_accept_result_t {
                     qd: new_qd.into(),
-                    addr: sin,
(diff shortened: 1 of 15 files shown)


src/rust/catnip/interop.rs
```diff
@@ -39,12 +39,20 @@ pub fn pack_result(rt: Rc<DPDKRuntime>, result: OperationResult, qd: QDesc, qt:
             qr_qt: qt,
             qr_value: unsafe { mem::zeroed() },
         },
-        OperationResult::Accept(new_qd) => {
-            let sin = unsafe { mem::zeroed() };
-            let qr_value = demi_qr_value_t {
+        OperationResult::Accept((new_qd, addr)) => {
+            let saddr: SockAddrIn = {
+                // TODO: check the following byte order conversion.
```
BrianZill, line 44: I thought I'd do this TODO as part of this review.

I believe the byte-order here is inadvertently correct.  `addr` is a `std::net::SocketAddrV4`, whose `ip()` method returns a `std::net::Ipv4Addr`, whose `octets()` method returns a `[u8; 4]` where the order of the bytes contained therein is **big-endian**.  Line 48's `pal::create_sin_addr()` function appears to expect its input bytes in **little-endian** byte order.

This still works (but only on our little-endian machines!) because `create_sin_addr()` is buggy.  As written, it creates a native byte order u32 from little-endian bytes.  What it **should** be doing is creating a network (i.e., big-endian) byte order u32 from big-endian bytes.  In other words, it shouldn't reverse the byte order.  And on little-endian machines, where the native byte order is (no surprise) little-endian, creating a native byte order u32 from little-endian bytes doesn't reverse the byte order either.  So `create_sin_addr()`'s bug doesn't manifest itself on little-endian machines.  But on a big-endian machine, it would reverse the byte order, which would be wrong.

I'll file an Issue for the `create_sin_addr()` bug.  This code is correct.  So, you can remove this TODO comment (and the identical one on line 75 you presumably copied this from).

BrianZill (approved):
Looks good!

src/rust/catnip/interop.rs
ppenna: Sure.

PR #440, opened 2023-01-20
[runtime] Unify memory runtime

iyzhang:
Currently there is a lot of redundant code between the libOSes for the memory runtime. This changes moves the implementation of the MemoryRuntime trait for heap-allocated memory to the default implementation and leaves specially allocated memory in the relevant libOS. At this time, we only have a separate implementation for Catnip. 

Summary of changes:
* [runtime] Move implementation of heap-allocated DemiBuffers to runtime/memory/mod.rs
* [catcollar] Eliminate runtime/memory.rs and use default implementation of MemoryRuntime trait
* [catnap] Eliminate redundant code and use default implementation of MemoryRuntime trait
* [catnapw] Eliminate redundant code and use default implementation of MemoryRuntime trait
* [catpowder] Eliminate runtime/memory.rs and use default implementation of MemoryRuntime trait
* [catmem] Eliminate redundant code. 
 
Note: Catmem is a little awkward because it doesn't not contain a runtime, so directly implements the trait, which the demikernel layer then directly uses. Since the MemoryRuntime trait is only accessible from Catmem and not the Demikernel layer, I had to wrap some of the functions, rather than directly using the default. I will fix this architecture eventually.


ppenna (approved):
Looks great.

@iyzhang could you open issues for the next work itens that mentioned in the PR's description?

Issue #430, opened 2023-01-18
[runtime] Queue Descriptors Should Be 32 Bit


ppenna:
## Description

Our Rust definition for a Queue Descriptor is based on a `usize`, which is 64 bits in 64-bit machines.

This introduces three problems:
1. It does not match the C API that relies on a `int`
2. It introduces some lossy conversions
3. It uses more memory.

https://github.com/demikernel/demikernel/blob/5d9446dd70c3ef602595f3b44675c1e3588d1624/src/rust/runtime/queue/qdesc.rs#L8-L10

## Proposed Solution

Queue descriptor should be based on a 32-bit values to avoid all the aforementioned issues.

PR #441, opened 2023-01-23
[runtime] 32-Bit Queue Descriptors

ppenna:
## Description

This PR closes https://github.com/demikernel/demikernel/issues/430.

## Summary of Changes

- Changed `QDesc` to use a `u32` type representation for queue descriptors.
- Fixed `IoQueueTable` accordingly.
- Fixed concerned unit tests.

### DIFF
--- a/src/rust/runtime/queue/mod.rs
+++ b/src/rust/runtime/queue/mod.rs
@@ -44,7 +44,7 @@ impl IoQueueTable {
     /// When Demikernel is interposing system calls of the underlying OS
     /// this offset enables us to distinguish I/O queue descriptors from
     /// file descriptors.
-    const BASE_QD: usize = 1_000_000;
+    const BASE_QD: u32 = 1_000_000;
 
     /// Creates an I/O queue descriptors table.
     pub fn new() -> Self {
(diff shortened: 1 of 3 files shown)


src/rust/runtime/queue/mod.rs
```diff
@@ -44,7 +44,7 @@ impl IoQueueTable {
     /// When Demikernel is interposing system calls of the underlying OS
     /// this offset enables us to distinguish I/O queue descriptors from
     /// file descriptors.
```
BrianZill, line 46: Is this really necessary for something?  (I know this is existing code, but we've shrunk the number space from 2^64 to 2^32.  And now this limits the queue descriptor space even more).

src/rust/runtime/queue/mod.rs
```diff
@@ -44,7 +44,7 @@ impl IoQueueTable {
...
-    const BASE_QD: usize = 1_000_000;
+    const BASE_QD: u32 = 1_000_000;
```
BrianZill, line 47: At first I thought this number was arbitrary, but now I'm seeing that it is roughly 2^20, which apparently is the maximum number of file descriptors that modern Linux allows (requires explicit configuration as this isn't the default).  One can go beyond this yet on Linux, but it takes more than adjusting the well-documented way to increase the file descriptor limit.

If this value is chosen to match the Linux limitation, then it probably should be in the PAL.  It should probably also have a comment explaining why it is the value it is.

src/rust/runtime/queue/mod.rs
```diff
@@ -54,28 +54,33 @@ impl IoQueueTable {
     /// Allocates a new entry in the target I/O queue descriptors table.
     pub fn alloc(&mut self, qtype: u32) -> QDesc {
         let index: usize = self.table.insert(qtype);
-        QDesc::from(index + Self::BASE_QD)
+
+        // Ensure that the allocation would yield to a safe conversion between usize to u32.
+        // Note: This imposes a limit on the number of open queue descriptors in u32::MAX.
+        assert!(index <= u32::MAX as usize, "I/O descriptors table overflow");
```
BrianZill, line 60: Shouldn't this assert be for `index + Self::BASE_QD <= u32::MAX`?  Because after that, line 62 below could fail.

src/rust/runtime/queue/mod.rs
ppenna: Yes it is and this will be clear when the SHIM layer comes into play.

src/rust/runtime/queue/mod.rs
ppenna: This is a pending fix that I have not yet submitted. This PR is about changing the type of queue descriptors.

src/rust/runtime/queue/mod.rs
ppenna: Yest it should. It is fixed now.

src/rust/runtime/queue/mod.rs
BrianZill: Got it, thanks.  Wanted to make sure this wasn't just a debugging aid.

src/rust/runtime/queue/mod.rs
BrianZill: Yes, I know.  I was asking these questions because the new queue descriptor type has fewer bits, but still has to be big enough to hold (essentially) both the file descriptor namespace and the actual queue descriptor namespace.  This BASE_QD value is what determines how much of the u32's namespace goes to each.

src/rust/runtime/queue/mod.rs
```diff
@@ -54,28 +54,36 @@ impl IoQueueTable {
...
     pub fn alloc(&mut self, qtype: u32) -> QDesc {
         let index: usize = self.table.insert(qtype);
-        QDesc::from(index + Self::BASE_QD)
+
+        // Ensure that the allocation would yield to a safe conversion between usize to u32.
+        // Note: This imposes a limit on the number of open queue descriptors in u32::MAX.
+        assert!(
+            (index as u32) + Self::BASE_QD <= u32::MAX,
+            "I/O descriptors table overflow"
+        );
```
BrianZill, line 63: Since `Slab` appears to allocate indexes sequentially, I think it's reasonable to just assert here as we're highly unlikely to ever exceed the max value.  Otherwise, we'd want `alloc()` to optionally return a failure code.

src/rust/runtime/queue/mod.rs
BrianZill: Great, thanks!

BrianZill (approved):
Looks good to me.

src/rust/runtime/queue/mod.rs
ppenna: Yes. Furthermore `Slab` does try to re-use empty entries.

src/rust/runtime/queue/mod.rs
ppenna: Oh sure! Let me give you some extra information: we will decrease this value :). It should be less than `FD_SETSIZE` [[1](https://www.man7.org/linux/man-pages/man2/select.2.html)] to play nice with `epoll()`.

Issue #407, opened 2023-01-10
[catnap] Merge `PushtoFuture` and `PushFuture`


ppenna:
## Context

In Catnap LibOS the `PushtoFuture` and `PushFuture` abstraction is roughly the same.

## Proposed Solution

 We should merge logic in  `PushtoFuture` and `PushFuture` in a single abstraction

PR #442, opened 2023-01-23
[catnap] Merge `Pushto` and `Push` Futures

ppenna:
## Description

 This PR closes https://github.com/demikernel/demikernel/issues/407.

## Summary of Changes

- Merged `PushtoFuture` and `PushFuture` to void code replication.


iyzhang (changes requested):
We should standardize on push, instead of pushto, because I would like connected, not unconnected, Demikernel queues to be the more general case.

ppenna:
> We should standardize on push, instead of pushto, because I would like connected, not unconnected, Demikernel queues to be the more general case.

Sure, but can you be more specific? Do mean rename "Pushto" to "Push" or is it something else?

iyzhang:
Yes, just rename and make the default API without an address.

iyzhang (approved):
Perfect, thanks!

Issue #449, opened 2023-01-30
[runtime] Bad Offset for Queue Descriptors


ppenna:
## Description

When distributing queue descriptors to user applications, we offset the identifier so that it does not collide with file descriptors in Linux.

Currently, this offset is set as follows:

https://github.com/demikernel/demikernel/blob/9b21ac35f3023108ca64bfc4c80f7516b1baa77a/src/rust/runtime/queue/mod.rs#L42-L47

Unfortunately, this value does not play nice with `select()` [[ref](https://www.man7.org/linux/man-pages/man2/select.2.html)]:

```plain
WARNING: select() can monitor only file descriptors numbers that
are less than FD_SETSIZE (1024)—an unreasonably low limit for
many modern applications—and this limitation will not change.
All modern applications should instead use poll(2) or epoll(7),
which do not suffer this limitation.
```

## Expected Behavior

The offset should be less than `FD_SETSIZE`.

PR #450, opened 2023-01-30
[runtime] Make `BASE_QD < FD_SETSIZE`

ppenna:
## Description

This PR fixes https://github.com/demikernel/demikernel/issues/449.

## Summary of Changes

- Changed the `BASE_QD` constant to `500` which is less than `FD_SETSIZE`.


iyzhang (approved):
Look fine to me. I would add some more info in the comments about why this particular number was chosen (e.g., that it is half of the 1024 max on select).

PR #464, opened 2023-02-07
[runtime] Remove Unused Buffer Types

BrianZill:
This PR removes the no longer used `Buffer` and `DataBuffer` types.


ppenna (approved):
Thanks for this!

Issue #465, opened 2023-02-08
[catloop] Malformed `OperationResult::Accept`


ppenna:
## Description

In Catloop LibOS we are badly constructing the `OperationResult::Acept` structure:

https://github.com/demikernel/demikernel/blob/4f22cfa203a2caf03d4175787ded4df8f2c94bae/src/rust/catloop/futures/mod.rs#L71-L77

The `qd` and `new_qd` are swapped.

## Expected Behavior

Code should be:

```rust
Operation::Accept(
    qd,
    FutureResult {
        future: _,
        done: Some(Ok((new_qd, remote, duplex_pipe))),
    },
) => (qd, OperationResult::Accept((new_qd, (remote, duplex_pipe)))),
```

PR #466, opened 2023-02-08
[catloop] Bad Operation Result for Accept

ppenna:
## Description

- This commit fixes https://github.com/demikernel/demikernel/issues/465


iyzhang (approved):
looks correct to me

PR #476, opened 2023-02-13
[examples] Improving TCP Echo Program

ppenna:
## Description

This PR improves the TCP Echo program in several points:

- Added dynamic payload generation on client side
- Added payload checking on client side
- Added support for running a running a fixed number of requests.


iyzhang (approved):
looks good to me

Issue #478, opened 2023-02-22
[test] System Level Test Program for Connection Establishment


ppenna:
## Context

We should have a system level test program for connection establishment.

## Proposed Solution

This test should exercise the following scenario:

- A client TCP program issues `N` connection requests to a TCP server program.
- The client TCP program waits for the `N` connection requests to be established.
- The client TCP program closes the `N` connections.

PR #479, opened 2023-02-22
[examples] Adding tcp-accept Program

ppenna:
## Description

This PR closes https://github.com/demikernel/demikernel/issues/478

## Summary of Changes

- Introduced a system level test program for connection establishment
- Added the new test program as an optional system level test in `demikernel_ci`


iyzhang (approved):
Thanks for putting this test together. It looks good.

Issue #480, opened 2023-02-24
[catmem] Don't Close Pipe When EoF is Reached


ppenna:
## Description

When we read end of file from a pipe, we close the concerned file descriptor.

Consequently, any operation on that queue descriptor, including `close()` fails:

https://github.com/demikernel/demikernel/blob/801dac0c82d44c77c870b05fdeed3a6d789ace0c/src/rust/catmem/mod.rs#L214-L217

## Expected Behavior

We should have a semantic behavior that matches [`read()`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/read.html).

When we read end of file from a pipe, we should set flag on that queue descriptor. Then, `push()` and `pop()` on that queue descriptor should fail with `ECONNRESET`, and `close()` should release the underlying resources associated to that queue descriptor.

PR #482, opened 2023-02-24
[catmem] Don't Close Pipe When EoF is Reached

ppenna:
## Description

This PR closes https://github.com/demikernel/demikernel/issues/480

## Summary of Changes

- Added a new example application to expose https://github.com/demikernel/demikernel/issues/480.
- Added new example application to `demikernel_ci`.
- Added Pipe abstraction to Catmem LibOS.
- Fixed https://github.com/demikernel/demikernel/issues/480.

### DIFF
--- a/examples/pipe-open/args.rs
+++ b/examples/pipe-open/args.rs
@@ -1,0 +1,104 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+//======================================================================================================================
+// Imports
+//======================================================================================================================
+
+use anyhow::Result;
+use clap::{
+    Arg,
+    ArgMatches,
+    Command,
+};
+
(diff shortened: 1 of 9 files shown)


iyzhang (approved):
The changes to catmem conflict with my changes in PR #486. It should be easy to replace the ring field in CatmemQueue without disturbing any of the rest of the code, but I'm not sure what order in which these two PRs should go in. Otherwise, the changes look good to me.

ppenna:
> The changes to catmem conflict with my changes in PR #486. It should be easy to replace the ring field in CatmemQueue without disturbing any of the rest of the code, but I'm not sure what order in which these two PRs should go in. Otherwise, the changes look good to me.

Yes, let us work together on your enhancement changes to make sure that they also cover this bug fix.

Issue #463, opened 2023-02-04
[script] Missing Build Dependency for DPDK


carvalhof:
In the commit `7fbeca2e65871ddbdd61978aa99385d5d3db8617` and fresh Ubuntu 20.04 server, installing the DPDK (https://github.com/demikernel/demikernel/blob/dev/doc/setup.md#2-install-third-party-requirements), I got the following error:

`buildtools/meson.build:52:8: ERROR: Problem encountered: missing python module: elftools`

I resolved this error by installing the `python3-pyelftools` package (`apt install python3-pyelftools`).

In the previous version of `scripts/setup/debian.sh`, there was the command (`pip3 install pyelftools`). Was it removed because it was only needed for the DPDK?

ppenna:
@iyzhang we have recently dropped this, right? It looks like this is a dependency required by meson. If you agree I'll rollback the change.

iyzhang:
Yes, it is a DPDK requirement, not Demikernel, so I *think* we talked about it and agreed that we shouldn't install it by default.

ppenna:
@iyzhang oh yes, thanks for refreshing my memory. How about adding that to `scripts/setup/dpdk.sh` ?

ppenna:
Hello @iyzhang any ideas on this?

PR #487, opened 2023-02-27
[script] Missing Build Dependency for DPDK

ppenna:
## Description

This PR closes https://github.com/demikernel/demikernel/issues/463


iyzhang (approved):
Yes, I think I took this out because we thought we didn't need it, but looks like it is still a dependency.

PR #490, opened 2023-02-28
[catloop] Several Bug Fixes for Connection Establishment / Shutdown

ppenna:
## Description

This PR fixes several bug fixes that were exposed by the `tcp-accept` program https://github.com/demikernel/demikernel/issues/478.

## Summary of Changes

- When closing duplex pipe, half-close underlying receiving pipe.
- Re-issue accept operation on bad connect request.
- Re-try on connection timeout.
- Half-Close duplex pipe when connecting.
- Wait for connect ACK when accepting a connection.

### DIFF
--- a/src/rust/catloop/futures/accept.rs
+++ b/src/rust/catloop/futures/accept.rs
@@ -174,23 +174,27 @@ fn listen_and_accept(
     // Check if a connection request arrived.
     if let Some(handle) = DuplexPipe::poll(&self_.catmem, qt_rx)? {
         // Check if this is a valid connection request.
-        check_connect_request(&self_.catmem, handle, qt_rx)?;
+        if !check_connect_request(&self_.catmem, handle, qt_rx).is_err() {
+            // Create underlying pipes before sending the port number through the
+            // control duplex pipe. This prevents us from running into a race
+            // condition were the remote makes progress faster than us and attempts
+            // to open the duplex pipe before it is created.
+            let duplex_pipe: Rc<DuplexPipe> = Rc::new(DuplexPipe::create_duplex_pipe(
+                self_.catmem.clone(),
+                &self_.ipv4,
+                self_.new_port,
+            )?);
 
-        // Create underlying pipes before sending the port number through the
-        // control duplex pipe. This prevents us from running into a race
-        // condition were the remote makes progress faster than us and attempts
-        // to open the duplex pipe before it is created.
-        let duplex_pipe: Rc<DuplexPipe> = Rc::new(DuplexPipe::create_duplex_pipe(
-            self_.catmem.clone(),
-            &self_.ipv4,
-            self_.new_port,
-        )?);
+            // Send port number.
+            let qt_tx: QToken = send_port_number(&self_.catmem, self_.control_duplex_pipe.clone(), self_.new_port)?;
 
-        // Send port number.
-        let qt_tx: QToken = send_port_number(&self_.catmem, self_.control_duplex_pipe.clone(), self_.new_port)?;
-
-        // Advance to next state in the connection establishment protocol.
-        self_.state = ServerState::Connect(qt_tx, duplex_pipe.clone());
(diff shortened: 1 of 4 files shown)


src/rust/catloop/futures/connect.rs
```diff
@@ -46,7 +46,7 @@ enum ClientState {
     InitiateConnectRequest,
     ConnectRequestSent(QToken),
     ConnectAckReceived(QToken),
-    Connected(SocketAddrV4, Rc<DuplexPipe>),
+    Connected(SocketAddrV4, Rc<DuplexPipe>, QToken),
```
iyzhang, line 49: Perhaps move this to the beginning to match the order of other enums like ServerState.

src/rust/catloop/futures/accept.rs
```diff
@@ -47,7 +47,7 @@ use ::std::{
 enum ServerState {
     ListenAndAccept(QToken),
     Connect(QToken, Rc<DuplexPipe>),
-    Connected(QDesc, SocketAddrV4, Rc<DuplexPipe>),
+    Connected(QToken, QDesc, SocketAddrV4, Rc<DuplexPipe>),
```
iyzhang, line 50: We might consider using named fields here since there are four of them and it might be hard to remember the order they should be listed later.

src/rust/catmem/mod.rs
```diff
@@ -112,6 +117,22 @@ impl CatmemLibOS {
         Ok(())
     }
 
+    /// Half-closes a memory queue.
```
iyzhang, line 120: It would be good to document what half-close means in this context. I think it drops the reference to the underlying pipe but doesn't close it. If that is the case, then we could rename it something more meaningful like free_qd()? I'm not sure that is the best name, but something like that.

iyzhang (changes requested):
Overall looks good. As I mentioned, I'm not sure why having a half-close/shutdown method is important because I don't quite understand what half-close means in this case. I think once that is clear, we are good.

src/rust/catloop/futures/accept.rs
ppenna: Sure!

src/rust/catloop/futures/connect.rs
ppenna: Sure!

src/rust/catmem/mod.rs
ppenna: I added a comment to give more details on this operation.

On the function name, we can have a discussion on it, but I don't believe that `free_qd()`, because this is actually the equivalent of disallowing further receptions on this pipe, like `shutdown(fd, SHUT_RD)`.

src/rust/catloop/futures/accept.rs
ppenna: Sure. I applied the change. let me know if you prefer it this way.

src/rust/catloop/futures/accept.rs
```diff
@@ -45,9 +45,19 @@ use ::std::{
 //======================================================================================================================
 
 enum ServerState {
-    ListenAndAccept(QToken),
-    Connect(QToken, Rc<DuplexPipe>),
-    Connected(QDesc, SocketAddrV4, Rc<DuplexPipe>),
+    ListenAndAccept {
```
iyzhang, line 48: Yes, I like this much better!

src/rust/catmem/mod.rs
iyzhang: Should this be close_endpoint()? Semantically, close and shutdown have the same meaning, so we should try to come up with a more descriptive name. It seems like it closes the local end of the pipe, so I think close_endpoint() could make sense or close_local().

iyzhang: OK, now that I have seen the documentation for shutdown on sockets, I agree the function should be called shutdown.

iyzhang (approved):
LGTM

Issue #491, opened 2023-03-01
[catnap] Missing free() for queue descriptors on close()


iyzhang:
## Description

We currently never free queue descriptors that we allocate for sockets.

## How to Reproduce

1. Call close() on queue descriptor in catnap
2. Look up queue descriptor in qtable. 

## Expected Behavior

We should not be able to look up a queue descriptor after the libOS has closed it.

ppenna:
@iyzhang I ACK that this is related to the following code snippet:

https://github.com/demikernel/demikernel/blob/df4255bcdc57b7456e08cdeda7016d31bffb648d/src/rust/catnap/mod.rs#L223-L233

PR #492, opened 2023-03-01
[catnap] Free queue descriptor on `close()`

iyzhang:
This commit fixes issue #491.


ppenna (approved):
Looks good to me. Thanks!

Issue #493, opened 2023-03-02
[catcollar] Does not remove socket from sockets table or free queue descriptor on close()


iyzhang:
## Description

While the close() function closes the underlying socket with a syscall, it does not remove the socket from the sockets table or free the queue descriptor.

## How to Reproduce

1. Call close() on qd.
2. Try to push to qd again.
 
## Expected Behavior

The libOS should not be able to find or push to the closed qd, so there should be an invalid queue descriptor error returned.

PR #494, opened 2023-03-02
[catcollar] Correctly Dree Data Structures on `close()`

iyzhang:
This fixes issue #493. I realized that catcollar does not remove the queue descriptor from the sockets table or free it from the IoQueueTable.


ppenna (approved):
Looks good. Thanks for fixing this.

PR #502, opened 2023-03-06
[runtime] Move OperationResult out of inetstack

iyzhang:
This change moves the OperationResult data structure out of the inetstack. This move lets us remove the dependency on the inetstack from the libOSes that do not use it. I also deprecated the QResult data structure as I believe it is not actually used. We can decide later whether to rename the OperationResult data structure or have separate result data structures for each libOS.


src/rust/runtime/queue/qresult.rs
```diff
@@ -7,16 +7,33 @@
...
 //==============================================================================
-// Enumerations
+// Structures
 //==============================================================================
 
+pub enum OperationResult {
+    Connect,
+    Accept((QDesc, SocketAddrV4)),
+    Push,
+    // TODO: Drop wrapping Option.
```
ppenna, line 30: The option is required because the underlying pop may fail to return the address of the remote.

src/rust/runtime/queue/qresult.rs
```diff
@@ -7,16 +7,33 @@
...
+    Pop(Option<SocketAddrV4>, DemiBuffer),
+    Failed(Fail),
+}
+
 /// Result for IO Queue Operations
+#[deprecated]
```
ppenna, line 36: What is deprecated on this?

ppenna (changes requested):

src/rust/runtime/queue/qresult.rs
iyzhang: It looks like we always just fill in the demi_qresult_t without using the Result, so I thought perhaps we could deprecate it. It looks like nothing in our examples or test directories depend on it.

src/rust/runtime/queue/qresult.rs
iyzhang: This comment was there from before. I can take it out however.

ppenna: Oh, sure! Could you remove it?

src/rust/runtime/queue/qresult.rs
ppenna: Agree! Could we have an issue for this?

iyzhang: Sure, I took it out of this PR!

src/rust/runtime/queue/qresult.rs
iyzhang: Yup, done

ppenna (approved):

Issue #507, opened 2023-03-07
[runtime] Enable Iteration over `IoQueueTable`


ppenna:
## Context

When performing some control-path operations, we should be able to traverse all queues that are registered in the underlying `IoQueueTable`.

For instance, this is required to check if there is a socket already bound to a given address.

## Proposed Solution

The `IoQueueTable` should support iteration over registered queues.

PR #513, opened 2023-03-07
[runtime] Iterator over IO Queue Table

ppenna:
## Description

This PR closes https://github.com/demikernel/demikernel/issues/507

## Summary of Changes

- Added a new function `get_values()` to `IoQueueTable`.


iyzhang (approved):
Looks good

Issue #497, opened 2023-03-02
[catloop] Do not Allow Multiple `bind()` to Same Address


ppenna:
## Description

We should not allow multiple calls to bind `bind()` on the same address.

https://github.com/demikernel/demikernel/blob/9c02963f23ea08be30465d4711d9e2791f1856f6/src/rust/catloop/mod.rs#L146-L165

PR #514, opened 2023-03-07
[catloop] Do not Allow Multiple `bind()` to Same Address

ppenna:
## Description

This PR fixes https://github.com/demikernel/demikernel/issues/497

## Summary of Changes

- Added extra check for whether the bind address is already in use.
- Fixed error value when trying to bind a bound socket to a different address from `EBADF` to `EINVAL`.

## Additional Information

This PR takes dependency on https://github.com/demikernel/demikernel/pull/513

### DIFF
--- a/src/rust/catloop/mod.rs
+++ b/src/rust/catloop/mod.rs
@@ -138,36 +138,58 @@ impl CatloopLibOS {
     pub fn bind(&mut self, qd: QDesc, local: SocketAddrV4) -> Result<(), Fail> {
         trace!("bind() qd={:?}, local={:?}", qd, local);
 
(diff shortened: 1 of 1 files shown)


src/rust/runtime/queue/mod.rs
```diff
@@ -10,7 +10,10 @@ mod qtype;
 // Imports
 //======================================================================================================================
 
-use ::slab::Slab;
+use ::slab::{
```
iyzhang, line 13: This seems to be subsumed by #513

iyzhang (approved):
Looks good as we mostly wrote this together.

Issue #510, opened 2023-03-07
[catcollar] Do not Allow Multiple `bind()` to Same Address


ppenna:
## Description

We should not allow multiple calls to `bind()` on the same address.

## Description

We should not allow multiple calls to `bind()` on the same address.

https://github.com/demikernel/demikernel/blob/bdb41ea54d4a75151dbb56e57fcd2e15b4d5c958/src/rust/catnap/mod.rs#L133-L162

Issue #509, opened 2023-03-07
[catcollar] Wrong Check for Bad `close()`


ppenna:
## Description

In `close()`, we are performing a wrong check for deciding whether the operation succeed.

https://github.com/demikernel/demikernel/blob/bdb41ea54d4a75151dbb56e57fcd2e15b4d5c958/src/rust/catcollar/mod.rs#L249-L250

PR #516, opened 2023-03-07
[catcollar] Do not Allow Multiple `bind()` to Same Address

ppenna:
## Description

- This PR fixes https://github.com/demikernel/demikernel/issues/509
- This PR fixes https://github.com/demikernel/demikernel/issues/510

## Summary of Changes

- Fixed check for bad `close()`
- Added extra check for whether the bind address is already in use.
- Fixed error value when trying to bind a bound socket to a different address from `EBADF` to `EINVAL`.

## Additional Information

This PR takes dependency on https://github.com/demikernel/demikernel/pull/513


iyzhang (approved):
Looks good. If it seems like we might always want to keep an address associated with the queue, then we might consider changing it in the Queue trait.

Issue #512, opened 2023-03-07
[tests] Closing Active Sockets using `safe_close_passive`


ppenna:
## Description

The `tcp_bad_listen()` and `tcp_bad_bind()` tests units are wrongly using `safe_close_passive()` to close active sockets.

In the code snippets bellow, `sockqd` is an active socket because the previous call to `listen()` did not succeed.

https://github.com/demikernel/demikernel/blob/bdb41ea54d4a75151dbb56e57fcd2e15b4d5c958/tests/rust/tcp.rs#L455-L462

https://github.com/demikernel/demikernel/blob/bdb41ea54d4a75151dbb56e57fcd2e15b4d5c958/tests/rust/tcp.rs#L476-L483

In the code snippets bellow,  `sockqd` is an active socket because no previous call to `listen()` is performed previously.

https://github.com/demikernel/demikernel/blob/bdb41ea54d4a75151dbb56e57fcd2e15b4d5c958/tests/rust/tcp.rs#L476-L483

https://github.com/demikernel/demikernel/blob/bdb41ea54d4a75151dbb56e57fcd2e15b4d5c958/tests/rust/tcp.rs#L476-L483

## Expected Behavior

We should use the `safe_close_active()` routine to close active sockets.

## Related Issues

This bug is exposed when #511 is fixed.

Issue #511, opened 2023-03-07
[inetstack] Fail to Close Bound and Unbound Sockets


ppenna:
## Description

When closing unbound and bound TCP sockets, that are sockets that are not bound to an address or are bound not in the listening, connecting, established or connection states, we fail:

https://github.com/demikernel/demikernel/blob/bdb41ea54d4a75151dbb56e57fcd2e15b4d5c958/src/rust/inetstack/protocols/tcp/peer.rs#L439-L449

## Expected Behavior

Closing bound sockets should be possible.

PR #517, opened 2023-03-07
[inetstack] Fail to Close Bound and Unbound Sockets

ppenna:
## Description

- This PR fixes https://github.com/demikernel/demikernel/issues/511
- This PR fixes https://github.com/demikernel/demikernel/issues/512

## Summary of Changes

- Fixed `close()` to succeed when closing bound and unbound sockets
- Fixed test code to use `safe_close_active()` to correctly close active sockets.

### DIFF
--- a/src/rust/inetstack/protocols/tcp/peer.rs
+++ b/src/rust/inetstack/protocols/tcp/peer.rs
@@ -434,24 +434,46 @@ impl TcpPeer {
         // 2. We do not remove the queue from the queue table.
         // As a result, we have stale closed queues that are labelled as closing. We should clean these up.
         // look up socket
-        let addr: SocketAddrV4 = match inner.qtable.borrow_mut().get_mut(&qd) {
+        let (addr, result): (SocketAddrV4, Result<(), Fail>) = match inner.qtable.borrow_mut().get_mut(&qd) {
             Some(InetQueue::Tcp(queue)) => {
                 match queue.get_socket() {
-                    // Closing an active socket
+                    // Closing an active socket.
                     Socket::Established(socket) => {
                         socket.close()?;
                         queue.set_socket(Socket::Closing(socket.clone()));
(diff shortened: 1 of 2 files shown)


src/rust/inetstack/protocols/tcp/peer.rs
```diff
@@ -434,24 +434,45 @@ impl TcpPeer {
...
+                    Socket::Closing(_) => {
+                        let cause: String = format!("cannot close a socket that is closing (qd={:?})", qd);
+                        error!("do_close(): {}", &cause);
+                        return Err(Fail::new(libc::ENOTSUP, &cause));
+                    },
                 }
             },
             _ => return Err(Fail::new(libc::EBADF, "bad queue descriptor")),
         };
         inner.addresses.remove(&SocketId::Passive(addr));
```
iyzhang, line 474: Can you add a TODO comment to remove active sockets from the addresses table when the close completes?

tests/rust/tcp.rs
```diff
@@ -459,7 +459,7 @@ fn tcp_bad_listen() {
         Err(e) if e.errno == libc::EINVAL => (),
         _ => panic!("invalid call to listen() should fail with EINVAL"),
     };
-    safe_close_passive(&mut libos, sockqd);
+    safe_close_active(&mut libos, sockqd);
```
iyzhang, line 462: Inside the safe_close_active function, the error still says passive socket.

iyzhang (approved):
LGTM

src/rust/inetstack/protocols/tcp/peer.rs
ppenna: Sure.

tests/rust/tcp.rs
ppenna: Fixed.

PR #533, opened 2023-03-09
[catcollar] Implement async_close

iyzhang:
This PR implements async_close for catcollar. It is almost identical in code to catnap.

### DIFF
--- a/src/rust/catcollar/futures/close.rs
+++ b/src/rust/catcollar/futures/close.rs
@@ -1,0 +1,86 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+//==============================================================================
+// Imports
+//==============================================================================
+
+use crate::runtime::{
+    fail::Fail,
+    QDesc,
+};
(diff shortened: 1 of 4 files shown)


src/rust/catcollar/futures/close.rs
```diff
@@ -0,0 +1,86 @@
...
+                let errno: libc::c_int = unsafe { *libc::__errno_location() };
+
+                // Operation was interrupted, retry?
+                if errno == libc::EINTR {
+                    ctx.waker().wake_by_ref();
+                    return Poll::Pending;
+                }
+                // Operation failed.
+                else {
+                    let message: String = format!("close(): operation failed (errno={:?})", errno);
```
ppenna, line 79: Could you rename `message` to `cause`?

ppenna (approved):

src/rust/catcollar/futures/close.rs
iyzhang: yup

PR #535, opened 2023-03-10
[runtime] Add ability to set result in FutureResult

iyzhang:
Currently, the only way for a FutureResult to complete is for the Future to complete with a ready and a result. This PR adds the ability to externally trigger the completion of the FutureResult (e.g., because the socket was closed). The code that sets the result should also wake the Future to ensure that the scheduler runs it on the next iteration and returns the set result the next time that the application waits on the qtoken.


iyzhang:
Moving to a new task abstraction

PR #538, opened 2023-03-10
[catnap] Move to new Demikernel scheduler

iyzhang:
This PR moves catnap to the new Demikernel scheduler. Basically the only changes required are to instantiate the other scheduler and put a Box around the coroutines so we can eventually dynamically dispatch over multiple libOSes.


iyzhang:
Moving in a different direction

PR #540, opened 2023-03-10
[catnip, catpowder] Move to newly renamed scheduler

iyzhang:
This PR moves the inetstack, catnip and catpowder to our newly renamed scheduler. It requires us to box all of our futures for dynamic dispatch later.


iyzhang:
Moving to a different direction

Issue #542, opened 2023-03-14
[runtime] Improve Split Functionality in `DemiBuffer`


ppenna:
## Context

The `DemiBuffer::split_off()` function splits a buffer into two and returns the caller the half back of the split buffer.

This functionality enables a partial number of bytes of a `DemiBuffer` to be consumed, and it finds application when we are processing receive queues in our network stack (although we are not currently using it).

Unfortunately in some use-cases, the caller may be interested in consuming the front half of the split buffer instead, thus `DemiBuffer::split_off()` is not enough.

https://github.com/demikernel/demikernel/blob/e8825ec1a000894b7a9e86c574a20da06ab2215d/src/rust/runtime/memory/demibuffer.rs#L431-L475

## Proposed Solution

- Rename `DemiBuffer::split_off()` to `DemiBuffer::split_back()` so that we make it explicit that we are splitting a buffer and returning the back half.
- Introduce `DemiBuffer:split_front()`, which splits the target `DemiBuffer` and returns the front half to be consumed.

## Alternative Solutions

We could multiplex the split front and back functionalities in the same function.

PR #546, opened 2023-03-14
[runtime] Improve Split Functionality in `DemiBuffer`

ppenna:
## Description

This PR closes https://github.com/demikernel/demikernel/issues/542

## Summary of Changes

- Renamed `DemiBuffer::split_off()` to `DemiBuffer::split_back()`
- Improved documentation for `DemiBuffer::split_back()`
- Introduced `DemiBuffer::split_front()`
- Introduced a test unit for `DemiBuffer::split_front()`

### DIFF
--- a/src/rust/runtime/memory/demibuffer.rs
+++ b/src/rust/runtime/memory/demibuffer.rs
@@ -1014,7 +1014,7 @@ mod tests {
     #[test]
     fn split_back() {
         // Create a new (heap-allocated) `DemiBuffer` by copying a slice of a `String`.
-        let mut str: String = String::from("word one two three four five six seven eight nine");
+        let str: String = String::from("word one two three four five six seven eight nine");
         let slice: &[u8] = str.as_bytes();
         let result: Result<DemiBuffer, Fail> = DemiBuffer::from_slice(slice);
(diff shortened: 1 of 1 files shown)


src/rust/runtime/memory/demibuffer.rs
```diff
@@ -428,52 +428,106 @@ impl DemiBuffer {
...
+    ///
+    /// On successful completion, a new [DemiBuffer] containing the data before the split point is returned. On failure,
+    /// a [Fail] structure encoding the failure condition is returned instead.
+    ///
+    /// **Notes**
+    ///
+    /// - The target [DemiBuffer] must be a single buffer segment (not a chain).
+    /// - The target [DemiBuffer] should be large enough to hold `offset`.
+    ///
+    pub fn split_front(&mut self, offset: usize) -> Result<Self, Fail> {
```
iyzhang, line 499: To eliminate duplicate code, we should probably have a shared function here that gives both halves back and then split_front and split_back just return the front_half or back_half.

iyzhang (approved):
Looks good to me other than the duplicated code.

src/rust/runtime/memory/demibuffer.rs
ppenna: Sure... I merged the two functions in one, called `split()`

PR #548, opened 2023-03-14
[ci] Enable `tcp-accept` in Regressions

ppenna:
## Description

This PR enables `tcp-accept` in regressions.


ppenna:
Merging this PR because it concerns the regression system

PR #553, opened 2023-03-15
[ci]  Support Test Alias

ppenna:
## Description

This PR address support for test alias, so that future tests that support multiple run modes may be correctly reported.


ppenna:
Merging this PR because it concerns the regression system.

PR #556, opened 2023-03-19
[runtime] Add cooperative yielding to the scheduler

iyzhang:
This PR adds calls for cooperatively yielding an async function/coroutine to the scheduler. It wraps a future that either yields until woken (yield_until_wake) or yields one time quanta and immediately reschedules itself to run again (yield_once). We should aim to only use yield_until_wake for the best efficiency but yield_once lets us perform polling in async functions.


src/rust/scheduler/coop_yield.rs
```diff
@@ -0,0 +1,89 @@
...
+        self.runnable
+    }
+
+    pub fn set_runnable(&mut self) {
+        self.runnable = true;
+    }
+
+    pub fn yield_once(&self) -> bool {
+        self.once
+    }
```
ppenna, line 51: Could you add an API documentation to these functions?

src/rust/scheduler/coop_yield.rs
```diff
@@ -0,0 +1,89 @@
...
+            return Poll::Ready(());
+        }
+
+        self_.set_runnable();
+        // If we are only yielding for one scheduler tick, then schedule immediately. Otherwise, wait for another
+        // coroutine to send a signal via the waker.
+        if self_.yield_once() {
+            context.waker().wake_by_ref();
+        }
+        Poll::Pending
```
ppenna, line 75: Could you add a new line before this one?

src/rust/scheduler/coop_yield.rs
```diff
@@ -0,0 +1,89 @@
...
+
+    /// Polls the underlying accept operation.
+    fn poll(self: Pin<&mut Self>, context: &mut Context) -> Poll<Self::Output> {
+        let self_: &mut Yield = self.get_mut();
+
+        if self_.is_runnable() {
+            return Poll::Ready(());
+        }
+
+        self_.set_runnable();
```
ppenna, line 69: Could you add a new line after this one?

src/rust/scheduler/coop_yield.rs
```diff
@@ -0,0 +1,89 @@
...
+// Standalone Functions
+//==============================================================================
+
+pub async fn yield_once() {
+    Yield::new(true).await
+}
+
+pub async fn yield_until_wake() {
+    Yield::new(false).await
+}
```
ppenna, line 89: Could we make this private associated functions of `Yield`?

iyzhang: These can't be private because these are the ones that the libOSes called. I can make the Yield associated function private though.

ppenna (approved):
Great, thanks!

Issue #536, opened 2023-03-10
[catnip] First DPDK eal_init Argument is Skipped


ihchoi12:
# Context
In the default configuration file (scripts/config/default.yaml), we put the dpdk eal initialization arguments (`eal_init`) starting from the first slot. However, `eal_init(int argc, char **argv)` function explicitly skips the first argument `argv[0]` and starts processing the arguments from `argv[1]`. So, currently the first argument is not applied in the configuration. 

# Proposed Solution
We can simply update the the default configuration file by adding a empty string at the beginning of `eal_init`.

PR #558, opened 2023-03-20
[catnip] Append an empty string at eal_init args to be skipped

ihchoi12:
# Description
This PR solves https://github.com/demikernel/demikernel/issues/536

# Summary of Changes
Just append an empty string to `eal_init` in the `default.yaml` configuration file.


iyzhang:
Thanks for doing this! Can you do a git rebase against dev and squash your commits into one commit?

ihchoi12:
Done. Thanks for the suggestion!

carvalhof:
Another possibility is to insert a string (_e.g.,_ "") at the first position of `eal_init_refs` vector in the `initialize_dpdk` function.  Branch `bugfix-dpdk-eal` presents this possibility.

ppenna (approved):

PR #568, opened 2023-03-23
[catnip] Bad Assumption on Packet Header Size

### LINKED ISSUE #85: [catnip] Bad Assumption on Packet Header Size
Description
========

When creating memory pool for holding packet headers, the memory manager is assuming a wrong size for these structures:

```
let header_size: usize = ETHERNET2_HEADER_SIZE + IPV4_HEADER_DEFAULT_SIZE + MAX_TCP_HEADER_SIZE;
```

This should be at least large enough to hold the maximum possible header for a TCP packet.

carvalhof:
### Description

- This PR closes #85.

### Summary of Changes
- Introduced the `IPV4_HEADER_MAX_SIZE` which is 60 bytes;
- Considered the maximum size for headers as `ETHERNET2_HEADER_SIZE + IPV4_HEADER_MAX_SIZE + MAX_TCP_HEADER_SIZE`;
- Fixed the `compute_size` function, calculating the IPv4 header size according to the `ihl` field.

### Observation
- This PR does not consider IEEE 802.1Q (Simple VLAN) nor 802.1ad (Double VLAN) where `ETHERNET2_HEADER_SIZE` could be greater than 14 bytes (as defined in [src/rust/inetstack/protocols/ethernet2/frame.rs](https://github.com/demikernel/demikernel/blob/dev/src/rust/inetstack/protocols/ethernet2/frame.rs#L18)).

### DIFF
--- a/src/rust/catnip/runtime/memory/manager.rs
+++ b/src/rust/catnip/runtime/memory/manager.rs
@@ -9,7 +9,7 @@ use super::mempool::MemoryPool;
 use crate::{
     inetstack::protocols::{
         ethernet2::ETHERNET2_HEADER_SIZE,
-        ipv4::IPV4_HEADER_DEFAULT_SIZE,
+        ipv4::IPV4_HEADER_MAX_SIZE,
         tcp::MAX_TCP_HEADER_SIZE,
     },
     runtime::{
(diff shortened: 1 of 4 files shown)


ppenna (commented):
Looks good now.

Could you just squash commits into a single one or reword the title of the last commit so that it matches our coding standards? Also, could you rebase the branch with latest dev?

PR #562, opened 2023-03-22
[test] Integration Tests for TCP

ppenna:
## Description

- This PR partially addresses https://github.com/demikernel/demikernel/issues/554
- This PR partially addresses https://github.com/demikernel/demikernel/issues/549
- This PR partially addresses https://github.com/demikernel/demikernel/issues/526
- This PR partially addresses https://github.com/demikernel/demikernel/issues/501
- This PR partially addresses https://github.com/demikernel/demikernel/issues/567
- This PR partially addresses https://github.com/demikernel/demikernel/issues/566

## Summary of Changes

Added the following integration tests for TCP `accept()`:

- Attempt to accept connections on an invalid queue descriptor
- Attempt to accept connections on a TCP socket that is not bound
- Attempt to accept connections on a TCP socket that is not listening
- Attempt to accept connections on a TCP socket that is listening
- Attempt to accept connections on a TCP socket that is connecting
- Attempt to accept connections on a TCP socket that is already accepting connections

Added the following integration tests for TCP `bind()`:

- Attempt to bind an address to an invalid queue descriptor
- Attempt to bind multiple addresses to the same socket
- Attempt to bind the same address to two sockets
- Attempt to bind to all private ports

Added the following integration tests for TCP `close()`:

- Attempt to close an invalid queue descriptor
- Attempt to close a TCP socket multiple times
- Attempt to close a TCP socket that is not bound
- Attempt to close a TCP socket that is bound
- Attempt to close a TCP socket that is listening
- Attempt to close a TCP socket that is accepting
- Attempt to close a TCP socket that is connecting

Added the following integration tests for TCP `connect()`:

- Attempt to connect an invalid queue descriptor
- Attempt to connect a TCP socket that is not bound
- Attempt to connect a TCP socket that is listening
- Attempt to connect a TCP socket that is already connecting
- Attempt to connect a TCP socket that is accepting connections

Added the following integration tests for TCP `listen()`:

- Attempt to listen for connections on an invalid queue descriptor
- Attempt to listen for connections on a TCP socket that is not bound
- Attempt to listen for connections on a TCP socket that is already listening for connections
- Attempt to listen for connections on a TCP socket that is bound
- Attempt to listen for connections on a TCP socket that is connecting
- Attempt to listen for connections on a TCP socket that is accepting connections

Added the following integration tests for TCP `socket()`:

- Attempt to create a TCP socket using an unsupported domain
- Attempt to create a TCP socket using an unsupported socket type

PR #569, opened 2023-03-23
[catnap] Several Fixes Around Socket State

ppenna:
## Description

- This PR fixes several bugs in Catnap LibOS that were exposed by https://github.com/demikernel/demikernel/pull/562

## Summary of Changes

- Introduced `Socket` abstraction
  - It stores the state of a socket, as well as the local address to which a socket is bound and the remote address to which a socket is connected.
  - It encodes error conditions when attempting to perform invalid transitions in the lifetime of a socket.
- Changed the `CatmemQueue` abstraction to use the `Socket` abstraction.
- Changed the PDPIX layer for `CatmemQueue` to use the `Socket` abstraction.

### DIFF
--- a/src/rust/catnap/mod.rs
+++ b/src/rust/catnap/mod.rs
@@ -181,12 +181,20 @@ impl CatnapLibOS {
         }
 
         // Get a mutable reference to the queue table.
-        // It is safe to unwrap because we checked before that the queue descriptor is valid.
-        let queue: &mut CatnapQueue = qtable.get_mut(&qd).expect("queue descriptor should be in queue table");
+        // The following call to expect() is safe because we checked before that the queue descriptor is valid.
+        let queue: &mut CatnapQueue = qtable
+            .get_mut(&qd)
+            .expect("queue descriptor should be in the queue table");
 
         // Get reference to the underlying file descriptor.
-        // It is safe to unwrap because when creating a queue we assigned it a valid file descritor.
+        // The following call to expect() is safe because when creating a queue we assigned it a valid file descritor.
         let fd: RawFd = queue.get_fd().expect("queue should have a file descriptor");
+
+        // Create a socket that is bound to the target local address.
+        let bound_socket: Socket = {
+            let unbound_socket: &Socket = queue.get_socket();
+            unbound_socket.bind(local)?
+        };
 
         // Bind underlying socket.
         let sockaddr: libc::sockaddr_in = linux::socketaddrv4_to_sockaddr_in(&local);
@@ -198,7 +206,8 @@ impl CatnapLibOS {
             )
         } {
             stats if stats == 0 => {
-                queue.set_addr(local);
+                // Update socket.
+                queue.set_socket(&bound_socket);
                 Ok(())
             },
             _ => {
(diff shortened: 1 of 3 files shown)


src/rust/catnap/socket.rs
```diff
@@ -0,0 +1,310 @@
...
+    pub fn new() -> Self {
+        Self {
+            state: SocketState::NotBound,
+            local: None,
+            remote: None,
+        }
+    }
+
+    /// Constructs from [self] a socket that is bound to the `local` address.
+    pub fn bind(&self, local: SocketAddrV4) -> Result<Self, Fail> {
```
iyzhang, line 60: Is there a reason not to use a &mut self rather than returning a new socket?

src/rust/catnap/mod.rs
```diff
@@ -328,14 +413,29 @@ impl CatnapLibOS {
...
+                                // Update socket.
+                                let mut qtable_: RefMut<IoQueueTable<CatnapQueue>> = qtable_ptr.borrow_mut();
+                                match qtable_.get_mut(&qd) {
+                                    Some(queue) => queue.set_socket(&closed_socket),
+                                    None => {
+                                        let cause: &String = &format!("invalid queue descriptor: {:?}", qd);
+                                        error!("{}", &cause);
+                                        return (qd, OperationResult::Failed(Fail::new(libc::EBADF, cause)));
+                                    },
+                                }
```
iyzhang, line 438: I think setting a socket to closed will never be exercised because the IoQueueTable owns the CatnapQueue and once you call free on it, the CatnapQueue will be freed because there are no more references to it.

src/rust/catnap/mod.rs
```diff
@@ -210,14 +223,23 @@ impl CatnapLibOS {
...
+                    };
+
                     if unsafe { libc::listen(fd, backlog as i32) } != 0 {
                         let errno: libc::c_int = unsafe { *libc::__errno_location() };
                         error!("failed to listen ({:?})", errno);
                         return Err(Fail::new(errno, "operation failed"));
                     }
+
+                    // Update socket.
+                    queue.set_socket(&listening_socket);
```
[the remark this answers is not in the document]
ppenna: Yes there is a reason for which I have used this design pattern through my changes.

As noted in the API documentation, in the following code snippet we create a listening socket from a bound socket. Note that this creates a new object, and does not change the socket object that is stored in the queue.

```rust
let listening_socket: Socket = {
    let bound_socket: &Socket = queue.get_socket();
    bound_socket.listen()?
};
```

Then, once we know that the underlying operation has succeeded, we set the socket to the new state:

```rust
queue.set_socket(&listening_socket);
```

src/rust/catnap/mod.rs
```diff
@@ -305,19 +380,29 @@ impl CatnapLibOS {
...
+                    };
+
+                    match unsafe { libc::close(fd) } {
+                        stats if stats == 0 => {
+                            // Update socket.
+                            queue.set_socket(&closed_socket);
+                            qtable.free(&qd);
+                            Ok(())
+                        },
+                        _ => Err(Fail::new(libc::EBADF, "invalid queue descriptor")),
```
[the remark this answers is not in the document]
ppenna: Sure. I will fix it. Note however that this behavior existed in previous code.

src/rust/catnap/mod.rs
ppenna: I cannot assess this function, because it is not yet supported.

src/rust/catnap/mod.rs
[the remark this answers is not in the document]
iyzhang: Ok, sounds good

src/rust/catnap/socket.rs
ppenna: Yes, there is, because we are not changing the state of the target socket, but rather constructing a new one from the target socket.

Issue #565, opened 2023-03-22
[catmem] Check If Queue Is Still Valid When Processing Result


iyzhang:
## Description

We currently expect the queue to still be in the IoQueueTable when we finish the async future. This is not safe to assume because the queue might be removed by a close at any time and we do not currently cancel pending operations when we close queues.

## How to Reproduce

1. Run TCP-echo with 8 or more clients

## Expected Behavior

We should either check after each .await call that all of the references that we hold are still valid or we should cancel all pending Tasks immediately before removing the queue metadata from the IoQueueTable.

## Related Issues

#552

PR #570, opened 2023-03-23
[catmem] Invalid Queue Descriptor in Pop

ppenna:
## Description

This commit fixes https://github.com/demikernel/demikernel/issues/565


iyzhang (approved):
LGTM

PR #573, opened 2023-03-23
[examples] Multi-Client Support for TCP Echo

ppenna:
## Description

This PR introduces multi-client support for our TCO echo example.


examples/tcp-echo/client.rs
```diff
@@ -0,0 +1,305 @@
...
+    fn handle_push(&mut self, qr: &demi_qresult_t) -> Result<()> {
+        let qd: QDesc = qr.qr_qd.into();
+        self.npushed += 1;
+
+        // Pop another packet.
+        self.issue_pop(qd, None)?;
+        Ok(())
+    }
+
+    /// Handles the completion of an expected operation.
```
iyzhang, line 233: This comment should say: unexpected operation

iyzhang (approved):
This looks great. Thanks for putting it together!

examples/tcp-echo/client.rs
ppenna: Thanks.

Issue #578, opened 2023-03-24
[ci] Enable `tcp-echo` in `demikernel-ci`


ppenna:
## Context

We should enable `tcp-echo` to `demikernel_ci`.

This test program should be runnable when the following LibOSes are targeted:

- [ ] Catnap LibOS
- [ ] Catcollar LibOS
- [ ] Catloop LibOS
- [ ] Catpowder LibOS
- [ ] Catnip LibOS

PR #579, opened 2023-03-24
[ci] Enable `tcp-echo` for Catnap LibOS

ppenna:
## Description

- This PR is a partial fix for https://github.com/demikernel/demikernel/issues/578

## Summary of changes

- Changed stop condition for TCP Echo server.
- Improved logs on TCP Echo application to use stdout and not conflict with stderr
- Enabled `tcp-echo` for Catap LibOS in `demikernel-ci`


iyzhang (approved):
LGTM

Issue #580, opened 2023-03-27
[example] Concurrent Mode for `tcp-echo`


ppenna:
## Context

Currently, the `tcp-echo` application exercises a "sequential connection setup" use case scenario:
- The TCP Echo client issues several `connect()` operations
- The TCP Echo client waits for all connections to be established
- The TCP echo client and server exchange data

## Proposed Solution

The `tcp-echo` application should also exercise a "concurrent connection setup" use case scenario:
- The TCP Echo client issues several `connect()` operation.
- As as the TCP Echo establishes a connection, it starts exchanging data with the TCP server (even if there are further connections to be established)

PR #581, opened 2023-03-27
[examples] Concurrent Mode for `tcp-echo`

ppenna:
## Description

This PR closes https://github.com/demikernel/demikernel/issues/580


iyzhang (approved):
LGTM

Issue #554, opened 2023-03-15
[test] Integration Tests for TCP `connect()`


ppenna:
## Context

We should have the following integration tests for TCP `connect()`:

- [x] Attempts to connect an invalid socket.
- [ ] Attempt to connect a TCP socket to a remote that is not accepting connections.
- [x] Attempt to connect a TCP socket that is not bound.
- [ ] Attempt to connect a TCP socket that is bound.
- [x] Attempt to connect a TCP socket that is listening.
- [x] Attempt to connect a TCP socket that is connecting.
- [x] Attempt to connect a TCP socket that is accepting connections.
- [ ] Attempt to connect a TCP socket that is closed.

PR #587, opened 2023-03-27
[test] Integration Tests for TCP `connect()`

ppenna:
## Description

This PR closes https://github.com/demikernel/demikernel/issues/554

## Summary of Changes

Added the following integration tests for TCP `connect()`:
- Attempt to connect a TCP socket to a remote that is not accepting connections.
- Attempt to connect a TCP socket that is bound.
- Attempt to connect a TCP socket that is closed.

### DIFF
--- a/tests/rust/tcp-test/connect/mod.rs
+++ b/tests/rust/tcp-test/connect/mod.rs
@@ -44,6 +44,7 @@ pub fn run(libos: &mut LibOS, local: &SocketAddrV4, remote: &SocketAddrV4) -> Re
     connect_listening_socket(libos, local, remote)?;
     connect_connecting_socket(libos, remote)?;
     connect_accepting_socket(libos, local, remote)?;
+    connect_closed_socket(libos, remote)?;
 
     Ok(())
 }
@@ -165,6 +166,25 @@ fn connect_accepting_socket(libos: &mut LibOS, local: &SocketAddrV4, remote: &So
 
     // Succeed to close socket.
     libos.close(sockqd)?;
+
+    Ok(())
+}
+
+/// Attempts to connect a TCP socket that is closed.
+fn connect_closed_socket(libos: &mut LibOS, remote: &SocketAddrV4) -> Result<()> {
+    println!("{}", stringify!(connect_closed_socket));
+
+    // Create a closed socket.
+    let sockqd: QDesc = libos.socket(AF_INET, SOCK_STREAM, 0)?;
+    libos.close(sockqd)?;
+
+    // Fail to connect().
+    let e: Fail = libos
+        .connect(sockqd, remote.to_owned())
+        .expect_err("connect() a closed socket should fail");
+
+    // Sanity check error code.
+    assert_eq!(e.errno, libc::EBADF, "connect() failed with {}", e.cause);
 
     Ok(())
 }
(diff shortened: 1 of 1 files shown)


tests/rust/tcp-test/connect/mod.rs
```diff
@@ -86,6 +92,59 @@ fn connect_unbound_socket(libos: &mut LibOS, remote: &SocketAddrV4) -> Result<()
...
+    let remote: SocketAddrV4 = {
+        let ipv4: Ipv4Addr = Ipv4Addr::UNSPECIFIED;
+        SocketAddrV4::new(ipv4, 0)
+    };
+
+    // Succeed to connect socket.
+    let qt: QToken = libos.connect(sockqd, remote)?;
+
+    // Poll for enough time to get the connection refused.
+    match libos.wait(qt, Some(Duration::from_secs(75))) {
```
iyzhang, line 112: Did you put a timeout on this just to keep the test from hanging? If so, we may want to check explicitly for time out for debugging to see why that happened.

tests/rust/tcp-test/connect/mod.rs
```diff
@@ -168,3 +227,22 @@ fn connect_accepting_socket(libos: &mut LibOS, local: &SocketAddrV4, remote: &So
 
     Ok(())
 }
+
+/// Attempts to connect a TCP socket that is closed.
+fn connect_closed_socket(libos: &mut LibOS, remote: &SocketAddrV4) -> Result<()> {
```
iyzhang, line 232: Not sure if we should add tests for async_close or just eventually remove close(). Close needs to eventually be an async because it is blocking in general and we don't want to block execution to wait for close to complete.

iyzhang (approved):
Some small comments but nothing big

tests/rust/tcp-test/connect/mod.rs
ppenna: This test is about attempting to connect to a TCP socket to a remote that is NOT accepting connections. The only way to be sure that the remote is NOT accepting connections is to wait for 75 seconds and get a `ECONNREFUSED` error. If we get `Err(_)` the test should not pass. Are you suggesting to print the error value?

tests/rust/tcp-test/connect/mod.rs
ppenna: I agree that once we have `async_close` implemented, `close` should be dropped alogside all its tests.

tests/rust/tcp-test/connect/mod.rs
iyzhang: We should probably get a response in less than 75 seconds. If we get a ETIMEOUT after 75 seconds, something else is wrong, so I'm saying that we should check specifically for timeout vs other errors.

ppenna: Are you suggesting something like this?

```rust
    // Poll for enough time to get the connection refused.
    match libos.wait(qt, Some(Duration::from_secs(75))) {
        Ok(qr) if qr.qr_ret == libc::ECONNREFUSED => {},
        Ok(_) => anyhow::bail!("wait() should return ECONNREFUSED"),
        Err(e) if e.errno == libc::ETIMEDOUT => anyhow::bail!("wait() should not timeout"),
        Err(_) => anyhow::bail!("wait() should not fail"),
    }
```

iyzhang: Yes!

ppenna: Sure.

Issue #590, opened 2023-03-28
[catmem] Spawn Failed Co-Routine for `ECONNRESET` in `pop()`


ppenna:
## Description

In Catmem LibOS, when end of file is reached and we call `pop()`, we synchronously fail.

## Code Snippet

https://github.com/demikernel/demikernel/blob/fa7ef1ba67ba07b44943217e09ee8c1fd91b43d2/src/rust/catmem/mod.rs#L217-L270

## Expected Behavior

We should spawn a co-routine that encodes the failed operation with a `ECONNRESET` error code.

PR #595, opened 2023-04-03
[catmem] Connection Reset in Push and Pop

ppenna:
## Description

- This PR closes https://github.com/demikernel/demikernel/issues/590

## Summary of Changes

- Fixed `Catmem::pop()` to spawn a failed co-routine when the connection is reset (we were returned an error before)
- Fixed `Catmem::push()` to spawn a failed co-routine when the connection is reset (we were returned an error before)
- Improved some error messages in Catmem to print information about the failure context.
- Fixed `pipe-open` program to handle connection reset.

### DIFF
--- a/examples/pipe-open/server.rs
+++ b/examples/pipe-open/server.rs
@@ -59,6 +59,7 @@ impl PipeServer {
         let qr: demi_qresult_t = self.libos.wait(qt, None)?;
         let sga: demi_sgarray_t = match qr.qr_opcode {
             demi_opcode_t::DEMI_OPC_POP => unsafe { qr.qr_value.sga },
+            demi_opcode_t::DEMI_OPC_FAILED => return Ok(self.handle_fail(&qr)?),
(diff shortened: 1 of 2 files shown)


iyzhang (approved):
This change looks good to me but is there a reason to asynchronously return the failure instead of just returning it right away and not allocating a qtoken at all?

ppenna:
> This change looks good to me but is there a reason to asynchronously return the failure instead of just returning it right away and not allocating a qtoken at all?

@iyzhang yes, there is a reason: this is not the behavior that we have for Catnap LibOS (which is validated and running in integration tests). All LibOSes must have the same behavior.

iyzhang:
> > This change looks good to me but is there a reason to asynchronously return the failure instead of just returning it right away and not allocating a qtoken at all?
> 
> @iyzhang yes, there is a reason: this is not the behavior that we have for Catnap LibOS (which is validated and running in integration tests). All LibOSes must have the same behavior.

Ah makes sense. Looks good to me then.

Issue #172, opened 2022-06-23
[build] Drop Use of Unstable Features


ppenna:
Description
------------

We currently rely on the following unstable features:

- [ ] [`test`](https://doc.rust-lang.org/unstable-book/library-features/test.html): mainly used for feature microbenchmarks
- [ ] [`allocator_api`](https://doc.rust-lang.org/unstable-book/library-features/allocator-api.html): used for implementing `WakerPageRef`

Expected Behavior
------------------

We should not rely on unstable features at all.

BrianZill:
Pedro, where does this list come from?  As I write this, there are 10 more optional "features" listed in lib.rs, is there an automatic way to determine if they are unstable or not?  Many of these others are listed in [The Unstable Book](https://doc.rust-lang.org/unstable-book/the-unstable-book.html).  Does that mean that they should be listed here as well?

ppenna:
@BrianZill thanks for checking this.

This list comes from the features that we list in the `lib.rs`, thus yes you are correct we are missing some here. This issues was transferred from  the old scheduler repository, and after that I have not updated it. I'm opening now a PR which gets rid of unused features, as well as I am updating the description of the issue here.

ppenna:
Assigning all engineers to this issue, because it is critical.

PR #596, opened 2023-04-03
[build] Drop Use of `new_uninit`

ppenna:
## Description

- This PR is a partial fix for https://github.com/demikernel/demikernel/issues/172

## Summary of Changes

- Drop use of `new_uninit` unstable feature


iyzhang (approved):
LGTM

PR #614, opened 2023-04-05
[doc] Reword TODOs in Source Tree

ppenna:
## Description

This PR rewords TODOs in our source tree, so that they may all be trackable by the "Todo Tree" plugging in VS Code.


ppenna:
Merging this PR because it does not introduce any changes to the source code.

Issue #622, opened 2023-04-06
[catloop] Collisions on Queue Token Space


ppenna:
## Description

We are naively distributing queue tokens to user applications, regardless if they were allocated by Catloop LibOS (for `accept()` and `connect()` or Catmem LibOS (for `push()` and `pop()`). With this naïve scheme, queue tokens may collide and thus we may misbehave.

For instance, if `push()` returns `qt = 1` and `connect()` returns `qt = 1`, when polling `qt = 1` Catloop cannot distinguish whether that queue token refers to `push()` or `connect()`.

## Code Snippets

The following code snippet shows that the returned queue token from `connect()` is the one immediately returned by Catloop's scheduler:

https://github.com/demikernel/demikernel/blob/b596f4f115accc8895d6d276c368b59c59948400/src/rust/catloop/mod.rs#L344-L355

The following code snippet shows that the returned queue token from `push()` is the one immediately returned by Catmem's scheduler:

https://github.com/demikernel/demikernel/blob/b596f4f115accc8895d6d276c368b59c59948400/src/rust/catloop/mod.rs#L409-L412

Note that this also happens in `accept()` and `pop()`.

PR #623, opened 2023-04-06
[catloop] Collisions on Queue Token Space

ppenna:
## Description

This PR closes https://github.com/demikernel/demikernel/issues/622

## Summary of Changes

- Improved some error logs in Catloop LibOS
- Fixed collisions in queue token space of Catloop LibOS
   - Queue tokens from Catmem LibOS are now registered in a different table

### DIFF
--- a/src/rust/catloop/mod.rs
+++ b/src/rust/catloop/mod.rs
@@ -125,8 +125,9 @@ impl CatloopLibOS {
 
         // Parse communication domain.
         if domain != libc::AF_INET {
-            error!("communication domain not supported (domain={:?})", domain);
-            return Err(Fail::new(libc::ENOTSUP, "communication domain not supported"));
+            let cause: String = format!("communication domain not supported (domain={:?})", domain);
+            error!("socket(): {}", cause);
+            return Err(Fail::new(libc::ENOTSUP, &cause));
         }
 
         // Parse socket type and protocol.
@@ -134,8 +135,9 @@ impl CatloopLibOS {
             libc::SOCK_STREAM => QType::TcpSocket,
             libc::SOCK_DGRAM => QType::UdpSocket,
             _ => {
-                error!("socket type not supported (typ={:?})", typ);
-                return Err(Fail::new(libc::ENOTSUP, "socket type not supported"));
+                let cause: String = format!("socket type not supported (typ={:?})", typ);
+                error!("socket(): {}", cause);
+                return Err(Fail::new(libc::ENOTSUP, &cause));
             },
         };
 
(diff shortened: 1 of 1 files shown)


src/rust/catloop/mod.rs
```diff
@@ -428,13 +446,17 @@ impl CatloopLibOS {
...
+                return Err(Fail::new(libc::EBADF, &cause));
+            },
         };
 
         let qt: QToken = self.catmem.borrow_mut().pop(catmem_qd, size)?;
-        self.qts.insert(qt, (demi_opcode_t::DEMI_OPC_POP, qd));
+        self.catmem_qts.insert(qt, (demi_opcode_t::DEMI_OPC_POP, qd));
 
-        Ok(qt)
+        Ok(Self::shift_qtoken(qt))
```
iyzhang, line 459: Not super important, but you could probably get away with a single HashMap if you use the shifted qtoken.

src/rust/catloop/mod.rs
```diff
@@ -447,42 +469,97 @@ impl CatloopLibOS {
...
+        if let Some((ref opcode, _)) = self.catloop_qts.get(&qt) {
+            // Check if the queue token concerns an expected operation.
+            if opcode != &demi_opcode_t::DEMI_OPC_ACCEPT && opcode != &demi_opcode_t::DEMI_OPC_CONNECT {
+                let cause: String = format!("unexpected queue token (qt={:?})", qt);
+                error!("schedule(): {:?}", &cause);
+                return Err(Fail::new(libc::EINVAL, &cause));
+            }
+
+            // Insert the queue token into the scheduler.
+            match self.scheduler.from_raw_handle(qt.into()) {
```
iyzhang, line 484: This code doesn't actually insert the token into the scheduler but resolves it from a queue token to a scheduler handle, correct? The actual insert happens when scheduler.insert() is called in the accept/connect/push/pop.

iyzhang (approved):
This PR looks good to me, but we should definitely do something with this code when we unify the scheduler to handle multiple libOSes. Should we file an issue?

src/rust/catloop/mod.rs
ppenna: Are you suggesting that we should replace this by a hashmap? We already have one, it is important to actually distribute unique queue tokens to applications.

src/rust/catloop/mod.rs
ppenna: You are right. Actually the function `schedule()` itself should be renamed to something more meaningful. I'll fix the comment in this PR.

src/rust/catloop/mod.rs
iyzhang: No, I'm saying if you use the shifted value, you can keep both sets of qtokens in the same hash map because the range no longer overlaps. Again, this is not super important.

ppenna: Oh, cool idea! I will work on this.

ppenna: I'll take this optimization in a different PR. I'm submitting an issue for this.

Issue #626, opened 2023-04-06
[test] Remove assumptions about when coroutines run from integration tests


iyzhang:
## Description

Currently our integration tests make two assumptions about wait: 
1. Demikernel (and its coroutines) only run in wait()
2. wait() only runs coroutines related to the queue tokens given to wait.

These assumptions do not always hold, so the test should not expect it. For example, at any time, we may choose to poll a coroutine inline with the initial asynchronous call for performance, especially with push, so the application shouldn't make any assumptions about when the coroutine might complete.

## How to Reproduce

1. Make an I/O call.
2. Call wait(). 
3. The test fails because it expects ETIMEOUT and the coroutine has actually already completed. 

## Expected Behavior

The test should check for either ETIMEOUT or an expected result from the completed operation.

PR #627, opened 2023-04-07
[test] Removing Scheduler Assumptions from Tests

iyzhang:
This PR closes issue #626. I removed 2 tests from the close and async_close tests because they are replicated by the wait after close tests. For example, close_accepting_socket and close_connecting_socket is now identical to wait_after_close_accepting_socket. There are additional tests that are duplicated, for example, accept_listening_socket is identical to wait_after_closing_accepting socket. Likewise, close_listening_socket is identical to listen_bound_socket. I chose not to remove those duplicates as they are in different modules and potentially serve different purposes, but I removed the duplicate tests that were all for close. Let me know if we want to do something else.

### DIFF
--- a/tests/rust/tcp-test/close/wait.rs
+++ b/tests/rust/tcp-test/close/wait.rs
@@ -121,11 +121,11 @@ fn wait_after_close_connecting_socket(libos: &mut LibOS, remote: &SocketAddrV4) 
             Ok(qr) if qr.qr_opcode == demi_opcode_t::DEMI_OPC_CONNECT && qr.qr_ret == 0 => {
                 anyhow::bail!("connect() should not succeed because remote does not exist")
             },
-
             Ok(_) => anyhow::bail!("wait() should return an error on connect() after close()"),
             Err(_) => anyhow::bail!("wait() should not time out"),
         }
     }
+
     Ok(())
 }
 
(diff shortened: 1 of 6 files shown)


tests/rust/tcp-test/close/wait.rs
ppenna, line 1: Could you revert changes in this file?

tests/rust/tcp-test/listen/mod.rs
```diff
@@ -180,15 +180,24 @@ fn listen_connecting_socket(libos: &mut LibOS, remote: &SocketAddrV4) -> Result<
...
+        // If completes successfully, something has gone wrong.
+        Ok(qr) if qr.qr_opcode == demi_opcode_t::DEMI_OPC_CONNECT && qr.qr_ret == 0 => {
+            anyhow::bail!("connect() should not succeed because remote does not exist")
+        },
         Ok(_) => anyhow::bail!("wait() should not succeed"),
         Err(_) => anyhow::bail!("wait() should timeout"),
     }
 
     // Fail to listen().
+    // TODO: Not sure if we should be able to listen after a failed connect().
```
ppenna, line 200: Yes, we should.

ppenna (commented):
Overall looks good. Could you work on the minor changes that I flagged?

ppenna (approved):
Looks good! I just left some minor comments.

tests/rust/tcp-test/close/wait.rs
iyzhang: I just wanted to take the opportunity to fix the formatting in this file.

PR #628, opened 2023-04-07
[test] Add sanity check to payload order

### LINKED ISSUE #394: [test] Sanity Check Payload Order in `tcp-ping-pong`
## Context

The `tcp-ping-pong` testing program performs a naïve sanity check on the data that is passed around. On client and server side, it checks:
- If the received scatter-gather array has the expected length.
- If the received the scatter-gather is filled in with an expected magic byte.

While this sanity check ensures that the message has the expected length, it does not guarantee that the contents of the received scatter-gather array is correct.

For instance, the order of bytes could have been messed up, and since we use the same magic byte to fill up the scatter-gather array, we are not able to identify this error.

## Proposed Solution

We should use  some bits of the magic byte to encode the order of that byte in the payload. Then, when receiving a scatter-gather array, we should check if the order of bytes also match what we are expecting.

Note that we do not have enough bits available in the magic byte to encode the order of a byte in the payload. Thus we may want to run some ECC with the current data-transfer round.

## Related Issues

- https://github.com/demikernel/demikernel/issues/390
- https://github.com/demikernel/demikernel/pull/391 (introduces round check)

iyzhang:
This PR closes #394. Currently we fill the packet with a single value. This change updates the fill to start with the iteration of the test and then fills the remainder of the packet with numerically increasing values.

### DIFF
--- a/examples/rust/tcp-ping-pong.rs
+++ b/examples/rust/tcp-ping-pong.rs
@@ -57,13 +57,17 @@ fn mksga(libos: &mut LibOS, size: usize, value: u8) -> demi_sgarray_t {
     };
 
     // Ensure that scatter-gather array has the requested size.
-    assert!(sga.sga_segs[0].sgaseg_len as usize == size);
+    assert_eq!(sga.sga_segs[0].sgaseg_len as usize, size);
 
     // Fill in scatter-gather array.
     let ptr: *mut u8 = sga.sga_segs[0].sgaseg_buf as *mut u8;
     let len: usize = sga.sga_segs[0].sgaseg_len as usize;
     let slice: &mut [u8] = unsafe { slice::from_raw_parts_mut(ptr, len) };
-    slice.fill(value);
+    let mut fill: u8 = value;
+    for x in slice {
+        *x = fill;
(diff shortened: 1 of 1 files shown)


examples/rust/tcp-ping-pong.rs
```diff
@@ -261,12 +266,15 @@ fn client(remote: SocketAddrV4) -> Result<()> {
...
 
+        let mut fill_check: u8 = (i % (u8::MAX as usize - 1) + 1) as u8;
+
         // Pop data, and sanity check it.
         {
             let mut recvbuf: [u8; BUFFER_SIZE] = [0; BUFFER_SIZE];
             pop_and_wait(&mut libos, sockqd, &mut recvbuf);
             for x in &recvbuf {
-                assert!(*x == fill_char);
+                assert!(*x == fill_check);
```
ppenna, line 276: Could you change this to `assert_eq()`, this would print the values if it fails?

ppenna (approved):
It looks good. Could you just fix the change that I flagged?

examples/rust/tcp-ping-pong.rs
iyzhang: Ah perfect

Issue #631, opened 2023-04-10
[ci] GitHub Actions Fails to Upload Artifacts


ppenna:
## Description

GitHub actions fails to upload artifacts because log names changed from `*.stdout` and `*.stderr` to `*.stdout.txt` and `*.stderr.txt`.

## Related Resources

- This issues was introduced by https://github.com/demikernel/demikernel/pull/629

PR #632, opened 2023-04-10
[ci] GitHub Actions Fails to Upload

ppenna:
## Description

This commit closes https://github.com/demikernel/demikernel/issues/631


ppenna:
Merging this PR because it concerns our CI system.

Issue #589, opened 2023-03-28
[catmem] Increase Size of Shared Ring Buffer to Mitigate Contention


ppenna:
## Context

Catmem LibOS is built on top of a lock-free, single consumer, single producer, ring buffer.

Inherently from this design, an operation may eventually not succeed, either because the ring buffer is full and we want to push more data, or it is empty and we want to pop some data.

The later situation happens due to a slow producer and it is hard to be addressed, but the contention caused by the former situation may be mitigated if the ring buffer is sized appropriately.

## Proposed Solution

Currently, ring buffers in Catmem LibOS [are dimensioned to 4 kb](https://github.com/demikernel/demikernel/blob/fa7ef1ba67ba07b44943217e09ee8c1fd91b43d2/src/rust/catmem/mod.rs#L60), and may effectively store 1 kb of data.

Given that Catmem LibOS is primally used as backend transport for Catloop LibOS, we may set the capacity of the ring buffer so as it is big enough to hold 16 kB of data, and thus match what fits in ethernet jumbo frames.

PR #636, opened 2023-04-10
[catmem] Redimensioning Ring Buffer

ppenna:
## Description

This commit closes https://github.com/demikernel/demikernel/issues/589


iyzhang (approved):
LGTM, I assume that this is because of contention that we need a bigger buffer?

ppenna:
> LGTM, I assume that this is because of contention that we need a bigger buffer?

Yes. This mitigates contention, but does not solve it.

PR #643, opened 2023-04-10
[test] Remove panics from tcp-echo system test

iyzhang:
This PR removes any use of panic! from the tcp-echo system test in examples/tcp-close. Instead, we use bail! to fail gracefully and add tags for where open sockets or scatter-gather arrays should be closed/freed.


ppenna (approved):
Thanks for working on this.

Please reword your commit title so that it follows our code standards.

Issue #615, opened 2023-04-05
[test] Integration Tests for TCP `create_pipe()`


ppenna:
## Context

We should have the following integration tests for `create_pipe()`:

- [ ] Attempt to create a pipe with an invalid name.
- [ ] Attempt to create two pipes with the same name.

PR #644, opened 2023-04-11
[test] Integration Tests for TCP `create_pipe()`

ppenna:
## Description

This PR closes https://github.com/demikernel/demikernel/issues/615

## Summary of Changes

- Added switch for integration tests in the build system.
  - Run integration tests in TCP queues: `make test-integration-rust TEST_INTEGRATION=tcp-test ...`
  - Run integration tests in pipe queues: `make test-integration-rust TEST_INTEGRATION=pipe-test ...`
- Added the following integration tests for `create_pipe()`:
  - Attempt to create a pipe with an invalid name.
  - Attempt to create two pipes with the same name.

### DIFF
--- a/tests/rust/pipe-test/create_pipe/mod.rs
+++ b/tests/rust/pipe-test/create_pipe/mod.rs
@@ -1,0 +1,44 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+//======================================================================================================================
+// Imports
+//======================================================================================================================
+
+use ::anyhow::Result;
+use ::demikernel::{
+    runtime::fail::Fail,
+    LibOS,
+};
+
+//======================================================================================================================
+// Standalone Functions
+//======================================================================================================================
+
+/// Drives integration tests for pipe queues.
+pub fn run(libos: &mut LibOS, pipe_name: &str) -> Result<()> {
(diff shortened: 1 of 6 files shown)


tests/rust/pipe-test/create_pipe/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    ret
+}
+
+/// Attempts to create a pipe with an invalid name.
+fn create_pipe_with_invalid_name(libos: &mut LibOS) -> Result<()> {
+    // Fail to create pipe with an invalid name.
+    match libos.create_pipe(&format!("")) {
+        Err(e) if e.errno == libc::EINVAL => Ok(()),
+        Ok(_) => Err(anyhow::anyhow!("create_pipe() with invalid name should fail")),
```
iyzhang, line 52: For places like this where anyhow::bail! would be equivalent, we should probably just use anyhow::bail! for conciseness.

iyzhang (approved):
Looks good to me, except for the small comment.

tests/rust/pipe-test/create_pipe/mod.rs
ppenna: Sure, I'll change to match your suggestion.

PR #650, opened 2023-04-12
[test] Remove panic from TCP system tests

iyzhang:
This PR removes panics (and expects and unwraps) from the tcp system tests in examples/rust/tcp-*. It also links a new issue for cleaning up state on errors.


ppenna (approved):
LGTM.

Issue #616, opened 2023-04-05
[test] Integration Tests for TCP `open_pipe()`


ppenna:
## Context

We should have the following integration tests for `create_pipe()`:

- [ ] Attempt to open a pipe with an invalid name.
- [ ] Attempt to open a pipe that does not exist.
- [ ] Attempt to open open the same pipe multiple times.

PR #655, opened 2023-04-13
[test] Integration Tests for TCP `open_pipe()`

ppenna:
## Description

- This PR closes https://github.com/demikernel/demikernel/issues/616

## Summary of Changes

- Added `test!()` helper macro that lunches a test a prints on the standard output if that test has passed or failed.
- Added `update_error!()` helper macro that updates an error variable with an error without overwriting previous error.
- Changed integration tests for `open_pipe()` to use the `test!()` and `update_error!()` helper macros.
- Added the following integration tests for `create_pipe()`:
  - Attempt to open a pipe with an invalid name.
  - Attempt to open a pipe that does not exist.
  - Attempt to open open the same pipe multiple times.

### DIFF
--- a/tests/rust/pipe-test/create_pipe/mod.rs
+++ b/tests/rust/pipe-test/create_pipe/mod.rs
@@ -16,32 +16,13 @@ use ::demikernel::{
 //======================================================================================================================
 
 /// Drives integration tests for pipe queues.
-pub fn run(libos: &mut LibOS, pipe_name: &str) -> Result<()> {
-    let mut ret: Result<(), anyhow::Error> = Ok(());
+pub fn run(libos: &mut LibOS, pipe_name: &str) -> Vec<(String, String, Result<(), anyhow::Error>)> {
+    let mut result: Vec<(String, String, Result<(), anyhow::Error>)> = Vec::new();
 
-    match create_pipe_with_invalid_name(libos) {
-        Ok(_) => println!("[passed] {}", stringify!(create_pipe_with_invalid_name)),
-        Err(e) => {
-            // Don't overwrite previous error.
-            if ret.is_ok() {
-                ret = Err(e);
-            }
-            println!("[FAILED] {}", stringify!(create_pipe_with_invalid_name))
-        },
-    };
+    crate::collect!(result, crate::test!(create_pipe_with_invalid_name(libos)));
+    crate::collect!(result, crate::test!(create_pipe_with_same_name(libos, pipe_name)));
 
-    match create_pipe_with_same_name(libos, pipe_name) {
-        Ok(_) => println!("[passed] {}", stringify!(create_pipe_with_same_name)),
-        Err(e) => {
-            // Don't overwrite previous error.
-            if ret.is_ok() {
-                ret = Err(e);
-            }
-            println!("[FAILED] {}", stringify!(create_pipe_with_same_name))
-        },
-    };
-
-    ret
+    result
 }
 
 /// Attempts to create a pipe with an invalid name.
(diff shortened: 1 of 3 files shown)


tests/rust/pipe-test/create_pipe/mod.rs
```diff
@@ -16,32 +16,13 @@ use ::demikernel::{
...
-        Err(e) => {
-            // Don't overwrite previous error.
-            if ret.is_ok() {
-                ret = Err(e);
-            }
-            println!("[FAILED] {}", stringify!(create_pipe_with_same_name))
-        },
-    };
+    nfailed += crate::test!(create_pipe_with_invalid_name(libos));
+    nfailed += crate::test!(create_pipe_with_same_name(libos, pipe_name));
```
iyzhang, line 23: Rather than just storing the first error and then incrementing a counter on each test, perhaps we should collect all of the errors/results into an Vec? If we don't print them, then it probably doesn't matter whether we keep the first or last one (or even just keep a flag). If we do print them, then we should probably print all of them, regardless of whether we encountered an error before.

tests/rust/pipe-test/open_pipe/mod.rs
```diff
@@ -0,0 +1,35 @@
...
+    nfailed
+}
+
+/// Attempts to open a pipe with an invalid name.
+fn open_pipe_with_invalid_name(libos: &mut LibOS) -> Result<()> {
+    // Fail to create pipe with an invalid name.
+    match libos.create_pipe(&format!("")) {
+        Err(e) if e.errno == libc::EINVAL => Ok(()),
+        Ok(_) => anyhow::bail!("create_pipe() with invalid name should fail"),
+        Err(e) => anyhow::bail!("create_pipe() failed with {}", e.cause),
```
iyzhang, line 35: I started leaving these as just e because I thought it might be nice to print the error code too and because our Fails can print both. You might consider doing that here, rather than throwing away the error code.

iyzhang (approved):
A few small comments. You might want to consider squashing the last 3 commits into a single on that just adds new pipe tests.

tests/rust/pipe-test/open_pipe/mod.rs
ppenna: Sure, I'll change this.

tests/rust/pipe-test/create_pipe/mod.rs
ppenna: I agree that collecting results would be the best solution possible - we are currently printing all failures, the end result of each test, and incrementing the failed counter to produce final report.

When you suggest to collect all errors/results into a vector, do you mean that we should change the return value of each test function to `Result<(), Vector::<anyhow::Error>>`?

ppenna:
> A few small comments. You might want to consider squashing the last 3 commits into a single on that just adds new pipe tests.

I actually did it intentionally, because I like to have three different commits, each of which introducing a new test that is buildable and runnable. Are you OK with me keeping it? That helps bisecting.

iyzhang:
> > A few small comments. You might want to consider squashing the last 3 commits into a single on that just adds new pipe tests.
> 
> I actually did it intentionally, because I like to have three different commits, each of which introducing a new test that is buildable and runnable. This helps bisecting. Are you OK with me keeping it?

Yup, that's fine.

tests/rust/pipe-test/create_pipe/mod.rs
iyzhang: No, I recommend that crate::test! still return a result and nfailed be a vector instead of a usize. Then we can choose to print all the errors at the end and/or count them.

ppenna: I see. But then we wouldn't be actually collecting all the failures that happened in the test, right?

ppenna: Cool, base don our offline discussion, I've introduced a new macro that collects results in a vector.

iyzhang (approved):
Looks good

PR #661, opened 2023-04-17
[test] Remove panic from inetstack unit tests

iyzhang:
This PR removes all panics, unwraps and asserts from the inetstack unit tests.


ppenna (approved):
Looks good to me.

Thanks for working on this 😄

PR #663, opened 2023-04-17
[test] Remove panic from pal and perftools unit tests

iyzhang:
This PR updates the unit tests in the pal and perftools directories to remove panics and asserts.


ppenna (approved):
LGTM

PR #664, opened 2023-04-17
[test] Remove panic from scheduler unit test

iyzhang:
This PR removes panics, asserts, expects and unwraps from the scheduler unit tests.


ppenna (approved):
LGTM

PR #665, opened 2023-04-17
[test] Bugfix: Remove panic from runtime unit test

iyzhang:
This PR removes panics, asserts, expects and unwraps from the unit tests in the src/rust/runtime directory.


ppenna (approved):
LGTM

Issue #591, opened 2023-03-28
[catmem] Panic on EoF in `push()`


ppenna:
## Description

In Catmem LibOS, we fail if end of file (EoF) is reached in `push()`.

Note however, pipes in Catmem LibOS are unidirectional, thus we don't expect a pipe on which we are pushing data to have its EoF flag set.

## Code Snippet

https://github.com/demikernel/demikernel/blob/fa7ef1ba67ba07b44943217e09ee8c1fd91b43d2/src/rust/catmem/mod.rs#L169-L213

## Expected Behavior

Panic if the EoF flag is encountered in `push()`, since this should not happen. If it does, it means that we are not handling `close()` correctly, or the application was operating on the pipe as if it was a reading one.

PR #667, opened 2023-04-18
[catmem] Don't Check for EoF on Push

ppenna:
## Description

This PR closes https://github.com/demikernel/demikernel/issues/591

### DIFF
--- a/src/rust/catmem/mod.rs
+++ b/src/rust/catmem/mod.rs
@@ -188,14 +188,15 @@ impl CatmemLibOS {
                 match self.qtable.borrow().get(&qd) {
                     Some(queue) => {
                         let pipe: &Pipe = queue.get_pipe();
-                        let coroutine: Pin<Box<Operation>> = if pipe.eof() {
-                            // Handle end of file.
-                            Box::pin(async move {
-                                let cause: String = format!("connection reset (qd={:?})", qd);
-                                error!("pop(): {:?}", &cause);
-                                (qd, OperationResult::Failed(Fail::new(libc::ECONNRESET, &cause)))
-                            })
-                        } else {
+
+                        // TODO: review the following code once that condition is enforced by the pipe abstraction.
+                        // We do not check for EoF because pipes are unidirectional,
+                        // and if EoF is set pipe for a push-only pipe, that pipe is closed.
+                        if pipe.eof() {
+                            unreachable!("push() called on a closed pipe");
+                        }
+
+                        let coroutine: Pin<Box<Operation>> = {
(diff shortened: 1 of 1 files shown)


src/rust/catmem/mod.rs
```diff
@@ -188,14 +188,11 @@ impl CatmemLibOS {
...
                         let pipe: &Pipe = queue.get_pipe();
-                        let coroutine: Pin<Box<Operation>> = if pipe.eof() {
-                            // Handle end of file.
-                            Box::pin(async move {
-                                let cause: String = format!("connection reset (qd={:?})", qd);
-                                error!("pop(): {:?}", &cause);
-                                (qd, OperationResult::Failed(Fail::new(libc::ECONNRESET, &cause)))
-                            })
-                        } else {
+
```
iyzhang, line 191: Should we return an error here instead of a qt? We know that the coroutine won't complete without an error. Also, it looks like the push coroutine doesn't check for eof, so we might want to avoid scheduling it.

iyzhang (approved):
Looks good except for my question.

src/rust/catmem/mod.rs
ppenna: I have not followed your suggestion. Could you clarify? Note that

```rust
// We do not check for EoF because pipes are unidirectional,
// and if EoF is set pipe for a push-only pipe, that pipe is closed.
```

iyzhang: Should we still be pushing to a closed pipe? Before this change, we scheduled a different coroutine but I'm suggesting that we should skip line 195 and just return an error because the push to a closed pipe won't succeed.

ppenna: 1) We should not be pushing to a closed pipe, that is the whole point of the PR.  Also, may I remind you that we have not addressed bugs for Catmem LibOS (like we did in Catnap, when we introduced the Socket abstraction) and this behavior may not be fully functinoal.
2) Line 195 is spawning the usual co-routine, we cannot skip it.
3) I have removed only the code that spawns the co-routine for `ECONNRESET` in `push()`, we still do it on `pop()`

iyzhang: yes that was my point. We still create a coroutine here, which we should skip on eof.

iyzhang: Basically I’m asking why we don’t return on line 194

ppenna: I have added the unreachable statement, as we agreed offline.

Issue #619, opened 2023-04-05
[test] Integration Tests for Pipe `wait()`


ppenna:
## Context

We should have the following integration tests for `wait()`:

- [ ] Attempt to wait on an invalid queue token.
- [ ] Attempt to wait for a `push()` operation to complete after closing a pipe.
- [ ] Attempt to wait for a `pop()` operation to complete after closing a pipe.
- [ ] Attempt to wait for a `push()` operation to complete after asynchronously closing a pipe.
- [ ] Attempt to wait for a `pop()` operation to complete after asynchronously closing a pipe.

PR #677, opened 2023-04-19
[test] Wait Push After Closing Pipe

ppenna:
## Description

This PR partially addresses https://github.com/demikernel/demikernel/issues/619

I intentionally will submit one PR for each of the test cases, because they are more complex than usual integration tests.

## Summary of Changes

- Added support for multiple running modes in `pipe-test`
- Added an integration test for "Attempt to wait for a push() operation to complete after closing a pipe".

### DIFF
--- a/tests/rust/pipe-test/push_wait/client.rs
+++ b/tests/rust/pipe-test/push_wait/client.rs
@@ -1,0 +1,149 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+//======================================================================================================================
+// Imports
+//======================================================================================================================
+
+use ::anyhow::Result;
+use ::demikernel::{
+    demi_sgarray_t,
+    runtime::types::demi_opcode_t,
+    LibOS,
+    QDesc,
+    QToken,
+};
+use ::std::slice;
+
+//======================================================================================================================
+// Structures
+//======================================================================================================================
+
+/// Pipe client
+pub struct PipeClient {
+    /// Underlying libOS.
+    libos: LibOS,
+    /// Underlying pipe descriptor.
+    pipeqd: Option<QDesc>,
+}
+
+//======================================================================================================================
+// Associated Functions
+//======================================================================================================================
+
+impl PipeClient {
+    /// Creates a new pipe client.
+    pub fn new(mut libos: LibOS, pipe_name: String) -> Result<Self> {
+        let pipeqd: QDesc = libos.open_pipe(&format!("{}:rx", pipe_name))?;
+        Ok(Self {
+            libos,
+            pipeqd: Some(pipeqd),
+        })
+    }
+
+    // Runs the target pipe client.
+    pub fn run(&mut self) -> Result<()> {
+        // Push some data.
+        // The number of pushes is set to an arbitrary value,
+        // but a small one to avoid contention on the underling ring buffer.
+        for _ in 0..16 {
+            self.push_and_wait()?;
+        }
+
+        // Push again, but don't wait the operation to complete.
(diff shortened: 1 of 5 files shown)


tests/rust/pipe-test/push_wait/client.rs
```diff
@@ -0,0 +1,120 @@
...
+            Ok(qt) => qt,
+            Err(e) => {
+                // Clean up open pipe on error.
+                // Clean up scatter-gather array on error.
+                // FIXME: https://github.com/demikernel/demikernel/issues/638
+                anyhow::bail!("push() failed (error={:?})", e)
+            },
+        };
+
+        // Succeed to release scatter-gather-array.
```
iyzhang, line 68: This is not correct semantics, we should not release the scatter-gather army until after the wait on push returns either ok or error. I would recommend moving this match statement to after line 96.

tests/rust/pipe-test/push_wait/client.rs
```diff
@@ -0,0 +1,120 @@
...
+    /// Makes a scatter-gather array.
+    fn mksga(&mut self, size: usize, value: u8) -> Result<demi_sgarray_t> {
+        // Allocate scatter-gather array.
+        let sga: demi_sgarray_t = match self.libos.sgaalloc(size) {
+            Ok(sga) => sga,
+            Err(e) => anyhow::bail!("failed to allocate scatter-gather array: {:?}", e),
+        };
+
+        // Ensure that scatter-gather array has the requested size.
+        assert_eq!(sga.sga_segs[0].sgaseg_len as usize, size);
```
iyzhang, line 110: Can you change this to demikernel::ensure_eq! ?

iyzhang (changes requested):
Is this set of tests designed to be run on two machines or one? I'm a bit confused because I thought these integration tests were all running on a single node and all of our multi-node tests are system tests located in examples.

tests/rust/pipe-test/push_wait/client.rs
ppenna: Well, this semantic is actually valid - it may not be correct from a design perspecive. Whenever we `push()`, we clone the scatter-gather array and bump the reference counter. We do it for all LibOSes, but here is the concerned code snippet for Catmem LibOS:

https://github.com/demikernel/demikernel/blob/f99757a9a8d10084cb8da579126b52ef46724e2d/src/rust/catmem/mod.rs#L176-L179

ppenna:
> Is this set of tests designed to be run on two machines or one? I'm a bit confused because I thought these integration tests were all running on a single node and all of our multi-node tests are system tests located in examples.

With changes in this PR, I added the ability for this integration test (`pipe-test`) to run in 2 machines. This enables us to have a single application that runs both standalone and E2E integration tests for a given transport (pipes for this PR).

tests/rust/pipe-test/push_wait/client.rs
ppenna: Sure.

tests/rust/pipe-test/push_wait/client.rs
iyzhang: Ah ok this is great, in our C++ code, this was not possible so we set the semantics in the C/C++ API accordingly

iyzhang (approved):
all of this make sense.

iyzhang (approved):
This looks great. Thanks for implementing drop functions for these tests for full cleanup.

tests/rust/pipe-test/push_wait/client.rs
ppenna: Yeah.. We have shallow copies on the rust implementation :D

Issue #552, opened 2023-03-15
[pdpix] When Closing a Queue Descriptor Pending Operations Should be Cancelled


ppenna:
## Context

When closing a queue descriptor, all pending operations shall be cancelled.

## Related Issues

- https://github.com/demikernel/demikernel/issues/498
```[tasklist]
### Tasks
- [x] Catnap
- [x] Catmem
- [x] Catloop PR #785
- [ ] Catcollar
- [x] Catpowder
- [x] Catnip PR #945 
```

Issue #521, opened 2023-03-07
[PDPIX] Return error code on wait() after close()


iyzhang:
## Description

Currently, if an application is holding any open qtokens and closes the associated queue descriptor (or it is closed from the other end), the code will crash inside the future associated with that qtoken the next time the application waits on that qtoken.

## How to Reproduce

1. Call pop() on qd, receive qtoken
2. Call close() on qd (or receive a signal from the other end of the connection to close).
3. Call wait() on qtoken

## Expected Behavior

The next call to wait() with that qtoken should not crash. There are two options for correct behavior: 
1. return an Ok result with a zero-length buffer
2. return a Failed result with ECONNRESET

In POSIX, the last read after all data from the connection has been consumed will return with option 1, every subsequent read will return 2.

For Demikernel, I propose the following behavior: when close is called (or a closing FIN received), Demikernel will immediately mark all pending qtokens (and their associated co-routine) as runnable and will return a Fail(ECONNRESET). There is no guaranteed order that the pending qtokens will complete. On receiving the first ECONNRESET error code, the application is free to cancel/drop all other qtokens on queue descriptor, which will let us free the per-queue meta-data that we are keeping, but the application is not required to cancel or drop them. If the application holds them indefinitely, then we will keep the associated queue meta-data, if the application puts them into a wait or wait_any, then they will return with Fail(ECONNRESET).

## Related Issues
* #367 
* #369

PR #683, opened 2023-04-20
[test] Wait Pop After Closing Pipe

ppenna:
## Description

This commit partially addresses https://github.com/demikernel/demikernel/issues/619

## Related Issues

- https://github.com/demikernel/demikernel/issues/684
- https://github.com/demikernel/demikernel/issues/521
- https://github.com/demikernel/demikernel/issues/552

### DIFF
--- a/tests/rust/pipe-test/pop_wait/client.rs
+++ b/tests/rust/pipe-test/pop_wait/client.rs
@@ -1,0 +1,121 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+//======================================================================================================================
+// Imports
+//======================================================================================================================
+
+use ::anyhow::Result;
+use ::demikernel::{
+    demi_sgarray_t,
+    runtime::types::demi_opcode_t,
+    LibOS,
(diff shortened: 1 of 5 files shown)


tests/rust/pipe-test/pop_wait/client.rs
```diff
@@ -0,0 +1,127 @@
...
+//======================================================================================================================
+// Structures
+//======================================================================================================================
+
+/// Pipe client
+pub struct PipeClient {
+    /// Underlying libOS.
+    libos: LibOS,
+    /// Underlying pipe descriptor.
+    pipeqd: Option<QDesc>,
```
iyzhang, line 28: I don't see any reason for this to be an Option value. Looks like it is set in the new function and removed in drop.

iyzhang (approved):
LGTM

tests/rust/pipe-test/pop_wait/client.rs
ppenna: Indeed, it does not. I'll remove the `Option` value before merging.

Issue #684, opened 2023-04-20
[catmem] Don't Infinite Loop in `push_eof()`


ppenna:
## Description

When we fail to push EoF, we may loop indefinetly.

## Code Snippet

https://github.com/demikernel/demikernel/blob/a7efdcb3400b670ce77d007238288dba736e0b2b/src/rust/catmem/mod.rs#L143-L156

PR #686, opened 2023-04-24
[catmem] Don't Push EoF Indefinitely

ppenna:
## Description

This PR closes https://github.com/demikernel/demikernel/issues/684

## Summary of Changes

- Removed infinite loop around `push_eof()`
- Changed `close()` to release queue descriptor even if `push_eof()` fails.


iyzhang (approved):
LGTM. I assume we need this because the queue might be full and the other end isn't popping from it so we have nowhere to put the EoF.

ppenna:
> LGTM. I assume we need this because the queue might be full and the other end isn't popping from it so we have nowhere to put the EoF.

Yes, you are correct.

PR #691, opened 2023-04-25
[catmem] Release Queue Descriptor Even If `close()` Fails

ppenna:
## Description

This PR is a follow up of https://github.com/demikernel/demikernel/pull/686

## Summary of Changes

- Changed `close()` to release the queue descriptor even if we fail to `push_eof()`.

https://github.com/demikernel/demikernel/pull/686

### DIFF
--- a/src/rust/catmem/mod.rs
+++ b/src/rust/catmem/mod.rs
@@ -167,17 +167,27 @@ impl CatmemLibOS {
     /// Closes a memory queue.
     pub fn close(&mut self, qd: QDesc) -> Result<(), Fail> {
         trace!("close() qd={:?}", qd);
-        let mut qtable = self.qtable.borrow_mut();
-        match qtable.get(&qd) {
-            Some(queue) => Self::push_eof(queue.get_pipe().buffer())?,
-            None => {
-                let cause: String = format!("invalid queue descriptor (qd={:?})", qd);
-                error!("close(): {}", cause);
-                return Err(Fail::new(libc::EBADF, &cause));
-            },
+        let mut qtable: RefMut<IoQueueTable<CatmemQueue>> = self.qtable.borrow_mut();
+
+        // Check if queue descriptor is valid.
+        if qtable.get(&qd).is_none() {
+            let cause: String = format!("invalid queue descriptor (qd={:?})", qd);
+            error!("close(): {}", cause);
+            return Err(Fail::new(libc::EBADF, &cause));
+        }
+
+        // Attempt to push EoF.
+        let result: Result<(), Fail> = {
+            // It is safe to call expect() here because the queue descriptor is guaranteed to be valid.
+            let queue: &CatmemQueue = qtable.get(&qd).expect("queue descriptor should be valid");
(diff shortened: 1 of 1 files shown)


src/rust/catmem/mod.rs
```diff
@@ -167,17 +167,27 @@ impl CatmemLibOS {
...
+        // Check if queue descriptor is valid.
+        if qtable.get(&qd).is_none() {
+            let cause: String = format!("invalid queue descriptor (qd={:?})", qd);
+            error!("close(): {}", cause);
+            return Err(Fail::new(libc::EBADF, &cause));
+        }
+
+        // Attempt to push EoF.
+        let result: Result<(), Fail> = {
+            // It is safe to call expect() here because the queue descriptor is valid guaranteed to be valid.
```
iyzhang, line 181: take out "is valid"

iyzhang, line 181: take out "is valid"

iyzhang (approved):
LGTM

iyzhang:
instead of freeing the queue descriptor, should we eventually have states and the queue should be placed into a CLOSING state, which continues to try to push the EoF?

ppenna:
> instead of freeing the queue descriptor, should we eventually have states and the queue should be placed into a CLOSING state, which continues to try to push the EoF?

Definitely we should have some abstraction for the states of the queue itself, such as we have in Catnap Loop. However,  the problem here about continuing try to push EoF is that we are not sure that this is due because the ring buffer is full or the other end has closed the pipe. In the latter case, we don't have a signal mechanism, because a pipe is unidirectional. Any ideias?

src/rust/catmem/mod.rs
ppenna: Sure 😄

iyzhang:
> > instead of freeing the queue descriptor, should we eventually have states and the queue should be placed into a CLOSING state, which continues to try to push the EoF?
> 
> Definitely we should have some abstraction for the states of the queue itself, such as we have in Catnap Loop. However, the problem here about continuing try to push EoF is that we are not sure that this is due because the ring buffer is full or the other end has closed the pipe. In the latter case, we don't have a signal mechanism, because a pipe is unidirectional. Any ideias?

My thought was that we should at least keep trying to push EoF regardless of whether the other end will read it because it closed the pipe or stopped pulling things out of it.

ppenna:
> > > instead of freeing the queue descriptor, should we eventually have states and the queue should be placed into a CLOSING state, which continues to try to push the EoF?
> > 
> > 
> > Definitely we should have some abstraction for the states of the queue itself, such as we have in Catnap Loop. However, the problem here about continuing try to push EoF is that we are not sure that this is due because the ring buffer is full or the other end has closed the pipe. In the latter case, we don't have a signal mechanism, because a pipe is unidirectional. Any ideias?
> 
> My thought was that we should at least keep trying to push EoF regardless of whether the other end will read it because it closed the pipe or stopped pulling things out of it.

Sure, I've opened an issue for this.

Issue #698, opened 2023-04-26
[pdpix] Have a System-Wide Maximum Size for a Receive Buffer


ppenna:
## Context

We are missing a system-wide maximum size for a receive buffer.

Consequently, we have multiple definitions of this variable across the code base.

https://github.com/demikernel/demikernel/blob/70b858eeea4dc9476d3f00e93703f1d37ef81e23/src/rust/catnap/coroutines/pop.rs#L26-L27

https://github.com/demikernel/demikernel/blob/70b858eeea4dc9476d3f00e93703f1d37ef81e23/src/rust/catcollar/mod.rs#L74-L75

https://github.com/demikernel/demikernel/blob/70b858eeea4dc9476d3f00e93703f1d37ef81e23/src/rust/catmem/futures/pop.rs#L43-L44

https://github.com/demikernel/demikernel/blob/70b858eeea4dc9476d3f00e93703f1d37ef81e23/src/rust/catpowder/runtime/network.rs#L64-L66

## Proposed Solution

- Have a `runtime::limits` module where we place all system-wide constants for maximum values.
- Create a `RECVBUF_SIZE_MAX` constant for holding the maximum size of receive buffer.

## Related Issues

- https://github.com/demikernel/demikernel/issues/559

PR #699, opened 2023-04-26
[pdpix] Have a System-Wide Maximum Size for a Receive Buffer

ppenna:
## Description

This PR closes https://github.com/demikernel/demikernel/issues/698

## Summary of Changes

- Added a system-wide constant for the maximum size of receive buffers.
- Changed Catnap LibOS to use that system-wide constant.
- Changed Catmem LibOS to use that system-wide constant.
- Changed Catcollar LibOS to use that system-wide constant.
- Changed Catpowder LibOS to use that system-wide constant.

### DIFF
--- a/src/rust/catcollar/mod.rs
+++ b/src/rust/catcollar/mod.rs
@@ -32,6 +32,7 @@ use crate::{
     pal::linux,
     runtime::{
         fail::Fail,
+        limits,
         memory::{
             DemiBuffer,
             MemoryRuntime,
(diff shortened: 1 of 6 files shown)


iyzhang (approved):
The code looks fine to me but we might not want to use a global RECVBUF size. For example, RDMA will accommodate up to 1Gb sends and receives.

ppenna:
> The code looks fine to me but we might not want to use a global RECVBUF size. For example, RDMA will accommodate up to 1Gb sends and receives.

Sure. Let us revisit this when we have a RDMA LibOS merged in.

Issue #571, opened 2023-03-23
[test] Integration Tests for TCP `async_close`


iyzhang:
## Context

We should have the following integration tests for TCP `async_close()`:

- [x] Attempt to asynchronously close an invalid socket. 816ebe0cfaa9a8fec792edac1c4f7af5d54bb902
- [x] Attempt to asynchronously close a TCP socket multiple times. 816ebe0cfaa9a8fec792edac1c4f7af5d54bb902
- [x] Attempt to asynchronously close a TCP socket that is not bound. 816ebe0cfaa9a8fec792edac1c4f7af5d54bb902
- [x] Attempt to asynchronously close a TCP socket that is bound. 816ebe0cfaa9a8fec792edac1c4f7af5d54bb902
- [x] Attempt to asynchronously  close a TCP socket that is listening. 816ebe0cfaa9a8fec792edac1c4f7af5d54bb902
- [x] Attempt to asynchronously close a TCP socket that is accepting. 816ebe0cfaa9a8fec792edac1c4f7af5d54bb902
- [x] Attempt to asynchronously close a TCP socket that is connecting. 816ebe0cfaa9a8fec792edac1c4f7af5d54bb902

### Additional Test Case for Asynchronous Close Multiple Times (i)

````
sockqd = socket();
qt1 = async_close(sockqd);
qt2 = async_close(sockqd);
wait(qt1);
wait(qt2);
````

### Additional Test Case for Asynchronous Close Multiple Times (ii)

````
sockqd = socket();
qt1 = async_close(sockqd);
qt2 = async_close(sockqd);
wait(qt2);
wait(qt1);
````

iyzhang:
I'm not sure exactly how to test this since we have no remote to connect to.

ppenna:
> I'm not sure exactly how to test this since we have no remote to connect to.

@iyzhang this test based on tcp-close. Sync up with @anandbonde .

iyzhang:
Ok, adding @anandbonde to this issue so we can coordinate.

anandbonde:
> > I'm not sure exactly how to test this since we have no remote to connect to.
> 
> @iyzhang this test based on tcp-close. Sync up with @anandbonde .

Yes, that test would require 2 endpoints. The tcp-close utility does exactly this for "close". We can file an issue to enhance the tcp-close utility to use close/async_close based on an input flag; and fix it in a separate PR.

anandbonde:
[...]

PR #705, opened 2023-04-27
[test] Add more async_close() tests

anandbonde:
Partially addresses #571


iyzhang (approved):
LGTM. I’ll note that the second test is probably not needed after the fix because there will only be one queue token, with the second call to async_close failing.

PR #710, opened 2023-05-01
[scheduler] Halve WakerPage struct size

### LINKED ISSUE #175: [scheduler] Remove drop bits
Description
========

We currently have 32 bytes in the `WakerPage` unused. We should make something useful out of this space.

One simple idea would be to halve the size of the structure itself.

anandbonde:
Related to #175

### DIFF
--- a/src/rust/scheduler/waker64.rs
+++ b/src/rust/scheduler/waker64.rs
@@ -15,7 +15,7 @@ use ::std::{
 //==============================================================================
 
 /// Log2 of [WAKER_BIT_LENGTH].
-pub const WAKER_BIT_LENGTH_SHIFT: usize = 6;
+pub const WAKER_BIT_LENGTH_SHIFT: usize = 5;
 
 /// Number of Bits in a [Waker64]
 pub const WAKER_BIT_LENGTH: usize = 1 << WAKER_BIT_LENGTH_SHIFT;
--- a/src/rust/scheduler/page/page.rs
+++ b/src/rust/scheduler/page/page.rs
@@ -15,7 +15,7 @@ use crate::scheduler::waker64::{
 //==============================================================================
 
 /// Size of Pages (in bytes)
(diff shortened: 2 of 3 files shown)


src/rust/scheduler/waker64.rs
```diff
@@ -15,7 +15,7 @@ use ::std::{
 //==============================================================================
 
 /// Log2 of [WAKER_BIT_LENGTH].
-pub const WAKER_BIT_LENGTH_SHIFT: usize = 6;
+pub const WAKER_BIT_LENGTH_SHIFT: usize = 5;
 
 /// Number of Bits in a [Waker64]
 pub const WAKER_BIT_LENGTH: usize = 1 << WAKER_BIT_LENGTH_SHIFT;
```
ppenna, line 21: This changes more than it appears.

ppenna (changes requested):
It looks like you were attempting to halve the structure entirely, as it is suggested in the issue description. Unfortunately, this would require major rework, because sizes of `WakerPage` and `Waker64` are very coupled. The correct fix int hat direction would be change everything to be 32-bit long. Note however, that this would have a major negative impact on 64 bit-machine, when traversing the scheduler bitmap. Instead, I would suggest us to really use that unsued field for something more interesting, such as introducing a priority-based scheduling with 2 scheduling levels (zero's and ones).

For instance, the way that we traverse

src/rust/scheduler/waker64.rs
iyzhang: Lol, this is the most mysterious comment that I have ever seen.

ppenna: I just wanted to point out that this is the place where we actually change the length of `WAKER`s.

iyzhang:
This issue is related to a recent one that I posted #723

anandbonde:
Closing this PR, @iyzhang will handle this in a separate PR.

PR #711, opened 2023-05-02
[runtime] Remove QResult

### LINKED ISSUE #503: [runtime] Remove QResult
## Context

The QResult data structure is no longer used, so we should deprecate and/or remove it from the runtime.
 
## Proposed Solution

Remove QResult from runtime and potentially rename file/module to operation.

anandbonde:
Closes #503


ppenna (approved):
Looks goof to me.

Thanks for working on this 😄

PR #724, opened 2023-05-02
[catnap] Implement drop for cleanup on exit

### LINKED ISSUE #668: [catnap] Implement drop for the libOS for correct clean up
## Description

Currently our libOSes do not have any code for cleaning up resources on exit, specifically for components that deal with external resources. For example, Catnap should ensure that the OS socket is closed when the socket data structure is dropped and when Catnap exits, it should drop everything from the IoQueueTable to ensure that all lingering sockets are closed.

## How to Reproduce

1. Run Catnap 
2. Crash on unit test
3. Cannot run again because of lingering open sockets

## Related Issues

- #633 
- #642 
- #640

iyzhang:
This PR close #668. I explicitly close each socket with a valid file descriptor. As an aside, we may want to move any code that only deals with a single queue or socket into the CatnapQueue or Socket data structure. For example, the RawFd associated with the socket should probably be stored inside the socket.

### DIFF
--- a/src/rust/catnap/mod.rs
+++ b/src/rust/catnap/mod.rs
@@ -694,6 +694,25 @@ impl CatnapLibOS {
     }
 }
 
+//======================================================================================================================
+// Trait Implementations
+//======================================================================================================================
+
+impl Drop for CatnapLibOS {
+    // Releases all sockets allocated by Catnap.
+    fn drop(&mut self) {
+        for (_, queue) in self.qtable.borrow().get_values() {
+            if let Some(fd) = queue.get_fd() {
+                if unsafe { libc::close(fd) } != 0 {
+                    let errno: libc::c_int = unsafe { *libc::__errno_location() };
+                    error!("close() failed (error={:?}", errno);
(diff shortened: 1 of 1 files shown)


src/rust/catnap/mod.rs
```diff
@@ -694,6 +694,25 @@ impl CatnapLibOS {
...
+
+impl Drop for CatnapLibOS {
+    // Releases all sockets allocated by Catnap.
+    fn drop(&mut self) {
+        for (_, queue) in self.qtable.borrow().get_values() {
+            if let Some(fd) = queue.get_fd() {
+                if unsafe { libc::close(fd) } != 0 {
+                    let errno: libc::c_int = unsafe { *libc::__errno_location() };
+                    println!("ERROR: close() failed (error={:?}", errno);
+                    println!("WARN: leaking fd={:?}", fd);
```
ppenna, line 709: Could you change `println!()` for `error!()` and `warn!()`?

ppenna (approved):
Looks good to me.

Issue #704, opened 2023-04-27
[pdpix] Rust Compiler Optimizes Size of Structures


ppenna:
## Description

The following Demikernel data structures are not packed, and the Rust compiler may end up optimizing their size:
- `demi_sgaseg_t`
- `demi_sgarray_t`
- `demi_accept_result_t`
- `demi_qr_value_t`
- `demi_qresult_t`

This bug may not be noticed when relying on Demikernel Rust API, but it is critical for correct interoperability with other languages.

## How to Reproduce

1. Check out `65b98a37733d10e7fec86bf5c2ede51656a44e5c`
2. Download [this patch](https://github.com/demikernel/demikernel/files/11345010/0001-test-Enhancement-Unit-Test-f-Size-of-Structures.patch), which introduces some unit-tests that expose this bug.
3. Apply the patch `git am -3 0001-test-Enhancement-Unit-Test-f-Size-of-Structures.patch`
4. Run the unit tests `make DEBUG=yes test-unit-rust TEST_UNIT=test_size`
5. See error.

Alternatively, [check CI logs](https://github.com/demikernel/demikernel/actions?query=branch%3Abugfix-pdpix-sizes).

PR #726, opened 2023-05-03
[pdpix] Don't Optimize Size of Structures

ppenna:
## Description

This PR closes https://github.com/demikernel/demikernel/issues/704.

### DIFF
--- a/examples/rust/udp-echo.rs
+++ b/examples/rust/udp-echo.rs
@@ -192,7 +192,7 @@ impl Application {
                 demi_opcode_t::DEMI_OPC_POP => {
                     let sockqd: QDesc = qr.qr_qd.into();
                     let sga: demi_sgarray_t = unsafe { qr.qr_value.sga };
-                    let saddr: SocketAddrV4 = match Self::sockaddr_to_socketaddrv4(unsafe { &qr.qr_value.sga.sga_addr })
+                    let saddr: SocketAddrV4 = match Self::sockaddr_to_socketaddrv4(&unsafe { qr.qr_value.sga.sga_addr })
                     {
                         Ok(saddr) => saddr,
                         Err(e) => {
--- a/src/rust/runtime/types/memory.rs
+++ b/src/rust/runtime/types/memory.rs
@@ -3,42 +3,39 @@
 
 #![allow(non_camel_case_types)]
 
-//==============================================================================
+//======================================================================================================================
 // Imports
-//==============================================================================
-
-use ::libc::c_void;
+//======================================================================================================================
 
 use crate::pal::data_structures::SockAddr;
 
-//==============================================================================
+//======================================================================================================================
 // Constants
-//==============================================================================
+//======================================================================================================================
 
 /// Maximum Length for Scatter-Gather Arrays
 pub const DEMI_SGARRAY_MAXLEN: usize = 1;
 
-//==============================================================================
+//======================================================================================================================
 // Structures
-//==============================================================================
+//======================================================================================================================
 
 /// Scatter-Gather Array Segment
-#[repr(C)]
+#[repr(C, packed)]
 #[derive(Copy, Clone)]
 pub struct demi_sgaseg_t {
     /// Underlying data.
-    pub sgaseg_buf: *mut c_void,
+    pub sgaseg_buf: *mut libc::c_void,
     /// Length of underlying data.
     pub sgaseg_len: u32,
 }
 
 /// Scatter-Gather Array
-// TODO: Review the inclusion of the sga_addr field (only used for recvfrom?) in this structure.
-#[repr(C)]
+#[repr(C, packed)]
 #[derive(Copy, Clone)]
 pub struct demi_sgarray_t {
     /// Reserved.
-    pub sga_buf: *mut c_void,
+    pub sga_buf: *mut libc::c_void,
     /// Number of segments in this scatter-gather array.
     pub sga_numsegs: u32,
     /// Scatter-gather array segments.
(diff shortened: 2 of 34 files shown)


examples/rust/udp-echo.rs
```diff
@@ -192,8 +192,8 @@ impl Application {
                 demi_opcode_t::DEMI_OPC_POP => {
                     let sockqd: QDesc = qr.qr_qd.into();
                     let sga: demi_sgarray_t = unsafe { qr.qr_value.sga };
-                    let saddr: SocketAddrV4 = match Self::sockaddr_to_socketaddrv4(unsafe { &qr.qr_value.sga.sga_addr })
```
iyzhang, line 195: Is this important to do in this change?

src/rust/runtime/types/memory.rs
```diff
@@ -3,46 +3,79 @@
...
-    pub sgaseg_buf: *mut c_void,
+    pub sgaseg_buf: *mut libc::c_void,
     /// Length of underlying data.
     pub sgaseg_len: u32,
 }
 
 /// Scatter-Gather Array
-// TODO: Review the inclusion of the sga_addr field (only used for recvfrom?) in this structure.
-#[repr(C)]
+#[repr(C, packed)]
```
iyzhang, line 34: Should we leave this todo in?

src/rust/runtime/types/memory.rs
```diff
@@ -3,46 +3,79 @@
...
     /// Number of segments in this scatter-gather array.
     pub sga_numsegs: u32,
     /// Scatter-gather array segments.
     pub sga_segs: [demi_sgaseg_t; DEMI_SGARRAY_MAXLEN],
     /// Source address of the data contained in this scatter-gather array (if present).
     pub sga_addr: SockAddr,
 }
+
+#[cfg(test)]
+mod test {
```
iyzhang, line 48: For consistency, these tests should probably return a Result<()> and use demikernel::ensure_eq!

src/rust/runtime/types/ops.rs
```diff
@@ -51,8 +47,57 @@ pub union demi_qr_value_t {
 #[repr(C)]
 pub struct demi_qresult_t {
     pub qr_opcode: demi_opcode_t,
-    pub qr_qd: c_int,
+    pub qr_qd: u32,
```
iyzhang, line 50: what is the reasoning behind using u32 for queue descriptors and u64 for return values? Is it for alignment? Since the return value is a libc errno, it must also be u32.

src/rust/runtime/types/ops.rs
```diff
@@ -51,8 +47,57 @@ pub union demi_qr_value_t {
...
+
+#[cfg(test)]
+mod test {
+
+    use super::*;
+    use ::std::mem;
+
+    /// Tests if `demi_accept_result_t` has the expected size.
+    #[test]
+    fn test_size_demi_accept_result_t() {
```
iyzhang, line 64: Again for consistency, we should have these tests return a Result<()> and use demikernel::ensure_eq

tests/rust/pipe-test/async_close/mod.rs
```diff
@@ -131,7 +131,9 @@ fn async_close_pipe_multiple_times_2(libos: &mut LibOS, pipe_name: &str) -> Resu
     // Poll once to ensure the async_close() coroutine runs and finishes the close.
     match libos.wait(qt1, Some(Duration::from_micros(0))) {
         Ok(qr) if qr.qr_opcode == demi_opcode_t::DEMI_OPC_CLOSE && qr.qr_ret == 0 => closed = true,
-        Ok(qr) if qr.qr_opcode == demi_opcode_t::DEMI_OPC_FAILED && qr.qr_ret == libc::ECANCELED => cancelled = true,
+        Ok(qr) if qr.qr_opcode == demi_opcode_t::DEMI_OPC_FAILED && qr.qr_ret == libc::ECANCELED as i64 => {
```
iyzhang, line 134: Does it make sense to cast the libc errno as i64 or would it make more sense to cast qr.qr_ret as i32?

iyzhang (approved):
I left a few relatively minor comments, which might be due to my misunderstanding, but other wise looks good.

examples/rust/udp-echo.rs
ppenna: It is not. Thanks for noting that.

src/rust/runtime/types/memory.rs
ppenna: Well noted.

src/rust/runtime/types/ops.rs
ppenna: Well noted.

src/rust/runtime/types/ops.rs
ppenna: Here is the reasons:
- Rationale for a u32 queue descriptor: https://github.com/demikernel/demikernel/pull/427, https://github.com/demikernel/demikernel/pull/441, https://github.com/demikernel/demikernel/issues/430
- Rationale for a u64 queue token: it ensures that the structure would have the correct size and it is big enough for holding a i32. We could alternatively: (a) leave it u32, but introduce a u32 unused field for padding, which would require even more intrusive changes; or (b) leave it u32 and pack the structure, which would require massive changes to the code base.

tests/rust/pipe-test/async_close/mod.rs
ppenna: No it does not make sense. We should always cast from the smallest to the largest number-space when comparing. Example:

```
let x: u64 = 0xffffffff_ffffffff;
let y: u32 = 0xffffffff;
(x as u32) == y // This comparison wrongly suceeds.
x == (y as u64) // This comparison correctly fails.
```

Issue #625, opened 2023-04-06
[runtime] Introduce `CLOSING` State for Sockets


ppenna:
## Context

The socket abstraction should have a `CLOSING` state.

PR #727, opened 2023-05-03
[runtime] Closing State for Sockets

ppenna:
## Description

This PR closes https://github.com/demikernel/demikernel/issues/625.

### DIFF
--- a/src/rust/catnap/mod.rs
+++ b/src/rust/catnap/mod.rs
@@ -423,11 +423,10 @@ impl CatnapLibOS {
         match qtable.get_mut(&qd) {
             Some(queue) => match queue.get_fd() {
                 Some(fd) => {
-                    // Create a closed socket.
-                    let closed_socket: Socket = {
-                        let socket: &Socket = queue.get_socket();
-                        socket.close()?
-                    };
+                    // Set socket as closing.
+                    let socket: &Socket = queue.get_socket();
+                    let closing_socket: Socket = socket.close()?;
+                    queue.set_socket(&closing_socket);
 
                     // Close underlying socket.
                     if unsafe { libc::close(fd) } != 0 {
@@ -439,8 +438,12 @@ impl CatnapLibOS {
 
                     // Cancel all pending operations.
                     queue.cancel_pending_ops(Fail::new(libc::ECANCELED, "This queue was closed"));
-                    // Update socket.
+
+                    // Update socket state.
+                    let socket: &Socket = queue.get_socket();
+                    let closed_socket: Socket = socket.closed()?;
                     queue.set_socket(&closed_socket);
+
                     qtable.free(&qd);
                     Ok(())
                 },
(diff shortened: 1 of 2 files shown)


src/rust/catnap/mod.rs
```diff
@@ -439,8 +438,12 @@ impl CatnapLibOS {
 
                     // Cancel all pending operations.
                     queue.cancel_pending_ops(Fail::new(libc::ECANCELED, "This queue was closed"));
-                    // Update socket.
+
+                    // Update socket state.
+                    let socket: &Socket = queue.get_socket();
```
iyzhang, line 443: This code is redundant since the socket is immediately destroyed when the queue is freed from the queue table.

src/rust/catnap/mod.rs
```diff
@@ -458,12 +461,14 @@ impl CatnapLibOS {
...
-                    let closed_socket: Socket = {
-                        let socket: &Socket = queue.get_socket();
-                        socket.close()?
-                    };
+                    // Set socket as closing.
+                    let socket: &Socket = queue.get_socket();
+                    let closing_socket: Socket = socket.close()?;
+                    queue.set_socket(&closing_socket);
+
+                    // Create a closed socket.
```
iyzhang, line 469: I think it would make more sense to put this into the async block since we already have the mutable queue reference there. Having it here is quite confusing, so I would recommend moving to line 488.

iyzhang (changes requested):
In general, the closed state and functions are redundant, so I would recommend removing them in favor of just having a closing state, which represents the application having called close() at least once on the socket. The closed state only occurs when we have successfully closed the socket, at which point we free the socket data structure (and queue data structure) because we are no longer holding external resources.

This would also make the code clearer around the socket.close() and the socket.closed() functions and the code in async_close, which calls socket.close() and then socket.closed() immediately after. 

I would recommend that we draw a state diagram in the socket file to show the different states that the socket can be in and when transitions happen.

src/rust/catnap/mod.rs
ppenna: It may be, but it is always good to be extra cautious - what if some bug is introduced in the `free()` function?

The thing is that this layer is (currently) responsible for diving the state transitions of a socket. Itis not safe to assume that we can leave it in a dangling state hoping that another module will do its job correctly and thus we "wont have a problem".

src/rust/catnap/mod.rs
ppenna: Sure.

ppenna:
> In general, the closed state and functions are redundant, so I would recommend removing them in favor of just having a closing state, which represents the application having called close() at least once on the socket. The closed state only occurs when we have successfully closed the socket, at which point we free the socket data structure (and queue data structure) because we are no longer holding external resources.

Well, I have a different opinion on this. The `closing` and `closed` states are different from one another:
- When a socket is `closing`, any new operations on that socket are no longer allowed, but some resources are still assigned to that socket.
- When a socket is in a `closed` state, all resources that were associated to that socket are ensure to be released.


> This would also make the code clearer around the socket.close() and the socket.closed() functions and the code in async_close, which calls socket.close() and then socket.closed() immediately after.

I believe that the clarity here may be addressed by naming the functions differently. Note that functions actually trigger the state transitions. I am open to ideas.

> I would recommend that we draw a state diagram in the socket file to show the different states that the socket can be in and when transitions happen.

Sure. I agree that we may have that kind of documentation. I have opened https://github.com/demikernel/demikernel/issues/728 to keep track of this.

src/rust/catnap/mod.rs
iyzhang: Ok, this is fine with me. Please carefully document the difference between the two and add comments for where you make the state transitions in the libOS, for example so that it is clear why we update the socket twice here.

iyzhang (approved):

src/rust/catnap/mod.rs
ppenna: Sure.

Issue #177, opened 2022-02-03
[runtime] Make `RECEIVE_BATCH_SIZE` Generic


ppenna:
Description
========

We currently have the `RECEIVE_BATCH_SIZE` harcode as a constant value in `network::consts`. Instead we should make this genericly available in `network::NetworkRuntime`;

Related Issues
==========

demikernel/runtime#9

PR #732, opened 2023-05-08
[runtime] Make `RECEIVE_BATCH_SIZE` Generic

ppenna:
## Description

This PR closes https://github.com/demikernel/demikernel/issues/177

## Summary of Changes

- Introduced constant parameter to `NetworkRuntime` trait.
- Changed code base accordingly.

Note that making the network runtime generic over a constant for the `RECEIVE_BATCH_SIZE` enables us to effectively have each LibOS having its own value for that constant, while enabling multiple LibOses to coexist.

This commit addresses
https://github.com/demikernel/demikernel/issues/177


anandbonde (approved):
LGTM

Issue #134, opened 2022-06-13
[pdpix] Fix Network Types to be Compatible With Each Other


BrianZill:
Now that we've replaced our home-grown Port16 type with u16, and the associated Ipv4Endpoint struct with std::net::SocketAddrV4, we should have an easier time of interfacing our code with Rust standard library code (which uses standard types).

But there are some places where existing code was converted in place that are no longer needed and/or make sense.   And there are some compatibility functions we still have that are no longer needed (or are now inefficient).

Example inefficiency problem:
catcollar/mod.rs (and catnap/mod.rs) contain a parse_addr function that (now) converts a std::net::SockAddrV4 into a ::nix::sys::socket::SockaddrStorage.  It does this by pulling the addr and port out of the input SockAddrV4 and then creating a new SockAddrV4 from that addr and port.  And then uses SockaddrStorage::from() to create the SockaddrStorage from the new SockAddrV4.  There is no point to creating this additional SockAddrV4, as the SockaddrStorage::from() could just be called using the input SockAddrV4.  Moreover, there is no real need for parse_addr anymore, as SockaddrStorage::from() by itself does the same thing directly.

Another example:
catcollar/mod.rs (and catnap/mod.rs) contain a pack_result function that (essentially) converts an OperationResult into a demi_qresult_t.  For OperationResult::Pop, it goes through some unnecessary convolutions to create the sga_addr field (a sockaddr_in) from the popped SocketAddrV4 (which is also just a sockadd_in), when it could just copy the bits over (if Rust would let it).

ppenna:
@BrianZill and @anandbonde this is related to https://github.com/demikernel/inetstack/issues/124, right?

BrianZill:
> @BrianZill and @anandbonde this is related to [demikernel/inetstack#124](https://github.com/demikernel/inetstack/issues/124), right?

Somewhat.  It's really about cleaning up properly after the removal of the Ipv4Endpoint type rather than having a polymorphic socket address.  But they're related since we replaced the Ipv4Endpoint type with SocketAddrV4s for the most part.

Your recent PR [demikernel/demikernel#145] fixed an inefficiency where we were doing an extra pair of byte swaps.  I don't know for sure that that particular case came about because of the removal of the Ipv4Endpoint type, but it seems likely.

BrianZill:
FYI, the Rust 1.64.0 release notes contain this little gem:

[...]

PR #735, opened 2023-05-09
[pdpix] Fix Network Types to be Compatible With Each Other

ppenna:
## Description

This PR closes https://github.com/demikernel/demikernel/issues/134.

# Summary of Changes

- Introduced utility function `socketaddrv4_to_sockaddr()`
- Introduced utility function `sockaddr_to_socketaddrv4()`
- Changed concerned code to only perform conversions from/to `SocketAddrV4` and `libc::sockaddr`


iyzhang (approved):
LGTM. Thanks for doing this!

Issue #728, opened 2023-05-04
[doc] Have a Drawing that Illustrates the State Diagram of a Socket


ppenna:
## Context

We should have a drawing that illustrates the state diagram of a socket.

In that diagram, nodes would represent the possible states of a socket, and transitions would be trigger by PDPIX calls.

PR #736, opened 2023-05-09
[doc] Socket States Diagram

ppenna:
## Description

This PR closes https://github.com/demikernel/demikernel/issues/728


iyzhang (changes requested):
Can you change the socket state from connected to established?

ppenna:
> Can you change the socket state from connected to established?

@iyzhang do you mean fixing the source code?

iyzhang:
> > Can you change the socket state from connected to established?
> 
> @iyzhang do you mean fixing the source code?

Yes, the diagram should match the code, so either change one or the other.

PR #737, opened 2023-05-09
[Runtime] Rename scheduler handle to task handle and other things to more useful things

### LINKED ISSUE #174: [scheduler] Better Semantics and Naming for `SchedulerHandle` Functions

iyzhang:
This PR closes #174. It renames the following variables to be more meaningful:
* SchedulerHandle -> TaskHandle
* id -> task_id
* offset -> index

### DIFF
--- a/src/rust/catnap/mod.rs
+++ b/src/rust/catnap/mod.rs
@@ -57,7 +57,7 @@ use crate::{
         QType,
     },
     scheduler::{
-        SchedulerHandle,
+        TaskHandle,
         Yielder,
         YielderHandle,
     },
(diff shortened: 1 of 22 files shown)


anandbonde (approved):
Thanks for doing this change! It definitely makes this area more readable and the intent is clearer.

src/rust/catnap/mod.rs
```diff
@@ -701,6 +698,7 @@ impl CatnapLibOS {
                 };
             },
         }
+
```
ppenna, line 701: Do we need to touch this line?

ppenna (approved):
LGTM

PR #739, opened 2023-05-11
[bugfix] Poll in Every PDPIX Call

ppenna:
## Description

This PR is a partial fix for https://github.com/demikernel/demikernel/issues/521


iyzhang (approved):
LGTM

PR #740, opened 2023-05-16
[runtime] Explicit drop of Tasks in scheduler 

### LINKED ISSUE #593: [runtime] More explicit removal of tasks from scheduler
## Context

Currently the scheduler will drop tasks when the scheduler handle is dropped. We currently avoid this by taking the key out of the handle before it goes out of scope. This is not a good mechanism for expressing that we no longer want the scheduler to schedule this task.

## Proposed Solution

An explicit remove() function on the scheduler handle when we want the scheduler to drop the task and no longer schedule it. We will need to explicitly call this function on take_result() (or whenever we have decided to no longer run the task).

iyzhang:
Currently, our scheduler uses the drop function on the scheduler handle to remove tasks from the scheduler. This means that any time a scheduler handle goes out of scope, it automatically deschedules the task. This PR closes #593 and adds a deschedule function to the task handle.


anandbonde (changes requested):

src/rust/catcollar/mod.rs
```diff
@@ -521,7 +521,12 @@ impl CatcollarLibOS {
 
     /// Takes out the operation result descriptor associated with the target scheduler handle.
     fn take_result(&mut self, handle: TaskHandle) -> (QDesc, OperationResult) {
-        let task: OperationTask = OperationTask::from(self.runtime.scheduler.remove(handle).as_any());
+        let task: OperationTask = if let Some(task) = self.runtime.scheduler.remove(&handle) {
+            OperationTask::from(task.as_any())
+        } else {
+            panic!("Taking the result more than once is not supported");
```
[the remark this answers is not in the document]
iyzhang: Done

src/rust/scheduler/handle.rs
```diff
@@ -52,33 +52,28 @@ pub struct YielderHandle {
...
+    /// Returns the task_id stored in the target [SchedulerHandle].
+    pub fn get_task_id(&self) -> u64 {
+        self.task_id
     }
 
-    /// Returns the raw key stored in the target [SchedulerHandle].
-    pub fn get_task_id(mut self) -> u64 {
-        self.index.take();
-        self.task_id.take().unwrap()
+    /// Removes the task from the scheduler and keeps it from running again.
```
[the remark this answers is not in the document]
iyzhang: I added some tests

src/rust/scheduler/handle.rs
```diff
@@ -52,33 +52,28 @@ pub struct YielderHandle {
...
+    pub fn has_completed(&self) -> bool {
+        let subpage_ix: usize = self.index & (WAKER_BIT_LENGTH - 1);
+        self.page.has_completed(subpage_ix)
     }
 
-    /// Queries whether or not the future associated with the target [SchedulerHandle] has completed.
-    pub fn has_completed(&self) -> bool {
-        let subpage_ix: usize = self.index.unwrap() as usize & (WAKER_BIT_LENGTH - 1);
-        self.chunk.has_completed(subpage_ix)
+    /// Returns the task_id stored in the target [SchedulerHandle].
```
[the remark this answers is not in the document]
iyzhang: done

src/rust/scheduler/scheduler.rs
```diff
@@ -199,13 +198,20 @@ impl Scheduler {
...
+
                 // Handle dropped tasks only.
                 for subpage_ix in BitIter::from(dropped) {
-                    if subpage_ix != 0 {
-                        let ix: usize = (page_ix << WAKER_BIT_LENGTH_SHIFT) + subpage_ix;
-                        tasks.remove(ix);
-                        pages[page_ix].clear(subpage_ix);
-                    }
+                    let index: usize = (page_ix << WAKER_BIT_LENGTH_SHIFT) + subpage_ix;
+                    println!("dropping task {:?}", index);
```
ppenna, line 206: If this is intentional, could you change it to a `trace!()` statement?

src/rust/scheduler/scheduler.rs
```diff
@@ -199,13 +198,20 @@ impl Scheduler {
...
+                    let len: usize = task_ids.len();
+                    // Remove the task id that points to this offset.
+                    // We could build a backmap, but we do not expect drop to happen very often.
+                    task_ids.retain(|_, v| *v != index);
+                    // If there is more than one task id pointing at the offset, something has gone very wrong.
+                    assert!(task_ids.len() == len - 1);
```
ppenna, line 212: Could you add a message to this `assert!()` statement? If we reach this bad condition, we can debug without having to open the source code and reading out the comment above.

src/rust/catnap/mod.rs
```diff
@@ -735,6 +740,7 @@ impl CatnapLibOS {
                 };
             },
         }
+
```
ppenna, line 743: Could you rollback this change?

src/rust/inetstack/mod.rs
```diff
@@ -636,7 +631,11 @@ impl<const N: usize> InetStack<N> {
     ///
     /// This function will panic if the specified future had not completed or is _background_ future.
     pub fn take_operation(&mut self, handle: TaskHandle) -> (QDesc, OperationResult) {
-        let task: OperationTask = OperationTask::from(self.scheduler.remove(handle).as_any());
+        let task: OperationTask = if let Some(task) = self.scheduler.remove(&handle) {
+            OperationTask::from(task.as_any())
+        } else {
+            panic!("Removing task that does not exist (either was previously removed or never inserted)");
```
ppenna, line 637: Shouldn't this condition be `unreachable!()`

src/rust/catnap/mod.rs
```diff
@@ -722,9 +722,14 @@ impl CatnapLibOS {
...
     }
 
-    /// Takes out the result from the [OperationTask] associated with the target [SchedulerHandle].
+    /// Takes out the result from the [OperationTask] associated with the target [TaskHandle].
     fn take_result(&mut self, handle: TaskHandle) -> (QDesc, OperationResult) {
-        let task: OperationTask = OperationTask::from(self.runtime.scheduler.remove(handle.clone()).as_any());
+        let task: OperationTask = if let Some(task) = self.runtime.scheduler.remove(&handle) {
+            OperationTask::from(task.as_any())
+        } else {
+            panic!("Removing task that does not exist (either was previously removed or never inserted)");
```
ppenna, line 730: Shouldn't this condition be `unreachable!()`? Even better, wouldn't be the case to change the function signature to return a `Fail` and we propagate it back to the PDPIX call?

src/rust/catmem/mod.rs
```diff
@@ -333,9 +333,14 @@ impl CatmemLibOS {
...
     }
 
-    /// Takes out the [OperationResult] associated with the target [SchedulerHandle].
+    /// Takes out the [OperationResult] associated with the target [TaskHandle].
     fn take_result(&mut self, handle: TaskHandle) -> (QDesc, OperationResult) {
-        let task: OperationTask = OperationTask::from(self.scheduler.remove(handle).as_any());
+        let task: OperationTask = if let Some(task) = self.scheduler.remove(&handle) {
+            OperationTask::from(task.as_any())
+        } else {
+            panic!("Removing task that does not exist (either was previously removed or never inserted)");
```
ppenna, line 341: Shouldn't this condition be `unreachable!()`? Even better, wouldn't be the case to change the function signature to return a `Fail` and we propagate it back to the PDPIX call?

src/rust/catcollar/mod.rs
```diff
@@ -521,7 +521,12 @@ impl CatcollarLibOS {
 
     /// Takes out the operation result descriptor associated with the target scheduler handle.
     fn take_result(&mut self, handle: TaskHandle) -> (QDesc, OperationResult) {
-        let task: OperationTask = OperationTask::from(self.runtime.scheduler.remove(handle).as_any());
+        let task: OperationTask = if let Some(task) = self.runtime.scheduler.remove(&handle) {
+            OperationTask::from(task.as_any())
+        } else {
+            panic!("Removing task that does not exist (either was previously removed or never inserted)");
```
ppenna, line 527: Shouldn't this condition be `unreachable!()`? Even better, wouldn't be the case to change the function signature to return a `Fail` and we propagate it back to the PDPIX call?

src/rust/catloop/mod.rs
```diff
@@ -568,9 +568,13 @@ impl CatloopLibOS {
...
     }
 
-    /// Takes out the [OperationResult] associated with the target [SchedulerHandle].
+    /// Takes out the [OperationResult] associated with the target [TaskHandle].
     fn take_result(&mut self, handle: TaskHandle) -> (QDesc, OperationResult) {
-        let task: OperationTask = OperationTask::from(self.scheduler.remove(handle).as_any());
+        let task: OperationTask = if let Some(task) = self.scheduler.remove(&handle) {
+            OperationTask::from(task.as_any())
+        } else {
+            panic!("Removing task that does not exist (either was previously removed or never inserted)");
```
ppenna, line 576: Shouldn't this condition be `unreachable!()`? Even better, wouldn't be the case to change the function signature to return a `Fail` and we propagate it back to the PDPIX call?

ppenna (approved):

ppenna (changes requested):
This change is consistently making system-level tests to fail. It may be a side-effect, but before merging it is better to investigate.

src/rust/inetstack/mod.rs
iyzhang: The scheduler doesn't have a dependency on the Demikernel fail errors. I certainly wouldn't say that this is unreachable, as we could write code to either remove the task from the scheduler more than once or accidentally call take_result in two places.

src/rust/scheduler/scheduler.rs
iyzhang: sorry, that shouldn't be there.

src/rust/inetstack/mod.rs
ppenna: Oh, I meant throwing the failure in the line that you are panicking. Something like:

```rust
    pub fn take_operation(&mut self, handle: TaskHandle) -> Result<(QDesc, OperationResult)> {
       if let Some(task) = self.scheduler.remove(&handle) {
             let task: OperationTask = OperationTask::from(task.as_any());
             Ok(task.get_result().expect("Coroutine not finished"))
        } else {
            Err(Fail::new(libc::EINVAL, "invalid co-routine")),
        }
    }
````

iyzhang: Well, this isn't really something that we would pass back to the user because if either the coroutine didn't finish or we can't find it, then something has gone pretty wrong because we just checked the handle and it said complete. So I think panicking (and expecting) is probably the right thing here. I don't like failing with something the user won't understand. 

Either way, lets table this for now because this PR has gotten quite large.

ppenna (approved):

PR #741, opened 2023-05-16
[tcp-ping-pong] Refactor to add Drop

### LINKED ISSUE #719: [tests] Refactor `tcp-ping-pong` to Follow "Associated Functions" Design
## Context

The `tcp-ping-pong` application was written as a collection of standalone functions. 

We should refactor the `tcp-ping-pong` application to follow an "associated functions" code structure, so that we can have correct trait implementations, such as `Drop`.

anandbonde:
Closes #719

### DIFF
--- a/examples/rust/tcp-ping-pong.rs
+++ b/examples/rust/tcp-ping-pong.rs
@@ -43,6 +43,9 @@ use ::demikernel::perftools::profiler;
 
 const BUFFER_SIZE: usize = 64;
 
+/// Number of rounds to execute.
+const NROUNDS: usize = 1024;
+
 //======================================================================================================================
 // mksga()
 //======================================================================================================================
@@ -186,10 +189,6 @@ fn pop_and_wait(libos: &mut LibOS, sockqd: QDesc, recvbuf: &mut [u8]) -> Result<
     Ok(())
 }
 
-//======================================================================================================================
-// close()
-//======================================================================================================================
-
 /// Closes a socket and warns if not successful.
 fn close(libos: &mut LibOS, sockqd: QDesc) {
     if let Err(e) = libos.close(sockqd) {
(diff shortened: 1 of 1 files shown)


examples/rust/tcp-ping-pong.rs
```diff
@@ -198,178 +197,174 @@ fn close(libos: &mut LibOS, sockqd: QDesc) {
...
-                anyhow::bail!("push and wait failed: {:?}", e);
-            }
-            if let Err(e) = libos.sgafree(sga) {
-                // If error, free socket.
-                close(&mut libos, sockqd);
-                anyhow::bail!("failed to release scatter-gather array: {:?}", e)
-            }
+    fn run(&mut self, remote: SocketAddrV4, nrounds: usize) -> Result<()> {
+        if let Err(e) = connect_and_wait(&mut self.libos, self.sockqd, remote) {
+            close(&mut self.libos, self.sockqd);
```
iyzhang, line 315: This close is not necessary, I think.

examples/rust/tcp-ping-pong.rs
```diff
@@ -198,178 +197,174 @@ fn close(libos: &mut LibOS, sockqd: QDesc) {
...
 
-            for x in &recvbuf {
-                if *x != fill_char {
-                    close(&mut libos, sockqd);
-                    anyhow::bail!("fill check failed: expected={:?} received={:?}", fill_char, *x);
+            // Pop data, and sanity check it.
+            {
+                let mut recvbuf: [u8; BUFFER_SIZE] = [0; BUFFER_SIZE];
+                if let Err(e) = pop_and_wait(&mut self.libos, accepted_qd, &mut recvbuf) {
+                    close(&mut self.libos, accepted_qd);
```
iyzhang, line 246: You could add `accepted_qd: Option<QDesc>` to the server data structure and then close it in the drop if there is qd. That way we won't have to specifically call close each time here and on line 252.

iyzhang (changes requested):

examples/rust/tcp-ping-pong.rs
anandbonde: Done.

examples/rust/tcp-ping-pong.rs
anandbonde: Done.

iyzhang (approved):

PR #744, opened 2023-05-20
[udp-push-pop] Refactor to add Drop

### LINKED ISSUE #716: [tests] Refactor `udp-push-pop` to Follow "Associated Functions" Design
## Context

The `udp-push-pop` application was written as a collection of standalone functions. 

We should refactor the `udp-push-pop` application to follow an "associated functions" code structure, so that we can have correct trait implementations, such as `Drop`.

anandbonde:
Closes #716

### DIFF
--- a/examples/rust/udp-push-pop.rs
+++ b/examples/rust/udp-push-pop.rs
@@ -20,6 +20,10 @@ use ::std::{
     slice,
     str::FromStr,
 };
+use log::{
+    error,
+    warn,
+};
 
 #[cfg(target_os = "windows")]
 pub const AF_INET: i32 = windows::Win32::Networking::WinSock::AF_INET.0 as i32;
@@ -42,6 +46,7 @@ use ::demikernel::perftools::profiler;
 
 const BUFFER_SIZE: usize = 64;
 const FILL_CHAR: u8 = 0x65;
+const NSENDS: usize = 1024;
 
 //======================================================================================================================
 // mksga()
@@ -83,8 +88,8 @@ fn mksga(libos: &mut LibOS, size: usize, value: u8) -> Result<demi_sgarray_t> {
 /// Free scatter-gather array and warn on error.
 fn freesga(libos: &mut LibOS, sga: demi_sgarray_t) {
     if let Err(e) = libos.sgafree(sga) {
-        println!("ERROR: sgafree() failed (error={:?})", e);
-        println!("WARN: leaking sga");
+        error!("sgafree() failed (error={:?})", e);
+        warn!("leaking sga");
     }
 }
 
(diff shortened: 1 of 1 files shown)


examples/rust/udp-push-pop.rs
```diff
@@ -100,168 +101,172 @@ fn close(libos: &mut LibOS, sockqd: QDesc) {
...
-                // If error, close socket.
-                close(&mut libos, sockqd);
-                anyhow::bail!("unexpected result")
-            },
-            Err(e) => {
-                // If error, close socket.
-                close(&mut libos, sockqd);
-                anyhow::bail!("operation failed: {:?}", e)
-            },
+        match self.libos.bind(self.sockqd, local) {
```
iyzhang, line 132: We should be consistent about whether we do bind in the constructor or in the run() function. I thought we had previously done it in the constructor and I think that is fine.

examples/rust/udp-push-pop.rs
```diff
@@ -100,168 +101,172 @@ fn close(libos: &mut LibOS, sockqd: QDesc) {
...
+                Ok(_) => anyhow::bail!("unexpected result"),
+                Err(e) => anyhow::bail!("operation failed: {:?}", e),
+            };
+
+            // Sanity check received data.
+            let ptr: *mut u8 = self.sga.expect("should be a valid sgarray").sga_segs[0].sgaseg_buf as *mut u8;
+            let len: usize = self.sga.expect("should be a valid sgarray").sga_segs[0].sgaseg_len as usize;
+            let slice: &mut [u8] = unsafe { slice::from_raw_parts_mut(ptr, len) };
+            for x in slice {
+                if *x != fill_char {
```
iyzhang, line 155: Can you change this to a demikernel::ensure_eq()?

examples/rust/udp-push-pop.rs
```diff
@@ -100,168 +101,172 @@ fn close(libos: &mut LibOS, sockqd: QDesc) {
...
+        Ok(())
+    }
+}
 
-    // TODO: close socket when we get close working properly in catnip.
-    Ok(())
+// The Drop implementation for the UDP client.
+impl Drop for UdpClient {
+    fn drop(&mut self) {
+        close(&mut self.libos, self.sockqd);
```
iyzhang, line 265: Is having a close() function necessary? I think this is the only place that we use it.

iyzhang (approved):
LGTM, thanks for cleaning this up!

examples/rust/udp-push-pop.rs
anandbonde: We discussed offline: we rely on the fact that the drop() function is not called if the constructor (new()) fails. So, if we bind() in new(), there's a chance that it fails and the socket may not be closed because drop() was not invoked.

examples/rust/udp-push-pop.rs
anandbonde: This function is used by both UdpServer and UdpClient. So I left the existing `close()` function as is.

examples/rust/udp-push-pop.rs
```diff
@@ -95,173 +100,175 @@ fn freesga(libos: &mut LibOS, sga: demi_sgarray_t) {
...
             }
-        }
 
-        if let Err(e) = libos.sgafree(sga) {
-            // If error, close socket.
-            close(&mut libos, sockqd);
-            anyhow::bail!("failed to release scatter-gather array: {:?}", e)
+            match self.libos.sgafree(self.sga.expect("should be a valid sgarray")) {
+                Ok(_) => self.sga = None,
+                Err(e) => anyhow::bail!("failed to release scatter-gather array: {:?}", e),
```
ppenna, line 164: You may avoid these calls to `except()`, but enclosing this in a `if let Some() { .. }` statement.

examples/rust/udp-push-pop.rs
```diff
@@ -95,173 +100,175 @@ fn freesga(libos: &mut LibOS, sga: demi_sgarray_t) {
...
+                Err(e) => anyhow::bail!("operation failed: {:?}", e),
+                _ => anyhow::bail!("unexpected result"),
+            };
+
+            match self.libos.sgafree(self.sga.expect("should be a valid sgarray")) {
+                Ok(_) => self.sga = None,
+                Err(e) => anyhow::bail!("failed to release scatter-gather array: {:?}", e),
+            }
+
+            println!("push ({:?})", i);
```
ppenna, line 253: You may avoid these calls to `except()`, but enclosing this in a `if let Some() { .. }` statement.

ppenna (approved):
LGTM

examples/rust/udp-push-pop.rs
anandbonde: Done.

PR #745, opened 2023-05-20
[tcp-push-pop] Refactor to add Drop

### LINKED ISSUE #718: [tests] Refactor `tcp-push-pop` to Follow "Associated Functions" Design
## Context

The `tcp-push-pop` application was written as a collection of standalone functions. 

We should refactor the `tcp-push-pop` application to follow an "associated functions" code structure, so that we can have correct trait implementations, such as `Drop`.

anandbonde:
Closes #718

### DIFF
--- a/examples/rust/tcp-push-pop.rs
+++ b/examples/rust/tcp-push-pop.rs
@@ -20,6 +20,10 @@ use ::std::{
     slice,
     str::FromStr,
 };
+use log::{
+    error,
+    warn,
+};
 
 #[cfg(target_os = "windows")]
 pub const AF_INET: i32 = windows::Win32::Networking::WinSock::AF_INET.0 as i32;
@@ -83,8 +87,8 @@ fn mksga(libos: &mut LibOS, size: usize, value: u8) -> Result<demi_sgarray_t> {
 /// Free scatter-gather array and warn on error.
 fn freesga(libos: &mut LibOS, sga: demi_sgarray_t) {
     if let Err(e) = libos.sgafree(sga) {
-        println!("ERROR: sgafree() failed (error={:?})", e);
-        println!("WARN: leaking sga");
+        error!("sgafree() failed (error={:?})", e);
+        warn!("leaking sga");
     }
 }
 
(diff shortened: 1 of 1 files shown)


examples/rust/tcp-push-pop.rs
```diff
@@ -100,218 +100,209 @@ fn close(libos: &mut LibOS, sockqd: QDesc) {
...
+            sockqd,
+            accepted_qd: None,
+            sga: None,
+        });
+    }
+
+    pub fn run(&mut self, local: SocketAddrV4, fill_char: u8, buffer_size: usize) -> Result<()> {
+        let nbytes: usize = buffer_size * 1024;
+
+        if let Err(e) = self.libos.bind(self.sockqd, local) {
```
iyzhang, line 135: Again, I think it would be fine to do this in the constructor. Then if the bind or listen fail, we don't also run drop()

examples/rust/tcp-push-pop.rs
```diff
@@ -100,218 +100,209 @@ fn close(libos: &mut LibOS, sockqd: QDesc) {
...
+                Ok(_) => anyhow::bail!("unexpected result"),
+            };
+
+            // Sanity check received data.
+            let ptr: *mut u8 = self.sga.expect("should be a valid sgarray").sga_segs[0].sgaseg_buf as *mut u8;
+            let len: usize = self.sga.expect("should be a valid sgarray").sga_segs[0].sgaseg_len as usize;
+            let slice: &mut [u8] = unsafe { slice::from_raw_parts_mut(ptr, len) };
+
+            for x in slice {
+                if *x != fill_char {
```
iyzhang, line 180: Can you change this to a demikernel::ensure_eq()?

iyzhang (approved):
LGTM with small nits

examples/rust/tcp-push-pop.rs
anandbonde: We discussed offline: we rely on the fact that the `drop()` function is not called if the constructor (`new()`) fails. So, if we `bind()` in `new()`, there's a chance that it fails and the socket may not be closed because `drop()` was not invoked.

examples/rust/tcp-push-pop.rs
anandbonde: Done.

examples/rust/tcp-push-pop.rs
```diff
@@ -95,223 +99,212 @@ fn freesga(libos: &mut LibOS, sga: demi_sgarray_t) {
...
-    };
-    let mut libos: LibOS = match LibOS::new(libos_name) {
-        Ok(libos) => libos,
-        Err(e) => anyhow::bail!("failed to initialize libos: {:?}", e.cause),
-    };
-    let nbytes: usize = BUFFER_SIZE * 1024;
+// Implementation of the TCP server.
+impl TcpServer {
+    pub fn new(mut libos: LibOS) -> Result<Self> {
+        // Create the localsocket.
```
ppenna, line 122: "local socket"

anandbonde: Done.

PR #746, opened 2023-05-20
[tcp-ping-pong] Additional refactors

anandbonde:
Moved the sga inside structs to avoid several calls to free scattered in the code.


examples/rust/tcp-ping-pong.rs
```diff
@@ -334,13 +350,18 @@ impl TcpClient {
...
-                let sga: demi_sgarray_t = mksga(&mut self.libos, BUFFER_SIZE, fill_char)?;
-                if let Err(e) = push_and_wait(&mut self.libos, self.sockqd, &sga) {
-                    freesga(&mut self.libos, sga);
+                self.sga = Some(mksga(&mut self.libos, BUFFER_SIZE, fill_char)?);
+                if let Err(e) = push_and_wait(
+                    &mut self.libos,
+                    self.sockqd,
+                    &self.sga.expect("should be a valid sgarray"),
+                ) {
+                    freesga(&mut self.libos, self.sga.expect("should be a valid sgarray"));
```
iyzhang, line 359: This can be removed and no expect is required then. The drop() function can clean it up like the sockets.

examples/rust/tcp-ping-pong.rs
```diff
@@ -334,13 +350,18 @@ impl TcpClient {
...
                     anyhow::bail!("push and wait failed: {:?}", e);
                 }
-                if let Err(e) = self.libos.sgafree(sga) {
-                    anyhow::bail!("failed to release scatter-gather array: {:?}", e)
+                match self.libos.sgafree(self.sga.expect("should be a valid sgarray")) {
```
iyzhang, line 362: You can use the take function here, eg go back to the if statement but use take() to set the Option to None.

iyzhang (changes requested):
Lets revisit this tomorrow per our discussion.

examples/rust/tcp-ping-pong.rs
anandbonde: Done.

examples/rust/tcp-ping-pong.rs
anandbonde: Done.

iyzhang (approved):
LGTM

PR #747, opened 2023-05-24
[catloop] Warn msg correction

anandbonde:
The message was misleading us to think that the client was invoking `accept()` in the logs.


iyzhang (approved):
Thanks for catching this

ppenna (approved):

Issue #748, opened 2023-05-30
[test] Concurrent Unit Tests in Catloop and Catmem


ppenna:
## Description

Our `demikernel_ci` utility launches unit tests on server and client machines.

For Catmem and Catloop LibOses, this means that we run two instances of unit-tests in the same machine, thus causing them to conflict and interfere in one another.

## Code Snippet

https://github.com/demikernel/demikernel/blob/adcae5ed3fe87879fda324bc4dc3a46794391914/tools/demikernel_ci.py#L142-L150

## Expected Behavior

For Catmem and Catloop LibOSes we should run unit tests in either the server or client endpoint.

PR #749, opened 2023-05-30
[test] Don't Run Unit Tests Concurrently

ppenna:
## Description

This PR closes https://github.com/demikernel/demikernel/issues/748


iyzhang (approved):
Looks fine to me but we probably don't need to run the unit tests twice for any of the other libOSes either, so I'm fine with just removing line 149.

ppenna:
> Looks fine to me but we probably don't need to run the unit tests twice for any of the other libOSes either, so I'm fine with just removing line 149.

Sure, I'll drop it as well.

Issue #731, opened 2023-05-08
[pdpix] Use Wrapped Network Types


ppenna:
## Context

We should drop usage of `libc` network types  in favor of wrapped types, as the latter types work on Linux and Windows.:

- [ ] Replace `libc::sockaddr` for `demikernel::pal::SockAddr`
- [ ] Replace `libc::sockaddr_in` for `demikernel::pal::SockAddrIn`
- [ ] Replace `libc::socklen_t` for `demikernel::pal::Socklen`

PR #755, opened 2023-05-31
[pdpix] Use Wrapped Network Types

ppenna:
## Description

This PR closes https://github.com/demikernel/demikernel/issues/731.


src/rust/catcollar/mod.rs
```diff
@@ -185,14 +192,8 @@ impl CatcollarLibOS {
...
-        let saddr: libc::sockaddr = linux::socketaddrv4_to_sockaddr(&local);
-        match unsafe {
-            libc::bind(
-                fd,
-                &saddr as *const libc::sockaddr,
-                mem::size_of::<libc::sockaddr_in>() as libc::socklen_t,
-            )
-        } {
+        let saddr: SockAddr = linux::socketaddrv4_to_sockaddr(&local);
+        match unsafe { libc::bind(fd, &saddr as *const SockAddr, mem::size_of::<SockAddrIn>() as Socklen) } {
```
iyzhang, line 196: I assume this mem::sizeof function still works the same for wrapped types?

iyzhang (approved):
LGTM, thanks for taking care of this

src/rust/catcollar/mod.rs
ppenna: Yes, these wrapped types are just aliases to the network types defined in the `windows` and `libc` crates.

Issue #760, opened 2023-06-01
[test] Closing Server-Side Passive Socket Twice in `tcp-close`


ppenna:
## Description

We are closing the server-side passive socket twice in `tcp-close`.

This socket is closed at the end of the `run()` and `run_close_sockets_on_accept()` functions, as well as in the `drop()` function.

## Code Snippets

https://github.com/demikernel/demikernel/blob/28cb91fcbca35f189db6fb552b0528047445bb15/examples/tcp-close/server.rs#L175

https://github.com/demikernel/demikernel/blob/28cb91fcbca35f189db6fb552b0528047445bb15/examples/tcp-close/server.rs#L219

https://github.com/demikernel/demikernel/blob/28cb91fcbca35f189db6fb552b0528047445bb15/examples/tcp-close/server.rs#L341-L344

## Expected Behavior

Since the constructor has created this object, the destructor should be responsible to release it.

We should close this socket only in the `drop()` function.

PR #761, opened 2023-06-01
[test] Don't Close Server-Side Passive Socket Twice in `tcp-close`

ppenna:
## Description

This PR closes https://github.com/demikernel/demikernel/issues/760


iyzhang (commented):
@anandbonde and I discussed this and we agreed that using Drop to close the socket should only happen on the error path. We have this double close issue in several places, not just this test, so you should probably talk to Anand about it.

iyzhang (approved):
LGTM

PR #766, opened 2023-06-07
[cleanup] Remove unnecessary code

anandbonde:
1. Removed unnecessary call to `get_page_indices()` in the scheduler.
2. Removed unnecessary `mut` keywords from: a) inetstack, b) runtime. These were causing compiler warnings.


iyzhang (approved):
Looks good, thanks for the cleanup

Issue #768, opened 2023-06-07
[catloop] Leaking Tasks on Connection Establishment Protocol


ppenna:
## Description

Due to the buggy drop semantics on the scheduler, we are leaking tasks on the connection establishment protocol, in Catloop LibOs.

## Code Snippets

https://github.com/demikernel/demikernel/blob/7863a6237df2e4cbecad9a233a731c8e93fa94b8/src/rust/catloop/futures/accept.rs#L127-L128

https://github.com/demikernel/demikernel/blob/7863a6237df2e4cbecad9a233a731c8e93fa94b8/src/rust/catloop/futures/accept.rs#L150

https://github.com/demikernel/demikernel/blob/7863a6237df2e4cbecad9a233a731c8e93fa94b8/src/rust/catloop/futures/accept.rs#L192-L194

https://github.com/demikernel/demikernel/blob/7863a6237df2e4cbecad9a233a731c8e93fa94b8/src/rust/catloop/futures/accept.rs#L234

https://github.com/demikernel/demikernel/blob/7863a6237df2e4cbecad9a233a731c8e93fa94b8/src/rust/catloop/futures/connect.rs#L130

https://github.com/demikernel/demikernel/blob/7863a6237df2e4cbecad9a233a731c8e93fa94b8/src/rust/catloop/futures/connect.rs#L168

https://github.com/demikernel/demikernel/blob/7863a6237df2e4cbecad9a233a731c8e93fa94b8/src/rust/catloop/futures/connect.rs#L189

PR #770, opened 2023-06-07
[catloop] Leaking Tasks on Connection Establishment Protocol

ppenna:
## Description

This PR closes https://github.com/demikernel/demikernel/issues/768

### DIFF
--- a/src/rust/catloop/futures/accept.rs
+++ b/src/rust/catloop/futures/accept.rs
@@ -125,9 +125,13 @@ impl Future for AcceptFuture {
                 duplex_pipe,
             } => {
                 if let Some(handle) = DuplexPipe::poll(&self_.catmem, *qt_close)? {
-                    check_connect_request(&self_.catmem, handle, *qt_close).expect("magic connect");
-                    debug!("connection accepted!");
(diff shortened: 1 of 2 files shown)


iyzhang (approved):
LGTM. Thanks for fixing this!

src/rust/catloop/futures/accept.rs
```diff
@@ -125,9 +125,13 @@ impl Future for AcceptFuture {
                 duplex_pipe,
             } => {
                 if let Some(handle) = DuplexPipe::poll(&self_.catmem, *qt_close)? {
-                    check_connect_request(&self_.catmem, handle, *qt_close).expect("magic connect");
-                    debug!("connection accepted!");
-                    return Poll::Ready(Ok((*remote, duplex_pipe.clone())));
+                    match check_connect_request(&self_.catmem, handle, *qt_close) {
+                        Ok(_) => {
```
iyzhang, line 129: If the server does not get a magic number back from the new pipe, it should probably close the pipe and clean it up, not continue to use it.

iyzhang (changes requested):

iyzhang (approved):

PR #779, opened 2023-06-19
[inetstack] Use TCP PSH bit

joshuafried:
Some TCP stacks may buffer received data until a segment is received with the PSH bit set. This PR modifies Demikernel's TCP stack to set the PSH bit when transmitting data to avoid high latencies when communicating with such TCP stacks. The PSH bit is only set on the final segment transmitted for each sent buffer.


ppenna (approved):
LGTM

PR #782, opened 2023-06-20
[runtime] Fixing issues in scheduler

iyzhang:
This PR fixes a rebase issue that overwrote a bug fix from PR #764 and it removes an infinite loop in the task allocation code.

### DIFF
--- a/src/rust/scheduler/scheduler.rs
+++ b/src/rust/scheduler/scheduler.rs
@@ -56,6 +56,8 @@ use ::std::{
 /// This value was chosen arbitrarily.
 #[cfg(debug_assertions)]
 const SCHEDULER_SEED: u64 = 42;
+const MAX_NUM_TASKS: usize = 16000;
+const MAX_RETRIES_TASK_ID_ALLOC: usize = 500;
 
 //======================================================================================================================
 // Structures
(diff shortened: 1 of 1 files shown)


src/rust/scheduler/scheduler.rs
```diff
@@ -136,12 +136,20 @@ impl Scheduler {
...
         // Generate a new id. If the id is currently in use, keep generating until we find an unused id.
         let mut task_ids: RefMut<HashMap<u64, usize>> = self.task_ids.borrow_mut();
-        let task_id: u64 = loop {
-            let id: u64 = id_gen.next_u64() as u16 as u64;
-            if !task_ids.contains_key(&id) {
-                task_ids.insert(id, index);
-                break id;
+        // If the address space for task ids is close to half full, it will become increasingly difficult to avoid
+        // collisions, so we cap the number of tasks at 16,000.
+        if task_ids.len() > 16000 {
```
ppenna, line 141: Could we have a constant for this?

src/rust/scheduler/scheduler.rs
```diff
@@ -136,12 +136,20 @@ impl Scheduler {
...
+            panic!("Too many concurrent tasks");
+        }
+        let task_id: u64 = 'get_id: {
+            for _ in 0..100 {
```
ppenna, line 145: Could we have a constant for this?

Also, should this number be higher?

ppenna (commented):
LGTM

ppenna (approved):

src/rust/scheduler/scheduler.rs
iyzhang: done

src/rust/scheduler/scheduler.rs
iyzhang: done

PR #784, opened 2023-06-21
[test] Missing cases on IPv4 test-unit

### LINKED ISSUE #781: [test] Missing cases on IPv4 test-unit
## Description

Currently, on the [test_ipv4_header_parse_good](https://github.com/microsoft/demikernel/blob/dev/src/rust/inetstack/protocols/ipv4/tests.rs#L86) test, the loop on the `ihl` variable varies from 0 (inclusive) to 15 (exclusive), this means that the value 15 is never used, and it is the maximum value that the IHL field on IPv4 supports.

Also, on the [build_ipv4_header](https://github.com/microsoft/demikernel/blob/dev/src/rust/inetstack/protocols/ipv4/tests.rs#L29) function, the `compute_checksum` function is calculating the IPv4 header checksum considering only the 20 bytes of the header, regardless of the `ihl` variable which defines the header length.

This last one also impacts the [test_ipv4_header_parse_invalid_ihl](https://github.com/microsoft/demikernel/blob/dev/src/rust/inetstack/protocols/ipv4/tests.rs#L184) test because it creates IPv4 headers with an invalid IHL field, and the `build_ipv4_header` function calculates the checksum passing 20 bytes as the header.

carvalhof:
## Description

This PR closes #781.

## Summary of Changes

- Updated the upper bound of the for in the test_ipv4_header_parse_good test to include the value 15.
- Added the header length in the build_ipv4_header function.
- Verified the minimum size of the buffer in the compute_checksum function.

### DIFF
--- a/src/rust/inetstack/protocols/ipv4/datagram.rs
+++ b/src/rust/inetstack/protocols/ipv4/datagram.rs
@@ -308,6 +308,14 @@ impl Ipv4Header {
     /// Computes the checksum of the target IPv4 header.
     pub fn compute_checksum(buf: &[u8]) -> u16 {
         let mut state: u32 = 0xffff;
+
+        // Do not compute checksum if buffer is too small.
+        if buf.len() < IPV4_HEADER_MIN_SIZE as usize {
+            // This should not happen by construction. If it does, log it.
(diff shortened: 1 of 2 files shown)


ppenna (approved):
LGTM

PR #785, opened 2023-06-22
[catloop] Move to async functions and add yield and cancel on close

### LINKED ISSUE #773: [catloop] Remove duplicate calls to catmem poll
## Context

Currently the Catmem scheduler is polled from the Catloop scheduler. However, it is currently done from three places: the poll function in Catloop, the Accept future and the Connect future. In the two futures, we use a borrow_mut, while we use a borrow in the Catloop poll function.

## Proposed Solution

We should invoke the catmem scheduler only once on each call to Catloop poll.

### LINKED ISSUE #774: [catloop] Move accept to async coroutine
## Context

Currently the accept functionality is inside a future.

## Proposed Solution

We should move it to an async function in the coroutine pattern

### LINKED ISSUE #775: [catloop] Move connect to async coroutine
## Context

Currently, the connect functionality is in a future.

## Proposed Solution

It should be in an async function in the coroutine pattern.

iyzhang:
This PR covers the following issues and enhancements:
1. Close #774 
2. Close #775 
3. Close #773

### DIFF
--- a/src/rust/catloop/futures/accept.rs
+++ b/src/rust/catloop/futures/accept.rs
@@ -19,124 +19,76 @@ use crate::{
             demi_qresult_t,
         },
     },
-    scheduler::TaskHandle,
+    scheduler::{
+        TaskHandle,
+        Yielder,
+    },
     QToken,
 };
 use ::std::{
-    cell::RefCell,
-    future::Future,
+    cell::{
+        RefCell,
+        RefMut,
+    },
     mem,
     net::{
         Ipv4Addr,
         SocketAddrV4,
     },
-    pin::Pin,
     rc::Rc,
     slice,
-    task::{
-        Context,
-        Poll,
-    },
 };
 
 //======================================================================================================================
-// Enumerations
+// Coroutine Functions
 //======================================================================================================================
 
-enum ServerState {
-    ListenAndAccept {
-        qt_rx: QToken,
-    },
-    Connect {
-        qt_tx: QToken,
-        duplex_pipe: Rc<DuplexPipe>,
-    },
-    Connected {
-        qt_close: QToken,
-        remote: SocketAddrV4,
-        duplex_pipe: Rc<DuplexPipe>,
-    },
-}
-
-//======================================================================================================================
-// Structures
-//======================================================================================================================
-
-/// Descriptor for accept operation.
-pub struct AcceptFuture {
+pub async fn accept_coroutine(
+    ipv4: &Ipv4Addr,
     catmem: Rc<RefCell<CatmemLibOS>>,
-    /// Local IPv4 address.
-    ipv4: Ipv4Addr,
-    /// Control duplex pipe used for establishing a the connection.
     control_duplex_pipe: Rc<DuplexPipe>,
-    /// Port number new connection.
     new_port: u16,
-    // State in the connection establishment protocol.
-    state: ServerState,
-}
+    yielder: Yielder,
+) -> Result<(SocketAddrV4, Rc<DuplexPipe>), Fail> {
+    loop {
+        // Grab next request from the control duplex pipe.
+        let new_duplex_pipe: Rc<DuplexPipe> = match pop_magic_number(
+            &catmem,
+            control_duplex_pipe.clone(),
+            Some(mem::size_of_val(&CatloopLibOS::MAGIC_CONNECT)),
+            &yielder,
+        )
+        .await
+        {
+            // Recieved a valid magic number so create the new connection. This involves create the new duplex pipe
+            // and sending the port number to the remote.
+            Ok(true) => create_pipe(&catmem, control_duplex_pipe.clone(), ipv4, new_port, &yielder).await?,
+            // Invalid request.
+            Ok(false) => continue,
+            // Some error.
+            Err(e) => return Err(e),
+        };
 
-//======================================================================================================================
-// Associate Functions
-//======================================================================================================================
-
-impl AcceptFuture {
-    /// Creates a descriptor for an accept operation.
-    pub fn new(
-        ipv4: &Ipv4Addr,
-        catmem: Rc<RefCell<CatmemLibOS>>,
-        control_duplex_pipe: Rc<DuplexPipe>,
-        new_port: u16,
-    ) -> Result<Self, Fail> {
-        // Issue first pop. Note that we intentionally issue an unbound
-        // pop() because the connection establishment protocol requires that
-        // only one connection request is accepted at a time.
-        let qt_rx: QToken = control_duplex_pipe.pop(None)?;
-        Ok(AcceptFuture {
-            catmem,
-            ipv4: ipv4.clone(),
-            control_duplex_pipe,
-            new_port,
-            state: ServerState::ListenAndAccept { qt_rx },
-        })
-    }
-}
-
-//======================================================================================================================
-// Trait Implementations
-//======================================================================================================================
-
-impl Future for AcceptFuture {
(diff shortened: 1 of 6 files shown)


src/rust/catloop/mod.rs
```diff
@@ -254,95 +256,102 @@ impl CatloopLibOS {
         let mut qtable: RefMut<IoQueueTable<CatloopQueue>> = self.qtable.borrow_mut();
 
         // Issue accept operation.
-        match qtable.get(&qd) {
+        let (control_duplex_pipe, local): (Rc<DuplexPipe>, SocketAddrV4) = match qtable.get_mut(&qd) {
```
ppenna, line 259: Did you have any motivation for moving out from the previous match construct?

src/rust/catloop/futures/connect.rs
```diff
@@ -46,288 +47,192 @@ use ::std::{
...
+) -> Result<(SocketAddrV4, Rc<DuplexPipe>), Fail> {
+    let ipv4: &Ipv4Addr = remote.ip();
+    let port: u16 = remote.port().into();
+
+    // Open connection to server.
+    let control_duplex_pipe: Rc<DuplexPipe> = Rc::new(DuplexPipe::open_duplex_pipe(catmem.clone(), ipv4, port)?);
+
+    // Gets the port for the new connection from the server by sending a connection request repeatedly until a port
+    // comes back.
+    let new_port: u16 = get_port(catmem.clone(), control_duplex_pipe.clone(), &yielder).await?;
```
ppenna, line 63: If we fail, won't we leave behind a dangling control duplex pipe opened?

src/rust/catloop/futures/connect.rs
```diff
@@ -46,288 +47,192 @@ use ::std::{
...
+    // Shut down control duplex pipe as we can open the new pipe now.
+    control_duplex_pipe.shutdown()?;
+    // Open underlying pipes.
+    let remote: SocketAddrV4 = SocketAddrV4::new(*ipv4, new_port);
+    let new_pipe: Rc<DuplexPipe> = Rc::new(DuplexPipe::open_duplex_pipe(catmem.clone(), ipv4, new_port)?);
+    // Send an ack to the server over the new pipe.
+    send_ack(catmem.clone(), new_pipe.clone(), &yielder).await?;
```
ppenna, line 70: If we fail, won't we leave behind a dangling duplex pipe?

src/rust/catloop/futures/connect.rs
```diff
@@ -46,288 +47,192 @@ use ::std::{
...
-            warn!("connect_request_sent(): qt_rx is None, allocating new qtoken");
-            self_.control_duplex_pipe.pop(Some(size))?
-        };
-
-        // Transition to the next state in the connection establishment protocol.
-        self_.state = ClientState::ConnectAckReceived { qt_rx, attempt: 0 };
+        // If we received a port back from the server, then unpack it. Otherwise, send the connection request again.
+        if handle.has_completed() {
+            // Re-acquire reference to catmem libos.
+            catmem_ = catmem.borrow_mut();
```
ppenna, line 150: What if is not possible to reborrow? This may happend if some concurrent, buggy async function yields without calling drop. Re-acquire a borrow is NOT like re-acquiring a lock. The former will cause a runtime panic, whereas the latter will spin. You should consider using `try_borrow_mut()`.

src/rust/catloop/futures/connect.rs
```diff
@@ -46,288 +47,192 @@ use ::std::{
...
-            duplex_pipe,
-        };
-    } else {
-        if attempt > MAX_ACK_RECEIVED_ATTEMPTS {
-            // Connection timeout, retry.
-            debug!(
-                "connect_ack_received(): connection timeout, retrying (qt_rx={:?})",
-                qt_rx
+    }
+    // Re-acquire reference to catmem libos.
```
ppenna, line 202: What if is not possible to reborrow? This may happend if some concurrent, buggy async function yields without calling drop. Re-acquire a borrow is NOT like re-acquiring a lock. The former will cause a runtime panic, whereas the latter will spin. We should consider using `try_borrow_mut()`.

src/rust/catloop/futures/connect.rs
```diff
@@ -46,288 +47,192 @@ use ::std::{
...
+    catmem_ = catmem.borrow_mut();
```
ppenna, line 203: What if is not possible to reborrow? This may happend if some concurrent, buggy async function yields without calling drop. Re-acquire a borrow is NOT like re-acquiring a lock. The former will cause a runtime panic, whereas the latter will spin. We should consider using `try_borrow_mut()`.

src/rust/catloop/futures/accept.rs
```diff
@@ -19,138 +19,112 @@ use crate::{
...
+    // Drop the mutable reference because we might yield.
+    drop(catmem_);
+    // Yield until pop completes.
+    while !handle.has_completed() {
+        if let Err(e) = yielder.yield_once().await {
+            return Err(e);
+        }
+    }
+    // Re-acquire mutable reference.
+    catmem_ = catmem.borrow_mut();
```
ppenna, line 125: See my previous comments on re-acquiring a borrow.

src/rust/catloop/futures/accept.rs
```diff
@@ -185,113 +159,63 @@ fn check_connect_request(catmem: &Rc<RefCell<CatmemLibOS>>, handle: TaskHandle,
...
+    drop(catmem_);
+
+    // Wait for push to complete.
+    while !handle.has_completed() {
+        if let Err(e) = yielder.yield_once().await {
+            return Err(e);
         }
     }
+    // Re-acquire mutable reference.
+    catmem_ = catmem.borrow_mut();
```
ppenna, line 196: See my previous comments on re-acquiring a borrow.

ppenna (commented):
This is indeed a nice improvement.

I have noted down some concerns that popped out for me. It is likely that previous code had them as well, but it was not that clear...

Let us elaborate on these issues, but overall looks good.

Thanks!

src/rust/catloop/mod.rs
iyzhang: Yes, I need to borrow the queue again to insert the pending op on line 322, and I can't do that within the same scope as line 261.

src/rust/catloop/futures/connect.rs
iyzhang: Yes,  thanks for catching that!

src/rust/catloop/futures/connect.rs
iyzhang: If this happens, we should panic and fix the buggy code. I don't know what situation we should let the buggy code remain and hope that it drops the reference later.

src/rust/catloop/futures/connect.rs
iyzhang: Yup, it's fixed now!

src/rust/catloop/mod.rs
ppenna: Sure.

src/rust/catloop/futures/connect.rs
ppenna: Thanks!

src/rust/catloop/futures/connect.rs
ppenna: Thanks!

src/rust/catloop/futures/connect.rs
ppenna: Agreed.

ppenna (approved):
Thanks for working on this.

PR #791, opened 2023-06-22
[catloop] Missing to Shift Catmem Queue Tokens When Packing Result

ppenna:
## Description

- This PR closes https://github.com/microsoft/demikernel/issues/786


iyzhang (approved):
We should consider making the Catmem queue tokens the ones that are shifted as they are the ones that are not externally visible. Then we will not have to make any external changes when we move to a single shared scheduler.

ppenna:
> We should consider making the Catmem queue tokens the ones that are shifted as they are the ones that are not externally visible. Then we will not have to make any external changes when we move to a single shared scheduler.

Indeed, but that would also fix once we get full support for multiple liboses to co-exist - the queue token space will not overlap after that, right?

PR #794, opened 2023-06-26
[catmem] Name Prefix Shared Memory Regions

ppenna:
## Description

This PR closes https://github.com/microsoft/demikernel/issues/318.


iyzhang (approved):
LGTM

PR #796, opened 2023-06-26
[ci] Improve CI configurability

anandbonde:
This PR attempts to simplify the configuration and control over our CI infra. With this change, the libos->test->scenario->parameters mapping will be config. driven.

This PR only addresses system tests. Other areas of CI will be evaluated and address in separate PRs if needed.

Related issue: https://github.com/microsoft/demikernel/issues/804

### DIFF
--- a/tools/ci/config/ci_map.yaml
+++ b/tools/ci/config/ci_map.yaml
@@ -1,0 +1,138 @@
+catcollar:
+  tcp_close:
+    scenario0:
+      nclients: 32
+      run_mode: sequential
+      who_closes: client
+    scenario1:
+      nclients: 32
+      run_mode: sequential
+      who_closes: server
+    scenario2:
+      nclients: 32
+      run_mode: concurrent
+      who_closes: client
+    scenario3:
+      nclients: 32
+      run_mode: concurrent
+      who_closes: server
+  tcp_ping_pong: {}
+  tcp_push_pop: {}
+  udp_ping_pong: {}
+  udp_push_pop: {}
+catmem:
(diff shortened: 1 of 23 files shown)


tools/ci/config/ci_map.yaml
```diff
@@ -0,0 +1,117 @@
...
+      run_mode: concurrent
+      who_closes: server
+  tcp_ping_pong: {}
+  tcp_push_pop: {}
+  udp_ping_pong: {}
+  udp_push_pop: {}
+catmem:
+  pipe_open: {}
+  pipe_ping_pong: {}
+  pipe_push_pop: {}
```
ppenna, line 26: Aren't we missing standalone tests for Catmem LibOS?

anandbonde: Standalone tests are still executed by the existing code. This PR only addresses system tests. We will work on that as a separate task/PR to make those tests configurable.

ppenna: I see. That was not briefed in the PR description. Could add some notes on this there? Also, if there is still work to be done here, could you create issues and link them to this PR?

anandbonde: This was discussed in my presentation to the team where I showed the block of code that I'm addressing (i.e. system tests). Added the same in the PR notes. Created new issue and linked in the PR.

ppenna: Thanks

ppenna (approved):

PR #797, opened 2023-06-26
[catloop|catmem] Fix typo, remove old comments

anandbonde:


src/rust/catloop/futures/connect.rs
```diff
@@ -88,7 +88,6 @@ async fn send_connection_request(
     let handle: TaskHandle = {
         // Get scheduler handle from the task id.
         catmem.borrow().from_task_id(qt)?
-        // Drop the mutable reference because we might yield.
```
iyzhang, line 91: I left this here in case someone is confused about why I put that function call inside a scope.

src/rust/catmem/mod.rs
```diff
@@ -409,7 +409,7 @@ impl CatmemLibOS {
             Some(handle) => Ok(handle),
             None => {
                 let cause: String = format!("invalid queue token (qt={:?})", qt);
-                error!("fromt_task_id(): {}", cause);
```
iyzhang, line 412: Thanks for catching this!

iyzhang (approved):

src/rust/catloop/futures/connect.rs
anandbonde: Merging this now without the comment. There is another instance in the same file which doesn't have the comment. If we decide to get it back in the future, we should do it at all the places.

PR #815, opened 2023-07-05
[catloop] Drop Trait Implementation

ppenna:
## Description

This PR closes https://github.com/microsoft/demikernel/issues/670.

### DIFF
--- a/src/rust/catloop/mod.rs
+++ b/src/rust/catloop/mod.rs
@@ -689,6 +689,11 @@ impl Drop for CatloopLibOS {
     // Releases all sockets allocated by Catnap.
     fn drop(&mut self) {
         for (_, queue) in self.qtable.borrow().get_values() {
+            if let Some(duplex_pipe) = queue.get_pipe() {
+                if duplex_pipe.close().is_err() {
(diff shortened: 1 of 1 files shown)


src/rust/catloop/mod.rs
```diff
@@ -664,6 +675,23 @@ impl CatloopLibOS {
...
+//======================================================================================================================
+
+impl Drop for CatloopLibOS {
+    // Releases all sockets allocated by Catnap.
+    fn drop(&mut self) {
+        for (_, queue) in self.qtable.borrow().get_values() {
+            if let Socket::Passive(addr) = queue.get_socket() {
+                if self.ephemeral_ports.borrow_mut().free(addr.port()).is_err() {
+                    warn!("drop(): leaking ephemeral port (port={})", addr.port());
+                }
```
iyzhang, line 689: Should we also close the (control duplex) pipe here if it is not ephemeral?

ppenna: Yes! Thanks for catching this one.

iyzhang (approved):
LGTM

PR #816, opened 2023-07-05
[catloop] Missing `async_close()`

ppenna:
## Description

This PR closes https://github.com/microsoft/demikernel/issues/812.

### DIFF
--- a/src/rust/catloop/mod.rs
+++ b/src/rust/catloop/mod.rs
@@ -461,6 +461,54 @@ impl CatloopLibOS {
         }
     }
 
+    /// Asynchronously closes a socket.
+    pub fn async_close(&mut self, qd: QDesc) -> Result<QToken, Fail> {
+        trace!("async_close() qd={:?}", qd);
+
+        let mut qtable: RefMut<IoQueueTable<CatloopQueue>> = self.qtable.borrow_mut();
+        // Remove socket from sockets table.
+        match qtable.get_mut(&qd) {
+            Some(queue) => {
+                if let Some(duplex_pipe) = queue.get_pipe() {
+                    let qt: QToken = duplex_pipe.async_close()?;
+                    self.catmem_qts.insert(qt, (demi_opcode_t::DEMI_OPC_CLOSE, qd));
+                    Ok(Self::shift_qtoken(qt))
+                } else {
+                    let yielder: Yielder = Yielder::new();
+                    let yielder_handle: YielderHandle = yielder.get_handle();
+                    let coroutine: Pin<Box<Operation>> = Box::pin(async move { (qd, OperationResult::Close) });
+                    let task_id: String = format!("Catloop::close for qd={:?}", qd);
+                    let task: OperationTask = OperationTask::new(task_id, coroutine);
+                    let handle: TaskHandle = match self.scheduler.insert(task) {
+                        Some(handle) => handle,
+                        None => {
+                            let cause: String = format!("cannot schedule co-routine (qd={:?})", qd);
+                            error!("connect(): {}", &cause);
+                            return Err(Fail::new(libc::EAGAIN, &cause));
+                        },
+                    };
+                    queue.add_pending_op(&handle, &yielder_handle);
+                    let qt: QToken = handle.get_task_id().into();
+                    self.catloop_qts.insert(qt, (demi_opcode_t::DEMI_OPC_CONNECT, qd));
+
+                    // Check if the returned queue token falls in the space of queue tokens of the Catmem LibOS.
+                    if Into::<u64>::into(qt) >= Self::QTOKEN_SHIFT {
+                        // This queue token may colide with a queue token in the Catmem LibOS. Warn and keep going.
+                        let message: String = format!("too many pending operations in Catloop");
+                        warn!("async_close(): {}", &message);
+                    }
+
+                    Ok(qt)
+                }
+            },
+            None => {
+                let cause: String = format!("invalid queue descriptor (qd={:?})", qd);
+                error!("close(): {}", &cause);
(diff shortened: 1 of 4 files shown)


src/rust/catloop/mod.rs
```diff
@@ -428,20 +432,75 @@ impl CatloopLibOS {
...
+        match qtable.get_mut(&qd) {
+            Some(queue) => {
+                if let Some(duplex_pipe) = queue.get_pipe() {
+                    let qt: QToken = duplex_pipe.async_close()?;
+                    self.catmem_qts.insert(qt, (demi_opcode_t::DEMI_OPC_CLOSE, qd));
+                    Ok(Self::shift_qtoken(qt))
+                } else {
+                    let yielder: Yielder = Yielder::new();
+                    let yielder_handle: YielderHandle = yielder.get_handle();
+                    let coroutine: Pin<Box<Operation>> = Box::pin(async move { (qd, OperationResult::Close) });
```
iyzhang, line 473: We should free the queue from the qtable here. Not sure how to do it above, I think you might want to use an async block there as well instead of having the duplex pipe insert it into the scheduler.

src/rust/catloop/mod.rs
```diff
@@ -547,38 +606,84 @@ impl CatloopLibOS {
...
+            match opcode {
+                demi_opcode_t::DEMI_OPC_ACCEPT | demi_opcode_t::DEMI_OPC_CONNECT | demi_opcode_t::DEMI_OPC_CLOSE => {},
+                _ => {
+                    let cause: String = format!("unexpected queue token (qt={:?})", qt);
+                    error!("pack_result(): {:?}", &cause);
+                    return Err(Fail::new(libc::EINVAL, &cause));
+                },
+            }
+
+            // If this was an async_close(), harvest resources.
```
iyzhang, line 652: I don't think this clean up code is getting called because the qt would be in the list of catmem qts for a close on the normal path.

iyzhang (changes requested):

src/rust/catloop/mod.rs
ppenna: I am not sure what you meant. The `harvest_resources_on_close` will do that in `pack_result`.

src/rust/catloop/mod.rs
ppenna: This is more a sanity check, as we do for accept and connect, right?

src/rust/catloop/mod.rs
```diff
@@ -564,38 +612,84 @@ impl CatloopLibOS {
         Err(Fail::new(libc::EINVAL, &cause))
     }
 
+    /// Harvests resources on close.
+    fn harvest_resources_on_close(&mut self, qd: QDesc) -> Result<(), Fail> {
+        // If this was an async_close(), harvest resources.
```
anandbonde, line 617: How does this function know whether it's called in the `async_close()` path?

src/rust/catloop/mod.rs
```diff
@@ -564,38 +612,84 @@ impl CatloopLibOS {
...
+        match qtable.get_mut(&qd) {
+            Some(queue) => {
+                // Rollback the port allocation.
+                if let Socket::Active(Some(addr)) | Socket::Passive(addr) = queue.get_socket() {
+                    if EphemeralPorts::is_private(addr.port()) {
+                        if self.ephemeral_ports.borrow_mut().free(addr.port()).is_err() {
+                            // We fail if and only if we attempted to free a port that was not allocated.
+                            // This is unexpected, but if it happens, issue a warning and keep going,
+                            // otherwise we would leave the queue in a dangling state.
+                            warn!("pakc_result(): leaking ephemeral port (port={})", addr.port());
```
anandbonde, line 628: pack_result() - typo

src/rust/catloop/mod.rs
```diff
@@ -564,38 +612,84 @@ impl CatloopLibOS {
...
+        // If this was an async_close(), harvest resources.
+        let mut qtable: RefMut<IoQueueTable<CatloopQueue>> = self.qtable.borrow_mut();
+        match qtable.get_mut(&qd) {
+            Some(queue) => {
+                // Rollback the port allocation.
+                if let Socket::Active(Some(addr)) | Socket::Passive(addr) = queue.get_socket() {
+                    if EphemeralPorts::is_private(addr.port()) {
+                        if self.ephemeral_ports.borrow_mut().free(addr.port()).is_err() {
+                            // We fail if and only if we attempted to free a port that was not allocated.
+                            // This is unexpected, but if it happens, issue a warning and keep going,
```
anandbonde, line 626: If it can only fail in this situation, should the message state that instead of leaking port?

src/rust/catloop/mod.rs
ppenna: Thanks.

src/rust/catloop/mod.rs
ppenna: Sure. I added an extra commit message there.

src/rust/catloop/mod.rs
ppenna: **Case 1** queue is not associated to a duplex pipe (socket is not bound)
- `async_close()` line 476
- `pack_result()` line 664
- `havest_resources_on_close()`

**Case 2** queue is associated to a duplex pipe (socket is bound)
- `async_close()` line 472
- `pack_result()` line 674
- `havest_resources_on_close()`

anandbonde: Oh, I didn't where it was called from. I was referring to the comment. It says that we call it if it was an async close. But there is no check in the function itself.

anandbonde (approved):

PR #818, opened 2023-07-05
[catmem] Move to framing-based protocol

### LINKED ISSUE #859: [catmem] Add support for framing variable sized messages and EOF
## Context

Currently the Catmem libOS writes 16 bits for every byte because it uses the higher byte to encode EOF. Instead we could use framing to encode messages and separate data from control bits.

## Proposed Solution

Each message begins with a magic number (one for data and a different one for EOF) immediately followed by a length field. All bytes after up to the length should be interpreted as data, then the next byte after that is either a magic number or an EOF.

## Alternative Solutions

Depending on how the shared memory ring is fixed per issue #842, we may need to change this design or push some of this down to the ring buffer level.

iyzhang:
Our current implementation uses half of the bytes in the shared ring buffer to transfer control bits but that is unnecessary. I have switched our implementation to the following protocol:

1. Begin with either MAGIC_NUMBER or EOF
2. If MAGIC_NUMBER, the next 2 bytes are [length] of the incoming data.
3. After [length] bytes, there must be either another MAGIC_NUMBER or EOF.

This PR closes #859


src/rust/catmem/futures/close.rs
```diff
@@ -30,9 +27,7 @@ const MAX_RETRIES_PUSH_EOF: u32 = 16;
 /// This function calls close on a file descriptor until it is closed successfully.
 /// TODO merge this with push_eof(), when async_close() and close() are merged.
 pub async fn close_coroutine(ring: Rc<CatmemRingBuffer>, yielder: Yielder) -> Result<(), Fail> {
-    // Maximum number of retries. This is set to an arbitrary small value.
```
ppenna, line 33: Could you rollback this change?

src/rust/catmem/futures/close.rs
```diff
@@ -63,10 +58,7 @@ pub async fn close_coroutine(ring: Rc<CatmemRingBuffer>, yielder: Yielder) -> Re
 /// Pushes the EoF signal to a shared ring buffer.
 /// TODO merge this with close_coroutine(), when async_close() and close() are merged.
 pub fn push_eof(ring: Rc<CatmemRingBuffer>) -> Result<(), Fail> {
-    // Maximum number of retries. This is set to an arbitrary small value.
```
ppenna, line 66: Could you rollback this change?

ppenna (changes requested):
Will that work if there are multiple producers (that is, multiple clients pushing)?

iyzhang:
Closing this because we are implementing framing in the concurrent ring buffer.

PR #819, opened 2023-07-06
[scheduler] Several scope, readability and test improvements

anandbonde:
This PR covers a series of commits for:
- Refactors to improve scope
- Refactors and renames to improve readability
- Simplification of test checks

### DIFF
--- a/src/rust/scheduler/scheduler.rs
+++ b/src/rust/scheduler/scheduler.rs
@@ -93,7 +93,7 @@ impl Scheduler {
             .remove(&task_id)
             .expect("Token should be in the token table");
         let (page, subpage_ix): (&WakerPageRef, usize) = {
-            let (pages_ix, subpage_ix) = self.get_page_indexes(index);
+            let (pages_ix, subpage_ix) = self.get_page_and_offset_indexes(index);
             (&pages[pages_ix], subpage_ix)
         };
         assert!(!page.was_dropped(subpage_ix), "Task was previously dropped");
@@ -121,7 +121,7 @@ impl Scheduler {
         };
         self.tasks.borrow().get(index)?;
         let page: &WakerPageRef = {
-            let (pages_ix, _) = self.get_page_indexes(index);
+            let (pages_ix, _) = self.get_page_and_offset_indexes(index);
             &pages[pages_ix]
         };
         let handle: TaskHandle = TaskHandle::new(task_id, index, page.clone());
@@ -164,16 +164,17 @@ impl Scheduler {
             pages.push(WakerPageRef::default());
         }
         let (page, subpage_ix): (&WakerPageRef, usize) = {
-            let (pages_ix, subpage_ix) = self.get_page_indexes(pin_slab_index);
+            let (pages_ix, subpage_ix) = self.get_page_and_offset_indexes(pin_slab_index);
             (&pages[pages_ix], subpage_ix)
         };
         page.initialize(subpage_ix);
         Some(TaskHandle::new(task_id, pin_slab_index, page.clone()))
     }
 
-    /// Computes the page and page offset of a given task based on its total offset.
-    fn get_page_indexes(&self, index: usize) -> (usize, usize) {
-        (index >> WAKER_BIT_LENGTH_SHIFT, index & (WAKER_BIT_LENGTH - 1))
+    fn get_page_and_offset_indexes(&self, pin_slab_index: usize) -> (usize, usize) {
+        let page_index: usize= pin_slab_index >> WAKER_BIT_LENGTH_SHIFT;
+        let offset_index: usize= pin_slab_index & (WAKER_BIT_LENGTH - 1);
+        (page_index, offset_index)
     }
 
     /// Poll all futures which are ready to run again. Tasks in our scheduler are notified when
--- a/src/rust/scheduler/scheduler.rs
+++ b/src/rust/scheduler/scheduler.rs
@@ -339,7 +339,7 @@ mod tests {
             None => anyhow::bail!("insert() failed"),
         };
         let task_id2: u64 = handle2.get_task_id();
-        crate::ensure_neq!(task_id2, task_id + 1);
+        crate::ensure_neq!(task_id2, task_id);
 
         Ok(())
     }
--- a/src/rust/scheduler/scheduler.rs
+++ b/src/rust/scheduler/scheduler.rs
@@ -130,6 +130,13 @@ impl Scheduler {
 
     /// Insert a new task into our scheduler returning a handle corresponding to it.
     pub fn insert<F: Task>(&self, future: F) -> Option<TaskHandle> {
+        let mut task_ids: RefMut<HashMap<u64, usize>> = self.task_ids.borrow_mut();
+        // If the address space for task ids is close to half full, it will become increasingly difficult to avoid
+        // collisions, so we cap the number of tasks.
+        if task_ids.len() > MAX_NUM_TASKS {
+            panic!("Too many concurrent tasks");
+        }
+
         let mut pages: RefMut<Vec<WakerPageRef>> = self.pages.borrow_mut();
         let mut id_gen: RefMut<SmallRng> = self.id_gen.borrow_mut();
         let task_name: String = future.get_name();
@@ -137,12 +144,6 @@ impl Scheduler {
         let index: usize = self.tasks.borrow_mut().insert(Box::new(future))?;
 
         // Generate a new id. If the id is currently in use, keep generating until we find an unused id.
-        let mut task_ids: RefMut<HashMap<u64, usize>> = self.task_ids.borrow_mut();
-        // If the address space for task ids is close to half full, it will become increasingly difficult to avoid
-        // collisions, so we cap the number of tasks at 16,000.
-        if task_ids.len() > MAX_NUM_TASKS {
-            panic!("Too many concurrent tasks");
-        }
         let task_id: u64 = 'get_id: {
             for _ in 0..MAX_RETRIES_TASK_ID_ALLOC {
                 let id: u64 = id_gen.next_u64() as u16 as u64;
--- a/src/rust/scheduler/scheduler.rs
+++ b/src/rust/scheduler/scheduler.rs
@@ -71,7 +71,7 @@ pub struct Scheduler {
     /// Maps between externally meaningful ids and the index of the task in the slab.
     task_ids: Rc<RefCell<HashMap<u64, usize>>>,
     /// Holds the waker bits for controlling task scheduling.
-    page_refs: Rc<RefCell<Vec<WakerPageRef>>>,
+    waker_page_refs: Rc<RefCell<Vec<WakerPageRef>>>,
     /// Small random number generator for tokens.
     rng: Rc<RefCell<SmallRng>>,
 }
@@ -84,7 +84,7 @@ pub struct Scheduler {
 impl Scheduler {
     /// Given a handle to a task, remove it from the scheduler
     pub fn remove(&self, handle: &TaskHandle) -> Option<Box<dyn Task>> {
-        let page_refs: Ref<Vec<WakerPageRef>> = self.page_refs.borrow();
+        let waker_page_refs: Ref<Vec<WakerPageRef>> = self.waker_page_refs.borrow();
         let task_id: u64 = handle.get_task_id();
         // We should not have a scheduler handle that refers to an invalid id, so unwrap and expect are safe here.
         let pin_slab_index: usize = self
@@ -92,12 +92,12 @@ impl Scheduler {
             .borrow_mut()
             .remove(&task_id)
             .expect("Token should be in the token table");
-        let (page_ref, page_offset): (&WakerPageRef, usize) = {
-            let (page_index, page_offset) = self.get_page_index_and_offset(pin_slab_index);
-            (&page_refs[page_index], page_offset)
+        let (waker_page_ref, waker_page_offset): (&WakerPageRef, usize) = {
+            let (waker_page_index, page_offset) = self.get_waker_page_index_and_offset(pin_slab_index);
+            (&waker_page_refs[waker_page_index], page_offset)
         };
-        assert!(!page_ref.was_dropped(page_offset), "Task was previously dropped");
-        page_ref.clear(page_offset);
+        assert!(!waker_page_ref.was_dropped(waker_page_offset), "Task was previously dropped");
+        waker_page_ref.clear(waker_page_offset);
         if let Some(task) = self.tasks.borrow_mut().remove_unpin(pin_slab_index) {
             trace!(
                 "remove(): name={:?}, id={:?}, pin_slab_index={:?}",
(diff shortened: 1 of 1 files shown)


src/rust/scheduler/scheduler.rs
```diff
@@ -339,7 +339,7 @@ mod tests {
             None => anyhow::bail!("insert() failed"),
         };
         let task_id2: u64 = handle2.get_task_id();
-        crate::ensure_neq!(task_id2, task_id + 1);
+        crate::ensure_neq!(task_id2, task_id);
```
ppenna, line 360: Thanks for fixing that. Important question is: why the test was passing previously?

src/rust/scheduler/scheduler.rs
```diff
@@ -130,19 +130,20 @@ impl Scheduler {
 
     /// Insert a new task into our scheduler returning a handle corresponding to it.
     pub fn insert<F: Task>(&self, future: F) -> Option<TaskHandle> {
+        let mut task_ids: RefMut<HashMap<u64, usize>> = self.task_ids.borrow_mut();
+        // If the address space for task ids is close to half full, it will become increasingly difficult to avoid
+        // collisions, so we cap the number of tasks.
+        if task_ids.len() > MAX_NUM_TASKS {
+            panic!("Too many concurrent tasks");
+        }
+
```
ppenna: What is the motivation for this change?

src/rust/scheduler/scheduler.rs
```diff
@@ -164,16 +164,17 @@ impl Scheduler {
...
         let (page, subpage_ix): (&WakerPageRef, usize) = {
-            let (pages_ix, subpage_ix) = self.get_page_indexes(pin_slab_index);
+            let (pages_ix, subpage_ix) = self.get_page_and_offset_indexes(pin_slab_index);
             (&pages[pages_ix], subpage_ix)
         };
         page.initialize(subpage_ix);
         Some(TaskHandle::new(task_id, pin_slab_index, page.clone()))
     }
 
-    /// Computes the page and page offset of a given task based on its total offset.
```
ppenna, line 171: Please, do not remove existing comments.

src/rust/scheduler/scheduler.rs
```diff
@@ -164,16 +164,17 @@ impl Scheduler {
...
-    fn get_page_indexes(&self, index: usize) -> (usize, usize) {
-        (index >> WAKER_BIT_LENGTH_SHIFT, index & (WAKER_BIT_LENGTH - 1))
+    fn get_page_and_offset_indexes(&self, pin_slab_index: usize) -> (usize, usize) {
+        let page_index: usize= pin_slab_index >> WAKER_BIT_LENGTH_SHIFT;
+        let offset_index: usize= pin_slab_index & (WAKER_BIT_LENGTH - 1);
+        (page_index, offset_index)
```
ppenna: There is a missing space between `usize` and `=`

src/rust/scheduler/scheduler.rs
```diff
@@ -137,108 +137,106 @@ impl Scheduler {
...
                     match poll_result {
-                        Poll::Ready(()) => pages[page_ix].mark_completed(subpage_ix),
+                        Poll::Ready(()) => page_refs[page_index].mark_completed(page_offset),
                         Poll::Pending => (),
                     }
                 }
             }
             // There is some dropped task in this page, so iterate through it.
-            if dropped != 0 {
-                // Handle dropped tasks only.
```
ppenna, line 219: Please, do not remove existing comments.

src/rust/scheduler/scheduler.rs
```diff
@@ -177,6 +175,15 @@ impl Scheduler {
...
 
+    // Add new page(s) to hold this future's status if the current page is filled. This may result in addition of
+    // multiple pages because of the gap between the pin slab index and the current page index.
+    fn add_new_pages_to_accommodate_index_gap(&self, pin_slab_index: usize) {
+        let mut page_ref: RefMut<Vec<WakerPageRef>> = self.page_refs.borrow_mut();
+        while pin_slab_index >= page_ref.len() << WAKER_BIT_LENGTH_SHIFT {
+            page_ref.push(WakerPageRef::default());
+        }
+    }
+
```
ppenna: I don't see any reason why this couldn't be inline as it was before.

src/rust/scheduler/scheduler.rs
```diff
@@ -130,22 +130,33 @@ impl Scheduler {
...
-        let mut rng: RefMut<SmallRng> = self.rng.borrow_mut();
         let task_name: String = future.get_name();
-
-        // The pin slab index is the total index which can be reverse-computed in a page index and an offset within the
-        // page.
+        // The pin slab index can be reverse-computed in a page index and an offset within the page.
         let pin_slab_index: usize = self.tasks.borrow_mut().insert(Box::new(future))?;
 
-        // Generate a new id. If the id is currently in use, keep generating until we find an unused id.
+        self.add_new_pages_to_accommodate_index_gap(pin_slab_index);
```
ppenna, line 139: Why was this added here?

src/rust/scheduler/scheduler.rs
```diff
@@ -130,22 +130,33 @@ impl Scheduler {
...
+        page_ref.initialize(page_offset);
+
+        let task_id = self.get_new_task_id(pin_slab_index);
+
+        trace!("insert(): name={:?}, id={:?}, pin_slab_index={:?}", task_name, task_id, pin_slab_index);
+        Some(TaskHandle::new(task_id, pin_slab_index, page_ref.clone()))
+    }
+
+    // Generate a new id. If the id is currently in use, keep generating until we find an unused id.
+    fn get_new_task_id(&self, pin_slab_index: usize) -> u64 {
```
ppenna, line 156: Before, this happened earlier. Why change?

ppenna (changes requested):
I'd like some clarifications.

src/rust/scheduler/scheduler.rs
anandbonde: I believe it was just a matter of time that it would fail... :)

src/rust/scheduler/scheduler.rs
anandbonde: This check is done regardless of what happens in the rest of the function, so I moved the check up to fail fast and not do any other work if this condition is met.

src/rust/scheduler/scheduler.rs
anandbonde: I fixed this in a subsequent commit. Do you think it's better to alter a prior commit and re-push?

src/rust/scheduler/scheduler.rs
anandbonde: I was mislead by the comment to think that there would be a check whether the task is dropped or not. But, this code only iterates over the `dropped` bits, so there is no need of a check.

src/rust/scheduler/scheduler.rs
```diff
@@ -84,157 +84,175 @@ pub struct Scheduler {
...
 
-        // Generate a new id. If the id is currently in use, keep generating until we find an unused id.
-        let mut task_ids: RefMut<HashMap<u64, usize>> = self.task_ids.borrow_mut();
-        // If the address space for task ids is close to half full, it will become increasingly difficult to avoid
-        // collisions, so we cap the number of tasks at 16,000.
-        if task_ids.len() > MAX_NUM_TASKS {
-            panic!("Too many concurrent tasks");
-        }
-        let task_id: u64 = 'get_id: {
+        self.add_new_pages_to_accommodate_index_gap(pin_slab_index);
```
iyzhang, line 139: maybe name this: allocate_new_pages_up_to_waker_index or something like that

src/rust/scheduler/scheduler.rs
```diff
@@ -84,157 +84,175 @@ pub struct Scheduler {
...
-        let mut pages: RefMut<Vec<WakerPageRef>> = self.pages.borrow_mut();
+        let mut page_refs: RefMut<Vec<WakerPageRef>> = self.page_refs.borrow_mut();
         let mut tasks: RefMut<PinSlab<Box<dyn Task>>> = self.tasks.borrow_mut();
 
         // Iterate through pages.
-        for page_ix in 0..pages.len() {
-            let (notified, dropped): (u64, u64) = {
-                let page: &mut WakerPageRef = &mut pages[page_ix];
-                (page.take_notified(), page.take_dropped())
+        for page_index in 0..page_refs.len() {
```
iyzhang, line 204: It's probably fine to treat these as: page in pages or waker_page in waker_pages

src/rust/scheduler/scheduler.rs
anandbonde: I was hoping to make it clear in the function name so that we don't need the comment. Is it still ambiguous?

src/rust/scheduler/scheduler.rs
iyzhang: I also wanted to check that the scheduler is not allocating task IDs sequentially. I suppose even with a random allocation, this might fail occasionally. Any better ideas for how to test this? Anand was the one that suggested that I write a test case for this.  :)

src/rust/scheduler/scheduler.rs
```diff
@@ -71,9 +71,9 @@ pub struct Scheduler {
     /// Maps between externally meaningful ids and the index of the task in the slab.
     task_ids: Rc<RefCell<HashMap<u64, usize>>>,
     /// Holds the waker bits for controlling task scheduling.
-    pages: Rc<RefCell<Vec<WakerPageRef>>>,
+    page_refs: Rc<RefCell<Vec<WakerPageRef>>>,
```
iyzhang, line 74: I think it's fine to call this waker_pages. It is a reference but the fact that it is a reference isn't important, they are actually being allocated here.

iyzhang (approved):
LGTM. Would appreciate it if you can add a test to ensure our task ids do not follow a pre-defined pattern.

src/rust/scheduler/scheduler.rs
anandbonde: There is no requisite temporal ordering within these operations, so I moved the operations to the closest place where they are used.

src/rust/scheduler/scheduler.rs
anandbonde: I'm moving as many lower levels of abstractions as possible to their specific scope so that the semantics of the calling function stand out clearly. This is a lower level abstraction that is complex enough that it deserves a separate logical block (a function in this case).

src/rust/scheduler/scheduler.rs
anandbonde: I called it `page_refs` because we have defined separate WakerPage and WakerPageRef structures in code explicitly. They are not implicit refs as part of the language, so wanted to have that distinction. Also, there is functionality in the WakerPageRef's non-trivial `default()` implementation (which we use to allocate the pages here).

So, I thought it might be better to call them what they are. Do let me know if you think otherwise.

src/rust/scheduler/scheduler.rs
anandbonde: Oh, I was just asking for a test covering this code :).

If we stick to a good/decent random number generator (RNG) then we don't need to be worried about this at all because that's a test for the RNG, and not for this code.

anandbonde:
> LGTM. Would appreciate it if you can add a test to ensure our task ids do not follow a pre-defined pattern.

Just testing for uniqueness should be good enough for this code because testing for a pattern falls under the umbrella of the RNG that we depend on.

src/rust/scheduler/scheduler.rs
anandbonde: I called it `page_refs` because we have defined separate WakerPage and WakerPageRef structures in code explicitly. They are not implicit refs as part of the language, so wanted to have that distinction. Also, there is functionality in the WakerPageRef's non-trivial `default()` implementation (which we use to allocate the pages here).

So, I thought it might be better to call them what they are. Do let me know if you think otherwise.

src/rust/scheduler/scheduler.rs
anandbonde: Just updated it to match up to that meaning. Let me know if that looks ok.

src/rust/scheduler/scheduler.rs
ppenna: > I believe it was just a matter of time that it would fail... :)

It will not fail if and only if the number generator is somehow is generating numbers in sequence. We can check that by changing the seed number.  Could you verify? I would say that there is something else going on here.

src/rust/scheduler/scheduler.rs
ppenna: Well, I have already expressed my opinion on this many times... I am not arguing against your renaming, but I am not ok with removing comments. I think that this is a middle ground, right?

src/rust/scheduler/scheduler.rs
ppenna: This is ok.

src/rust/scheduler/scheduler.rs
ppenna: The comment does not say "check for dropped tasks", it says to "handle" them. If the comment is not good as expressive as it could be, let's update it, not remove.

src/rust/scheduler/scheduler.rs
anandbonde: I didn't understand your point. This test will only fail (in the current state without my change) if the 2 task ids are in sequence. It has nothing to do with the seed. Which scenario are you thinking about?

src/rust/scheduler/scheduler.rs
anandbonde: The wording in this comment (the word only) led me to think that there would be a check, so I went searching for it. If we are changing the existing comment becuase it is not helpful, then removing it altogether makes sense. Do let me know if you have any suggestions for what needs to be commented at this place - especially when the comment 2 lines above explains what the code below is going to do.

src/rust/scheduler/scheduler.rs
anandbonde: We should consider such opinions case by case and cannot treat them as thumb rules. But, I will keep this comment because it does add context about the total offset.

src/rust/scheduler/scheduler.rs
iyzhang: I think we do not need to discuss this because we want to get rid of this. In fact if there is some way to mark the code as deprecated, that would be great.

src/rust/scheduler/scheduler.rs
iyzhang: sure can we call them waker_refs or waker_page_refs then? The waker part is more important than the page part.

src/rust/scheduler/scheduler.rs
iyzhang: I think we should leave comments for every function because we may want to generate rust docs in the future. In fact, I would advocate that we eventually document every input and output to every function. Sometimes I find myself searching the code for what the function actually outputs and it's really annoying.

src/rust/scheduler/scheduler.rs
iyzhang: I think Pedro is wondering if the random number generator ever generates two numbers in sequence for a given seed and if it does, we should just use a different seed. I think it's pretty unlikely but not impossible that the RNG produces to numbers in sequence, so we should probably not test it. I agree with this change.

iyzhang (approved):

src/rust/scheduler/scheduler.rs
anandbonde: Yes, that makes sense. I will do the change.

src/rust/scheduler/scheduler.rs
anandbonde: Ok, I reverted the comment. We can discuss offline about more on this, if needed.

src/rust/scheduler/scheduler.rs
anandbonde: Ok, I reverted the comment. We can discuss offline about more on deprecation.

ppenna (approved):

PR #836, opened 2023-07-07
[catloop] Always Bound Read from Control Duplex Pipe

### LINKED ISSUE #777: [catloop] Resend connect request causes server not to recognize magic number
## Description

Catloop may send more than one connection request on each connect if the server does not immediately respond. Since the memory queues are not lossy, this request gets stuck behind the next one. The next time that the server goes into the accept loop, it will pull both requests out as one pop and not recognize it as a magic number.

## How to Reproduce

1. Run any of the tcp-close tests
2. See invalid request errors on the server side.

## Expected Behavior

The connect should not try again or identify each request in a unique way in addition to the magic number, so that the server does one accept for each connect request. Essentially, we should ensure exactly once RPC on the control pipe or not do an unbounded read and resend the connect request on the client so soon.

iyzhang:
This PR removes the ability to ever ready from the control duplex pipe on the accept with anything other than the size of MAGIC_CONNECT. Because each MAGIC_CONNECT represents a separate connection request, the code will crash if we ever read more than one at a time. Thus, this PR closes #777 by simply removing the ability to read anything from the control duplex pipe that is not a single MAGIC_CONNECT value.

### DIFF
--- a/src/rust/catloop/futures/accept.rs
+++ b/src/rust/catloop/futures/accept.rs
@@ -52,32 +52,19 @@ pub async fn accept_coroutine(
 ) -> Result<(SocketAddrV4, Rc<DuplexPipe>), Fail> {
     loop {
         // Grab next request from the control duplex pipe.
-        let new_duplex_pipe: Rc<DuplexPipe> = match pop_magic_number(
-            &catmem,
-            control_duplex_pipe.clone(),
-            Some(mem::size_of_val(&CatloopLibOS::MAGIC_CONNECT)),
-            &yielder,
-        )
-        .await
-        {
-            // Received a valid magic number so create the new connection. This involves create the new duplex pipe
-            // and sending the port number to the remote.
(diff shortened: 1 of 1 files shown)


src/rust/catloop/futures/accept.rs
```diff
@@ -52,32 +52,19 @@ pub async fn accept_coroutine(
...
-            new_duplex_pipe.clone(),
-            Some(mem::size_of_val(&CatloopLibOS::MAGIC_CONNECT)),
-            &yielder,
-        )
-        .await
-        {
+        match pop_magic_number(&catmem, new_duplex_pipe.clone(), &yielder).await {
             // Valid response. Connection successfully established, so return new port and pipe to application.
             Ok(true) => return Ok((SocketAddrV4::new(ipv4.clone(), new_port), new_duplex_pipe.clone())),
             // Invalid reponse.
```
ppenna, line 70: "response" is misspelled.

iyzhang: I think it is correct?

iyzhang:
> 

We did. I just made it not optional.

PR #837, opened 2023-07-07
[catloop] Bug Fix: Close new duplex pipe when not MAGIC_CONNECT

### LINKED ISSUE #778: [catloop] Close pipes when ack from client fails
## Description

Currently, the last part of our handshake process expects a magic number back from the client on the newly opened pipe. If we do not get the magic number, we currently do not fail or return an error. We also do not clean up the pipes.

## How to Reproduce

1. Client sends incorrect magic number
2. Server returns from accept with pipe anyway.

We should handle the case where Ok(false) returns on [this line
](https://github.com/demikernel/demikernel/blob/7f745f41a46e9a4f7cb1a3718feaa5acc58bb509/src/rust/catloop/futures/accept.rs#L129).

## Expected Behavior

We should return an error to the application from the accept and close the new pipes.

iyzhang:
The first value that the server expects to receive on any new duplex pipe is MAGIC_CONNECT to validate that the client is using the right port. If we receive anything else, we should close the duplex_pipe and move on to the next request. This PR closes #778


ppenna (approved):
LGTM

PR #838, opened 2023-07-10
[scheduler] Avoid double remove()

anandbonde:
There was a double call to `tasks.remove()`. This PR removes the second call.


ppenna (approved):
Thanks for catching this.

PR #839, opened 2023-07-10
[scheduler] Deprecate TaskHandle.deschedule()

anandbonde:
We discussed offline that `TaskHandle.deschedule()` should be deprecated.


iyzhang (approved):
Thanks!

PR #840, opened 2023-07-11
[scheduler] Simplified the poll function and deprecated the drop part.

anandbonde:
Simplified the `poll` function and deprecated the drop part.


iyzhang (approved):
look great! Thanks for cleaning up this code!

PR #844, opened 2023-07-11
[catnap] Enhancement: More compact state machine code

### LINKED ISSUE #843: [catnap] More compact representation of state machine in Socket
## Context

The Socket structure in any libOS tracks the current state of the socket and whether the current operation is allowed on the current state. 

## Proposed Solution

We could make this into a clearly defined state machine where the current state and the op that you want to apply on it are used to index into an array, which returns whether the op is allowed and if it is not, an errno. This would make it much easier for us to confirm that our error codes conform to POSIX standard.

iyzhang:
This PR moves code that was dispersed into a more compact state machine. It closes #843

### DIFF
--- a/src/rust/catnap/socket.rs
+++ b/src/rust/catnap/socket.rs
@@ -33,6 +33,18 @@ enum SocketState {
     Closed,
 }
 
+#[derive(Copy, Clone, Debug, PartialEq)]
+enum SocketOp {
+    Bind,
+    Listen,
+    Accept,
+    Accepted,
+    Connect,
+    Connected,
+    Close,
+    Closed,
+}
+
 /// A socket.
 #[derive(Copy, Clone, Debug)]
 pub struct Socket {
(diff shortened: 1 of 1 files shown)


ppenna (approved):

src/rust/catnap/socket.rs
```diff
@@ -331,14 +181,235 @@ impl Socket {
     pub fn is_connecting(&self) -> bool {
         self.state == SocketState::Connecting
     }
+
+    /// Given the current state and the operation being executed, this function returns the next state on success and
+    fn get_transition(&self, op: SocketOp) -> Result<SocketState, Fail> {
```
anandbonde, line 186: This function actually returns the next state. Would it be more appropriate to call it `get_next_state()` instead?

src/rust/catnap/socket.rs
```diff
@@ -331,14 +181,235 @@ impl Socket {
...
+                libc::EISCONN,
+            )),
+            SocketOp::Close => Ok(SocketState::Closing),
+            SocketOp::Closed => Err(fail(op, &(format!("socket is busy")), libc::EBUSY)),
+        }
+    }
+
+    fn closing_state(&self, op: SocketOp) -> Result<SocketState, Fail> {
+        match op {
+            SocketOp::Bind => Err(fail(op, &(format!("socket is closing")), libc::EBADF)),
```
anandbonde, line 382: It would be good to evaluate if moving this to a common place (variable/function) would be better.

src/rust/catnap/socket.rs
```diff
@@ -331,14 +181,235 @@ impl Socket {
...
+    fn connected_state(&self, op: SocketOp) -> Result<SocketState, Fail> {
+        match op {
+            // Does this make sense if we didn't go through the Connecting state?
+            SocketOp::Bind => Ok(SocketState::Connected),
+            SocketOp::Listen => Err(fail(
+                op,
+                &(format!("socket is already connected to address: {:?}", self.remote)),
+                libc::EISCONN,
+            )),
+            SocketOp::Accept => Err(fail(
```
anandbonde, line 354: Many of these format strings and arguments are the same for several cases. It would be good to evaluate if these can be centralized. This comment applies to all the newly added functions.

anandbonde (approved):
Looks good!

src/rust/catnap/socket.rs
iyzhang: Yes, we might use a macro eventually, but I'm not sure it would make it much shorter. We could try to shorten the error messages themselves instead if we want it to be more compact

src/rust/catnap/socket.rs
iyzhang: I collapsed them and committed in a separate commit. Please take a look.

PR #849, opened 2023-07-19
[pal] Move SOMAXCONN to PAL

anandbonde:
Making sure that the SOMAXCONN constant is available on Windows as well. It's not a part of libc (like on Linux).


iyzhang (approved):
LGTM

ppenna (approved):
LGTM

PR #850, opened 2023-07-19
[scheduler] Centralize index computations, simplify TaskHandle

anandbonde:
This PR covers the following:

- Centralizes the index computations at one place
- It achieves this by simplifying the TaskHandle to contain the page offset instead of the pin slab index
- Other readability enhancements


src/rust/scheduler/handle.rs
```diff
@@ -55,26 +52,24 @@ pub struct YielderHandle {
...
-        Self { task_id, index, page }
+    pub fn new(task_id: u64, waker_page_ref: WakerPageRef, waker_page_offset: usize) -> Self {
+        Self { task_id, waker_page_ref, waker_page_offset }
     }
 
     /// Queries whether or not the coroutine in the Task has completed.
     pub fn has_completed(&self) -> bool {
-        let subpage_ix: usize = self.index & (WAKER_BIT_LENGTH - 1);
-        self.page.has_completed(subpage_ix)
+        self.waker_page_ref.has_completed(self.waker_page_offset)
```
iyzhang, line 61: Is this mask unnecessary?

src/rust/scheduler/handle.rs
```diff
@@ -55,26 +52,24 @@ pub struct YielderHandle {
...
     pub fn get_task_id(&self) -> u64 {
         self.task_id
     }
 
     /// Removes the task from the scheduler and keeps it from running again.
     #[deprecated]
     pub fn deschedule(&mut self) {
-        let subpage_ix: usize = self.index & (WAKER_BIT_LENGTH - 1);
-        self.page.mark_dropped(subpage_ix);
+        self.waker_page_ref.mark_dropped(self.waker_page_offset);
```
iyzhang, line 72: Same here

src/rust/scheduler/handle.rs
anandbonde: It is necessary becasue it allows us to compute waker page offset for the task from the pin slab index. With this PR, I just moved that existing computation to a common place.

src/rust/scheduler/handle.rs
anandbonde: It is necessary becasue it allows us to compute waker page offset for the task from the pin slab index. With this PR, I just moved that existing computation to a common place.

src/rust/scheduler/handle.rs
iyzhang: right so don't you need to mask waker_page_offset with WAKER_BIT_LENGTH - 1 still?

anandbonde: This masking is still happening, but at a centralized place in `scheduler.rs` now where this function is called.

src/rust/scheduler/handle.rs
```diff
@@ -55,14 +55,13 @@ pub struct YielderHandle {
 /// Associate Functions for Task Handlers
 impl TaskHandle {
     /// Creates a new Task Handle.
-    pub fn new(task_id: u64, pin_slab_index: usize, waker_page_ref: WakerPageRef) -> Self {
-        Self { task_id, pin_slab_index, waker_page_ref }
+    pub fn new(task_id: u64, waker_page_offset: usize, waker_page_ref: WakerPageRef) -> Self {
```
ppenna: I understand that you moved the responsibility to the upper layer, so to refactor and simplify the code.

However, I believe that this is not good the way it is:
- `TaskHandle` drives `WakerPageRef`.
- `TaskHandle` now may pass in invalid bit masks to `WakerPageRef`.
- Neither `TaskHandle` nor `WakerPageRef` will throw an error if an invalid bit mask is passed.

You can fix that by either:
- Placing the masking logic in `TaskHandle::new()`
- Changing the hierarchy of structures to Fail somehow.

anandbonde: It's not just a refactor change, by removing the visibility of the pin slab index from the TaskHandle, we are reducing the responsibility from the TaskHandle component. The TaskHandle does not need to know about the pin slab index and the inner structure based on offset.

Are you suggesting that we should decouple the actions on the `WakerPageRef` from the TaskHandle? If yes, I believe we can treat that as a separate work item and discuss/fix in a separate PR.

The other problems that you listed are still applicable in case of using the pin slab index in the TaskHandle (i.e. the existing implementation).

- The TaskHandle may still be able to pass invalid bit masks if the passed pin slab index itself is invalid.
- There will be no errors in the existing implementaton (even with the masking in TaskHandle) in case an invalid pin slab index is passed to new.

Additionally, by centralizing the computation, it's easier to maintain and add fail-safe checks going forward, if required.

iyzhang (approved):

ppenna (approved):
I take this PR as a step of making it easier to read, but not improving the architecture.

anandbonde:
> 

Yes, that's correct. This PR does not target architecture changes. It just makes the code easier to comprehend so that when we do the architecture change, it's easier.

anandbonde:
> I take this PR as a step of making it easier to read, but not improving the architecture.

Yes, that's correct. This PR does not target architecture changes. It just makes the code easier to comprehend so that when we do the architecture change, it's easier.

PR #851, opened 2023-07-20
[catnap] Enhancement: re-architect Catnap

iyzhang:
This PR re-architects the Catnap libOS such that the libOS follows an object-oriented architecture where the libOS aggregates CatnapQueues but does not know about sockets or the libc calls. The Socket is composed by CatnapQueue but CatnapQueue does not know about libc. And all libc syscalls are encapsulated inside Socket. Thus there is a strict encapsulation hierarchy, where each level only knows the level below: CatnapLibOS encapsulates-> CatnapQueue encapsulates-> Socket encapsulates-> libc syscalls.


src/rust/catnap/queue.rs
```diff
@@ -6,85 +6,220 @@
...
+
+    pub fn set_accept(&mut self) -> Result<(), Fail> {
+        self.socket.borrow_mut().accept()
+    }
+
+    pub async fn accept(&mut self, yielder: Yielder) -> Result<Self, Fail> {
+        loop {
+            let mut socket = self.socket.borrow_mut();
+            match socket.try_accept() {
+                Ok(socket) => {
```
anandbonde, line 88: Would it be better to name this `accepted_socket` instead, to avoid confusion?

src/rust/catnap/queue.rs
```diff
@@ -6,85 +6,220 @@
...
+                        break Err(e);
+                    }
+                },
+                Err(e) => break Err(e),
+            }
+        }
+    }
+
+    pub fn close(&mut self) -> Result<(), Fail> {
+        self.socket.borrow_mut().close()?;
```
anandbonde, line 129: What would be the implications of borrowing the socket here just once for both calls?

src/rust/catnap/queue.rs
```diff
@@ -6,85 +6,220 @@
...
+        &mut self,
+        buf: &mut DemiBuffer,
+        addr: Option<SocketAddrV4>,
+        yielder: Yielder,
+    ) -> Result<(), Fail> {
+        loop {
+            let socket: Ref<Socket> = self.socket.borrow();
+            match socket.try_push(buf, addr) {
+                Ok(()) => {
+                    if buf.len() > 0 {
```
anandbonde, line 164: Inverting this condition would allow to reduce the indentation for the block.

anandbonde (approved):
Overall, this looks good.

src/rust/catnap/queue.rs
iyzhang: Yup, renamed this new_accepted_socket

src/rust/catnap/queue.rs
iyzhang: Not exactly sure what this means but I need to check the buf.len() after the try_push call because it modifies the buf.

src/rust/catnap/queue.rs
iyzhang: I was just being lazy, have changed to a single borrow.

src/rust/catnap/queue.rs
anandbonde: If you check for `if buf.len() == 0` and break Ok, then you don't need the first part in `if` or `else`.

I mean this:

```
if buf.len() == 0 {
    break Ok(());
}
// Operation in progress. Check if cancelled.
drop(socket);
if let Err(e) = yielder.yield_once().await {
        break Err(e);
}
```
It's not a blocker though.

iyzhang: Ah got it

src/rust/catnap/socket.rs
```diff
@@ -62,107 +82,297 @@ pub struct Socket {
...
+    pub fn new(domain: libc::c_int, typ: libc::c_int) -> Result<Self, Fail> {
+        // Parse communication domain.
+        if domain != libc::AF_INET {
+            return Err(Fail::new(libc::ENOTSUP, "communication domain not supported"));
+        }
+
+        // Parse socket type and protocol.
+        if (typ != libc::SOCK_STREAM) && (typ != libc::SOCK_DGRAM) {
+            return Err(Fail::new(libc::ENOTSUP, "socket type not supported"));
         }
```
ppenna, line 94: We are not consistent on parameter checking logic. Some are now performed in `Socket`, others in `CatnapLibOS` and, yet others in both.

I have the following suggestion:
- Keep the parameter checking logic in `CatnapLibos` level.
- Add debug asserts on `Socket` level.

src/rust/catnap/mod.rs
```diff
@@ -634,43 +373,38 @@ impl CatnapLibOS {
         trace!("pushto() qd={:?}", qd);
 
         match self.runtime.clone_sgarray(sga) {
-            Ok(buf) => {
+            Ok(mut buf) => {
```
ppenna, line 376: Same here.

src/rust/catnap/mod.rs
```diff
@@ -634,43 +373,38 @@ impl CatnapLibOS {
...
                 if buf.len() == 0 {
                     return Err(Fail::new(libc::EINVAL, "zero-length buffer"));
                 }
 
+                let mut queue: CatnapQueue = match self.qtable.borrow_mut().get_mut(&qd) {
+                    Some(queue) => queue.clone(),
+                    None => return Err(Fail::new(libc::EBADF, "invalid queue descriptor")),
+                };
```
ppenna, line 384: Why not using the new `get_queue()` function?

src/rust/catnap/mod.rs
```diff
@@ -685,40 +419,34 @@ impl CatnapLibOS {
         // We just assert 'size' here, because it was previously checked at PDPIX layer.
         debug_assert!(size.is_none() || ((size.unwrap() > 0) && (size.unwrap() <= limits::POP_SIZE_MAX)));
 
+        let mut queue: CatnapQueue = match self.qtable.borrow_mut().get_mut(&qd) {
+            Some(queue) => queue.clone(),
+            None => return Err(Fail::new(libc::EBADF, "invalid queue descriptor")),
+        };
```
ppenna, line 425: Why not using the new `get_queue()` function?

ppenna (commented):
Overall looks good.

I noted down some changes in the semantics that we may need to review.

Also, if we are merging this, I would ask you to provide at least one-line comment for each public function that was introduced.

src/rust/catnap/socket.rs
iyzhang: Perfect, thanks!

iyzhang: btw, I would like to keep everything that depends on libc out of CatnapLibOS. Will that still be possible?

src/rust/catnap/mod.rs
```diff
@@ -214,34 +155,17 @@ impl CatnapLibOS {
...
-
-        // Bind underlying socket.
-        let saddr: SockAddr = linux::socketaddrv4_to_sockaddr(&local);
-        match unsafe { libc::bind(fd, &saddr as *const SockAddr, mem::size_of::<SockAddrIn>() as Socklen) } {
-            stats if stats == 0 => {
-                // Update socket.
-                queue.set_socket(&bound_socket);
+        // Issue bind operation.
+        let mut queue: CatnapQueue = self.get_queue(qd)?;
+        queue.prepare_bind()?;
```
iyzhang, line 160: Should we go ahead and put this into the queue.bind() call?

src/rust/catnap/mod.rs
```diff
@@ -256,28 +180,17 @@ impl CatnapLibOS {
...
-                        return Err(Fail::new(errno, "operation failed"));
-                    }
-
-                    // Update socket.
-                    queue.set_socket(&listening_socket);
-                    Ok(())
-                },
-                None => unreachable!("CatnapQueue has invalid underlying file descriptor"),
+        let mut queue: CatnapQueue = self.get_queue(qd)?;
+        queue.prepare_listen()?;
```
iyzhang, line 184: Same here, should we put the into queue.listen()?

src/rust/catnap/mod.rs
```diff
@@ -474,33 +279,22 @@ impl CatnapLibOS {
...
-
-                    qtable.free(&qd);
-                    Ok(())
-                },
-                None => unreachable!("CatnapQueue has invalid underlying file descriptor"),
+            Some(queue) => {
+                queue.prepare_close()?;
+                match queue.close() {
+                    Ok(()) => {
+                        queue.commit();
```
iyzhang, line 286: All of the next 3 lines could probably be rolled up into queue.close()?

iyzhang (commented):
Lol,  I cannot approve this PR because it is my PR, but I "approve" it. It is definitely a step in the right direction. I think we could go ahead and merge this and then think further about how to drive the state machine.

src/rust/catnap/mod.rs
ppenna: Yes, they could, but then the responsability for driving the state machine would be spread across the two layers. Eventually we want to place all of them in the queue layer, but this is not possible now.

src/rust/catnap/mod.rs
ppenna: Yes, it could, but then the responsability for driving the state machine would be spread across the two layers. Eventually we want to place all of them in the queue layer, but this is not possible now.

src/rust/catnap/mod.rs
ppenna: Yes, it could, but then the responsability for driving the state machine would be spread across the two layers. Eventually we want to place all of them in the queue layer, but this is not possible now.

ppenna (approved):

PR #853, opened 2023-07-22
[catmem] Restructuring

iyzhang:
This PR reorganizes the Catmem code to follow the new Catnap architecture.


src/rust/catmem/pipe.rs
```diff
@@ -5,32 +5,52 @@
...
+/// padding. Still, this is intentionally set so as the effective capacity is large enough to hold 16 KB of data.
+const RING_BUFFER_CAPACITY: usize = 65536;
 
 //======================================================================================================================
 // Structures
 //======================================================================================================================
 
+pub enum PipeType {
+    Pop,
+    Push,
```
ppenna, line 31: Could you rename these to `PopOnly` and `PushOnly`?

src/rust/catmem/pipe.rs
```diff
@@ -5,32 +5,52 @@
...
+/// Capacity of the ring buffer, in bytes.
+/// This does not correspond to the effective number of bytes that may be stored in the ring buffer due to layout and
+/// padding. Still, this is intentionally set so as the effective capacity is large enough to hold 16 KB of data.
+const RING_BUFFER_CAPACITY: usize = 65536;
 
 //======================================================================================================================
 // Structures
 //======================================================================================================================
 
+pub enum PipeType {
```
ppenna, line 29: Since this is a public enumeration, could we have a RustDoc for it?

src/rust/catmem/pipe.rs
```diff
@@ -39,36 +59,64 @@ pub struct Pipe {
...
             state: PipeState::Opened,
-            eof: false,
-            buffer: Rc::new(buffer),
-        }
+            pipe_type: PipeType::Push,
+            buffer: SharedRingBuffer::<u16>::create(name, RING_BUFFER_CAPACITY)?,
+        })
+    }
+
+    pub fn open(name: &str) -> Result<Self, Fail> {
```
ppenna, line 70: Since this is a public function, could we have a RustDoc for it?

src/rust/catmem/pipe.rs
```diff
@@ -39,36 +59,64 @@ pub struct Pipe {
...
+            state: PipeState::Opened,
+            pipe_type: PipeType::Pop,
+            buffer: SharedRingBuffer::open(name, RING_BUFFER_CAPACITY)?,
+        })
     }
 
-    /// Sets the end of file flag in the target pipe.
-    pub fn set_eof(&mut self) {
-        self.eof = true;
+    pub fn close(&mut self) -> Result<(), Fail> {
```
ppenna, line 78: Since this is a public function, could we have a RustDoc for it?

src/rust/catmem/pipe.rs
```diff
@@ -39,36 +59,64 @@ pub struct Pipe {
...
+        self.get_next_state()
     }
 
-    /// Gets the value of he end of file flag in the target pipe.
-    pub fn eof(&self) -> bool {
-        self.eof
+    pub fn try_close(&mut self) -> Result<(), Fail> {
```
ppenna, line 82: Since this is a public function, could we have a RustDoc for it?

src/rust/catmem/pipe.rs
```diff
@@ -39,36 +59,64 @@ pub struct Pipe {
...
+                Ok(())
+            },
+            Err(_) => Err(Fail::new(libc::EAGAIN, "Could not push")),
+        }
     }
 
-    /// Gets the state of the pipe.
-    pub fn state(&self) -> PipeState {
-        self.state
+    pub fn try_push(&mut self, byte: &u8) -> bool {
```
ppenna, line 92: Since this is a public function, could we have a RustDoc for it?

src/rust/catmem/pipe.rs
```diff
@@ -39,36 +59,64 @@ pub struct Pipe {
...
+        let x: u16 = (*byte & 0xff) as u16;
+        self.buffer.try_enqueue(x).is_ok()
     }
 
-    /// Sets the state of the pipe.
-    pub fn set_state(&mut self, state: PipeState) {
-        self.state = state;
+    pub fn try_pop(&mut self) -> (Option<u8>, bool) {
```
ppenna, line 97: Since this is a public function, could we have a RustDoc for it?

src/rust/catmem/pipe.rs
```diff
@@ -39,36 +59,64 @@ pub struct Pipe {
...
-    /// Gets the value of he end of file flag in the target pipe.
-    pub fn eof(&self) -> bool {
-        self.eof
+    pub fn try_close(&mut self) -> Result<(), Fail> {
+        match self.buffer.try_enqueue(EOF) {
+            Ok(()) => {
+                self.get_next_state()?;
+                Ok(())
+            },
+            Err(_) => Err(Fail::new(libc::EAGAIN, "Could not push")),
```
ppenna, line 88: Could you follow the same pattern that we have been using in the code, where we always log an error message when we fail?

src/rust/catmem/pipe.rs
```diff
@@ -39,36 +59,64 @@ pub struct Pipe {
...
-    pub fn buffer(&self) -> Rc<SharedRingBuffer<u16>> {
-        self.buffer.clone()
+    pub fn get_next_state(&mut self) -> Result<(), Fail> {
+        self.state = match self.state {
+            PipeState::Opened => match self.pipe_type {
+                PipeType::Push => PipeState::Closing,
+                PipeType::Pop => PipeState::Closed,
+            },
+            PipeState::Closing => PipeState::Closed,
+            PipeState::Closed => return Err(Fail::new(libc::EINVAL, "Cannot close an already closed pipe")),
```
ppenna, line 118: Could you follow the same pattern that we have been using in the code, where we always log an error message when we fail?

src/rust/catmem/queue.rs
```diff
@@ -7,66 +7,192 @@
...
 //======================================================================================================================
 // Associated Functions
 //======================================================================================================================
 
 impl CatmemQueue {
-    pub fn new(ring: SharedRingBuffer<u16>) -> Self {
-        Self {
-            pipe: Pipe::new(ring),
-            pending_ops: HashMap::<TaskHandle, YielderHandle>::new(),
+    pub fn new(name: &str) -> Result<Self, Fail> {
```
ppenna, line 55: Could we rename this so that it matches the operation on the libos layer? (ex: `create_pipe()`)

src/rust/catmem/queue.rs
```diff
@@ -7,66 +7,192 @@
...
+        Ok(Self {
+            pipe: Rc::new(RefCell::<Pipe>::new(Pipe::create(name)?)),
+            pending_ops: Rc::new(RefCell::<HashMap<TaskHandle, YielderHandle>>::new(HashMap::<
+                TaskHandle,
+                YielderHandle,
+            >::new())),
+        })
+    }
+
+    pub fn open(name: &str) -> Result<Self, Fail> {
```
ppenna, line 65: Could we rename this so that it matches the operation on the libos layer? (ex: `open_pipe()`)

src/rust/catmem/mod.rs
```diff
@@ -135,208 +114,117 @@ impl CatmemLibOS {
...
         trace!("push() qd={:?}", qd);
 
-        match self.clone_sgarray(sga) {
-            Ok(buf) => {
-                if buf.len() == 0 {
-                    let cause: String = format!("zero-length buffer (qd={:?})", qd);
-                    error!("push(): {}", cause);
-                    return Err(Fail::new(libc::EINVAL, &cause));
-                }
+        let buf = self.clone_sgarray(sga)?;
```
ppenna, line 193: Could you add a type annotation here?

ppenna (commented):
Shouldn't this PR have `dev` as head branch?

iyzhang:
> Shouldn't this PR have `dev` as head branch?

I needed some changes from PR #851

src/rust/catmem/mod.rs
```diff
@@ -348,80 +236,55 @@ impl CatmemLibOS {
...
-                trace!("pop() qt={:?}", qt);
-                Ok(qt)
-            },
+
+                // Release the queue descriptor, even if pushing EoF failed. This will prevent any further operations on the
+                // queue, as well as it will ensure that the underlying shared ring buffer will be eventually released.
+                qtable_.free(&qd);
+                let cause: String = format!("connection reset (qd={:?})", qd);
+                error!("pop(): {:?}", &cause);
+                return (qd, OperationResult::Failed(Fail::new(libc::ECONNRESET, &cause)));
```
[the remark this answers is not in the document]
iyzhang: Yup, that was a mistake.

src/rust/catmem/queue.rs
iyzhang: Not sure what to do about this. Would be ok to changing everything back to pipe instead of ring but it's weird to name it open_pipe when it actually opens a shared ring buffer.

src/rust/catmem/queue.rs
iyzhang: Not sure what to do about this. Would be ok to changing everything back to pipe instead of ring but it's weird to name it create_pipe when it actually create a shared ring buffer.

src/rust/catmem/ring/pop_ring.rs
```diff
@@ -0,0 +1,75 @@
...
+
+/// An endpoint for a unidirectional queue built on a shared ring buffer
+/// The state of the underlying ring reference
+#[derive(Copy, Clone, Debug, PartialEq)]
+enum RingState {
+    /// A pipe that is opened.
+    Opened,
+    /// A pipe that is closed. For a consumer (pop) pipe, this means that it has received an EOF.
+    Closed,
+}
```
ppenna, line 25: This can be moved to a sibling module and be shared with PushRing.

src/rust/catmem/ring/push_ring.rs
```diff
@@ -0,0 +1,114 @@
...
+/// The state of the underlying ring reference
+#[derive(Copy, Clone, Debug, PartialEq)]
+enum RingState {
+    /// A pipe that is opened.
+    Opened,
+    /// A pipe that is closing.
+    Closing,
+    /// A pipe that is closed. For a producer (push) pipe, this means that it has successfully set an EOF.
+    Closed,
+}
```
ppenna, line 34: This can be moved to a sibling module and be shared with PopRing.

src/rust/catmem/mod.rs
```diff
@@ -80,11 +54,28 @@ const RING_BUFFER_CAPACITY: usize = 65536;
...
+/// Operation Result
+pub enum OperationResult {
+    Push,
+    Pop(DemiBuffer),
+    Close,
+    Failed(Fail),
+}
+
+/// Whether this queue is open for push or pop (i.e., producer or consumer).
+pub enum QMode {
```
ppenna, line 67: Can't be this Encoded in QueueType, as we have for UdpSocket and TcpSocket?

src/rust/catmem/mod.rs
```diff
@@ -110,318 +101,199 @@ impl CatmemLibOS {
...
+
+            let mut qtable_: RefMut<IoQueueTable<CatmemQueue>> = qtable_ptr.borrow_mut();
+
+            if eof {
+                match qtable_.get_mut(&qd) {
+                    Some(queue) => {
+                        // return (
+                        //     qd,
+                        //     OperationResult::Failed(Fail::new(libc::ECONNRESET, "this queue was closed")),
+                        // );
```
ppenna, line 270: Should this be commented out?

src/rust/demikernel/libos/mod.rs
```diff
@@ -126,7 +126,22 @@ impl LibOS {
         result
     }
 
-    /// Opens an existing memory queue.
+    /// Creates a new memory queue and connects to producer end.
+    pub fn create_pipe_for_push(&mut self, name: &str) -> Result<QDesc, Fail> {
```
ppenna, line 130: Since this is just a temporary solution, shouldn't we annotate it as deprecated?

src/rust/demikernel/libos/mod.rs
```diff
@@ -137,6 +152,17 @@ impl LibOS {
         }
     }
 
+    /// Opens an existing memory queue and connects to consumer end.
+    pub fn open_pipe_for_pop(&mut self, name: &str) -> Result<QDesc, Fail> {
```
ppenna, line 156: Since this is just a temporary solution, shouldn't we annotate it as deprecated?

src/rust/demikernel/libos/mod.rs
iyzhang: That's true, we could. I wasn't sure if it is weird to add a deprecated function. :)

src/rust/catmem/ring/pop_ring.rs
iyzhang: Since the states are not exactly identical, per Anand's suggestion, we should only support the minimum set of features needed for each function. So we should never need to check that a PopRing is in the Closing state because that is impossible.

src/rust/catmem/mod.rs
iyzhang: Ah, yes it should. It does not comply with our current API spec, which is to return 0 bytes on the first pop with EOF set and then ECONNRESET on subsequent ones. If I uncomment that it will break the tests.

src/rust/catmem/mod.rs
iyzhang: Yes, I think we could but generally we do not use QueueType in arguments to functions when creating the queue so I wasn't sure what to do.

src/rust/catmem/queue.rs
ppenna: Sorry, I was reviewing old code.

src/rust/catmem/mod.rs
ppenna: Fixed.

src/rust/demikernel/libos/mod.rs
ppenna: I find it important :) I added that...

src/rust/catmem/mod.rs
ppenna: Should we add that as an argument to make the interface nicer? What is your opinion on this?

src/rust/catmem/queue.rs
```diff
@@ -5,68 +5,292 @@
...
+    }
+
+    fn try_close(&self) -> Result<(), Fail> {
+        match &mut *self.ring.borrow_mut() {
+            Ring::PushOnly(ring) => ring.try_close(),
+            Ring::PopOnly(_) => Ok(()),
+        }
+    }
+
+    /// This function
```
iyzhang, line 164: Whoops, I must have started writing this comment and just stopped. :)

src/rust/catmem/queue.rs
```diff
@@ -5,68 +5,292 @@
...
-        &mut self.pipe
+    /// This function commits the queue to closing.
+    pub fn commit(&self) {
+        match &mut *self.ring.borrow_mut() {
+            Ring::PushOnly(ring) => ring.commit(),
+            Ring::PopOnly(ring) => ring.commit(),
+        }
+    }
+
+    /// This function aborts a close operation.
```
anandbonde, line 149: Does this apply to a close operation only? Or can `abort()` be called at anytime (even if not closing)?

src/rust/catmem/queue.rs
```diff
@@ -5,68 +5,292 @@
...
+    fn try_close(&self) -> Result<(), Fail> {
+        match &mut *self.ring.borrow_mut() {
+            Ring::PushOnly(ring) => ring.try_close(),
+            Ring::PopOnly(_) => Ok(()),
+        }
+    }
+
+    /// This function
+    pub async fn async_close(&self, yielder: Yielder) -> Result<(), Fail> {
+        for _ in 0..MAX_RETRIES_PUSH_EOF {
```
anandbonde, line 166: This loop can perhaps be simplified? e.g. something similar to:

```
for _ in 0..MAX_RETRIES_PUSH_EOF {
    if let Ok(_) = self.try_close() {
        return Ok(());
    }
    if let Err(cause) = yielder.yield_once().await {
       return Err(cause),
    }
}
```

src/rust/catmem/queue.rs
```diff
@@ -5,68 +5,292 @@
...
+    /// This function pops a buffer of optional [size] from the queue. If the queue is connected to the push end of a
+    /// shared memory ring, this function returns an error.
+    pub async fn pop(&self, size: Option<usize>, yielder: Yielder) -> Result<(DemiBuffer, bool), Fail> {
+        let size: usize = size.unwrap_or(limits::RECVBUF_SIZE_MAX);
+        let mut buf: DemiBuffer = DemiBuffer::new(size as u16);
+        let mut index: usize = 0;
+        let eof: bool = loop {
+            match self.try_pop()? {
+                (Some(byte), eof) => {
+                    if eof {
```
anandbonde, line 205: Not that it must be fixed now, but the match blocks are long, *maybe* segregating into functions might make them simpler to reason about.

src/rust/catmem/queue.rs
```diff
@@ -5,68 +5,292 @@
...
-        Self {
-            pipe: Pipe::new(ring),
-            pending_ops: HashMap::<TaskHandle, YielderHandle>::new(),
+    /// This function creates a new CatmemQueue and a new shared ring buffer and connects to it to either the consumer
+    /// or producer end indicated by [mode].
+    pub fn create(name: &str, mode: QMode) -> Result<Self, Fail> {
+        match mode {
+            QMode::Push => Ok(Self {
+                ring: Rc::new(RefCell::<Ring>::new(Ring::create_push_ring(name)?)),
+                pending_ops: Rc::new(RefCell::<HashMap<TaskHandle, YielderHandle>>::new(HashMap::<
```
anandbonde, line 63: The creation of pending_ops is repeated. Possibly, it can be moved outside of the match statements?

src/rust/catmem/queue.rs
```diff
@@ -5,68 +5,292 @@
...
+        }
+    }
+
+    /// This function creates a new CatmemQueue and attaches to an existing share ring buffer as either a consumer or
+    /// producer as indicated by [mode].
+    pub fn open(name: &str, mode: QMode) -> Result<Self, Fail> {
+        match mode {
+            QMode::Push => Ok(Self {
+                ring: Rc::new(RefCell::<Ring>::new(Ring::open_push_ring(name)?)),
+                pending_ops: Rc::new(RefCell::<HashMap<TaskHandle, YielderHandle>>::new(HashMap::<
```
anandbonde, line 86: The creation of pending_ops is repeated. Possibly, it can be moved outside of the match statements?

src/rust/catmem/queue.rs
iyzhang: I'm sure the compiler will deduplicate it but do you think it would be easier to read?

src/rust/catmem/queue.rs
iyzhang: Good point, will update.

anandbonde (approved):
Looks good overall.

src/rust/catmem/queue.rs
anandbonde: Its better for readability and maintenance - fewer lines of code to read as well as change, if required. Plus, the decision should only apply to how the `ring` is created, and not how `pending_ops` is created. [Not a blocker]

src/rust/catmem/mod.rs
ppenna: As we discussed, this is be fixed on a different PR.

src/rust/catmem/queue.rs
ppenna: Sure.

src/rust/catmem/queue.rs
ppenna: Sure, but I don't agree with the argument presented here... On the one hand we are shrinking one line of code, on the other we increasing the scope of `pending_ops`  decoupling resorce creation from resource binding. This can lead to misuses in the future that may incurr in side-effects...

Still, I am applying this change

src/rust/catmem/queue.rs
ppenna: In some circustances I agree with that statement, but not with this one. Rationale that goes beyond opinions:
- The funtion is small, which according to our guidelines is whatever it fits in one screen.
- Match statements work with pattern matching and enable aggressive inlining.
- This code is performance criticial and agreessive inlining is required.

src/rust/catmem/queue.rs
ppenna: `abort` may be called at any time. Comment was update to reflect this.

src/rust/catmem/queue.rs
ppenna: Sure.

src/rust/catmem/queue.rs
ppenna: > You can annotate the functions with the inline keyword if necessary. 

- Using inline annotations does not ensure that the compiler will inline. It is a compiler hint. It is true for C code and it is true for rust as well.

> The nested indentation is a bigger concern for me than the actual length here.

Unfortunately Rust is a curly-braced and heavily idented language. 

> Btw, if this is performance critical code, we should measure it using benchmarks and decide based on actual performance numbers.

- No need to measure `pop()` to decide if it is performance critical: pop is on the datapath, and that is a primary concern for demikernel.

anandbonde: I meant we should decide which approach is better by measuring (not whether this is a critical path).

Also, if readability/maintainability comes at the cost of acceptable performance hit, we should take it.

Feel free to defer/ignore this suggestion if you are sure that refactoring this into functions will cause unacceptable performance hit.

ppenna: > I meant we should decide which approach is better by measuring (not whether this is a critical path).

Sure, but this is equivalent about measuring time spent on a context switch routine to decide whether or not it is performance critical.

> Indentation: It's not about rust or braces. Generally, multiple indentation levels indicate room for simplification. (Again, this is not critical because it's not common at other places in the overall codebase - just something to keep watch on because future modifications of this code will make complex and possibly add more logic/conditions).

It is, right? Because that is te pattern matching code adding an extra level of indentation there.

> Feel free to defer/ignore this suggestion if you are sure that refactoring this into functions will cause unacceptable performance hit.

Yeah, I would take these modifications if and only if another reviewer asks for the same thing.

anandbonde: I'm ok with the current code, so I already approved the PR.

**Benchmarking**
Measuring would be the same in either case, but the decision that I'm looking for is which code is better in both aspects (readability + acceptable performance) - specifically the existing code vs extracting into functions.

**Indentation**
If not pattern matching, you'd use if/else. So that wouldn't make a difference in the indentation. Sure, if there's a way to avoid using match and/or reduce the indentation, it would be interesting to see that variant.

ppenna (approved):

PR #855, opened 2023-07-25
[runtime] Enhancement: drain for IoQueueTable

ppenna:
## Description

Adding `drain()` to `IoQueueTable`.


iyzhang (approved):
Great thanks for doing this. It will let us mutate the queue on drop.

anandbonde (approved):
Looks good.

PR #857, opened 2023-07-29
[catnap] Move states and YielderHandle allocation into CatnapQueue

iyzhang:
This PR moves the Catnap libOS towards our eventual architecture. There are a few important steps that this PR implements:

1. The CatnapLibOS now only contains cross-queue functionality and no state. It is never referred to through a mutable reference.
2. All libOS level state is contained in the runtime and qtable, which will eventually move to the runtime. It is dynamically borrowed at runtime.
3. The CatnapQueue only contains single-queue functionality and no state. It is never referred to through a mutable reference.
4. All queue level state is kept in the platform-specific Socket data structure. It is dynamically borrowed at runtime.
5. All asynchronous operations are broken up into two functions at both the libOS level and the queue level: one function for synchronous functionality and another for asynchronous functionality. 

After this PR the following guidelines are recommended for our architecture:
1. Never use mut on stateless objects, specifically the libOS and queue structures. We should never pass &mut self for those.
2. State should be contained in platform-specific abstractions because generally the state that needs to be kept is platform-specific. The exceptions to these are the list of pending ops per queue and the SocketStateMachine, which are shared across libOSes and platforms.
3. All functionality for every call should be divided by whether the code operates on a single queue or more than one queue. All single-queue code shall be placed in the Cat\*Queue class and all multi-queue code shall be placed in the Cat\*LibOS class.
4. All asynchronous calls must be further divided into synchronous operations and asynchronous operations and placed in respective functions.
5. At the libOS level, there should be no state outside of the queue table (and a platform-specific backmap of addresses to queues). This state will eventually be kept in a shared runtime. If there are exceptions, they should undergo an architectural review.
6. Only asynchronous functions can block and only if they take in a yielder. Always separate synchronous and asynchronous code into two functions.


src/rust/catnap/mod.rs
```diff
@@ -181,281 +174,266 @@ impl CatnapLibOS {
...
-                    return Err(Fail::new(libc::EAGAIN, "cannot schedule co-routine"));
-                },
-            };
-            Ok(handle.get_task_id().into())
-        }
+        }))
     }
 
-    /// Pushes a scatter-gather array to a socket.
+    /// Synchronous code to pushto [buf] to a CatnapQueue and its underlying POSIX socket. This function schedules the
```
anandbonde, line 317: You mean: `Synchronous code to push`? (not pushto)

Also, in the comment, do you mean to call it `buf` or `sga`? Applicable in other places as well. I think the best is to not mention it in the comment at all.

src/rust/catnap/mod.rs
```diff
@@ -181,281 +174,266 @@ impl CatnapLibOS {
...
-            },
-            Err(e) => Err(e),
-        }
+                self.insert_coroutine(task_id, coroutine)
+            })?)
+    }
+
+    /// Asynchronous code to pushto [buf] to [remote] on a CatnapQueue and its underlying POSIX socket. This function
+    /// returns a coroutine that runs asynchronously to pushto a queue and its underlying POSIX socket and performs any
+    /// necessary mult-queue operations at the libOS-level after the pushto succeeds or fails.
```
anandbonde, line 380: multi-queue (typo)

src/rust/catnap/mod.rs
```diff
@@ -181,281 +174,266 @@ impl CatnapLibOS {
...
+    fn pushto_coroutine(
+        &self,
+        qd: QDesc,
+        mut buf: DemiBuffer,
+        remote: SocketAddrV4,
+        yielder: Yielder,
+    ) -> Result<Pin<Box<Operation>>, Fail> {
+        let queue: CatnapQueue = self.get_queue(qd)?;
+        Ok(Box::pin(async move {
+            // Parse result.
```
anandbonde, line 390: Is this comment necessary?

src/rust/catnap/mod.rs
```diff
@@ -181,281 +174,266 @@ impl CatnapLibOS {
...
-        });
-        let task_id: String = format!("Catnap::pop for qd={:?}", qd);
-        self.insert_task_into_scheduler(task_id, qd, &yielder_handle, coroutine)
+        }))
     }
 
-    /// Establishes a connection to a remote endpoint.
-    pub fn connect(&mut self, qd: QDesc, remote: SocketAddrV4) -> Result<QToken, Fail> {
+    /// Synchronous code to establish a connection to a remote endpoint. This function schedules the asynchronous
+    /// co-routine and performs any necessary synchronous, multi-queue operations at the libOS-level before beginning
```
anandbonde, line 228: In some places, `co-routine` is used in place of `coroutine`. It would be good to stick to one style throughout.

src/rust/catnap/mod.rs
```diff
@@ -181,281 +174,266 @@ impl CatnapLibOS {
...
-        }
+        self.get_queue(qd)?.close()?;
+        // Remove the queue from the queue table.
+        self.qtable.borrow_mut().free(&qd);
+        Ok(())
     }
 
-    /// Asynchronous close
-    pub fn async_close(&mut self, qd: QDesc) -> Result<QToken, Fail> {
+    /// Synchronous code to asynchronously close a queue. This function schedules the coroutine that asynchronously
```
anandbonde, line 277: This comment is a bit confusing. Is there a way to make it clearer? (e.g., possibly removing/rewording the first statement)

src/rust/catnap/mod.rs
```diff
@@ -181,281 +174,266 @@ impl CatnapLibOS {
...
+            .async_close(move |yielder: Yielder| -> Result<TaskHandle, Fail> {
+                // Async code to close this queue.
+                let coroutine: Pin<Box<Operation>> = self.close_coroutine(qd, yielder)?;
+                let task_id: String = format!("Catnap::close for qd={:?}", qd);
+                self.insert_coroutine(task_id, coroutine)
+            })?)
+    }
+
+    /// Asynchronous code to close a queue. This function returns a coroutine that runs asynchronously to close a queue
+    /// and the underlying POSIX socket and performs any necessary mult-queue operations at the libOS-level after
```
anandbonde, line 295: multi-queue (typo)

src/rust/catnap/mod.rs
```diff
@@ -181,281 +174,266 @@ impl CatnapLibOS {
...
-            },
-            Err(e) => Err(e),
-        }
+                self.insert_coroutine(task_id, coroutine)
+            })?)
+    }
+
+    /// Asynchronous code to push [buf] to a CatnapQueue and its underlying POSIX socket. This function returns a
+    /// coroutine that runs asynchronously to push a queue and its underlying POSIX socket and performs any necessary
+    /// mult-queue operations at the libOS-level after the push succeeds or fails.
```
anandbonde, line 341: multi-queue (typo)

src/rust/catnap/mod.rs
```diff
@@ -181,281 +174,266 @@ impl CatnapLibOS {
...
+                self.insert_coroutine(task_id, coroutine)
+            })?)
+    }
 
-        // Issue pop operation.
-        let yielder: Yielder = Yielder::new();
-        let yielder_handle: YielderHandle = yielder.get_handle();
-        let coroutine: Pin<Box<Operation>> = Box::pin(async move {
+    /// Asynchronous code to pop data from a CatnapQueue and its underlying POSIX socket of optional [size]. This
+    /// function returns a coroutine that asynchronously runs pop and performs any necessary mult-queue operations at
```
anandbonde, line 423: multi-queue (typo)

src/rust/catnap/mod.rs
```diff
@@ -181,281 +174,266 @@ impl CatnapLibOS {
...
+                // Asynchronous accept code.
+                let coroutine: Pin<Box<Operation>> = self.accept_coroutine(qd, yielder)?;
+                // Insert async coroutine into the scheduler.
+                let task_id: String = format!("Catnap::accept for qd={:?}", qd);
+                self.insert_coroutine(task_id, coroutine)
+            })?)
+    }
+
+    /// Asynchronous cross-queue code for accepting a connection. This function returns a coroutine that runs
+    /// asynchronously to accept a connection and performs any necessary mult-queue operations at the libOS-level after
```
anandbonde, line 200: multi-queue (typo)

src/rust/catnap/mod.rs
```diff
@@ -181,281 +174,266 @@ impl CatnapLibOS {
...
-                return (qd, OperationResult::Failed(e));
-            }
-            // Parse result.
-            match queue.connect(remote, yielder).await {
-                Ok(()) => {
-                    queue.commit();
-                    (qd, OperationResult::Connect)
-                },
+    /// Asynchronous code to establish a connection to a remote endpoint. This function returns a coroutine that runs
+    /// asynchronously to connect a queue and performs any necessary mult-queue operations at the libOS-level after
```
anandbonde, line 244: multi-queue (typo)

src/rust/catnap/queue.rs
```diff
@@ -81,131 +83,248 @@ impl CatnapQueue {
...
+        Ok(task_handle.get_task_id().into())
     }
 
-    /// Begins the accepted operation.
-    pub fn prepare_accepted(&self) -> Result<(), Fail> {
-        self.socket.borrow_mut().prepare_accepted()
-    }
+    /// Asynchronously accepts a new connection on the queue. This function contains all of the single-queue,
+    /// asynchronous code necessary to run an accept and any single-queue functionality after the accept completes.
+    pub async fn do_accept(&self, yielder: Yielder) -> Result<Self, Fail> {
```
anandbonde, line 146: Curious, why was this renamed from `accept()` to `do_accept()`? Was it because this is an `async` function? If that's the case, would it be better to call it `async_accept()` instead?

- this comment is applicable to ther do_* cases as well

anandbonde (changes requested):
Left a few comments.

src/rust/catnap/queue.rs
iyzhang: Yeah, the issue is that we currently have a close and an async_close. I could also call it accept_coroutine? I was actually using an async convention that we used in TAPIR, where the actual code to run an async operation is in a function called do_*

anandbonde: Ok, got it. If we follow the same convention throughout, then do_* sounds good to me.

src/rust/catnap/mod.rs
ppenna: I dont see how to reword this different from what Irene wrote initially. Please suggest.

anandbonde: The first statement can be removed.

anandbonde (approved):
Looks good.

ppenna (approved):

src/rust/catnap/mod.rs
ppenna: Not sure... It refers to the synchronous part of the asynchronous close call.

anandbonde: As discussed offline, it's ok to leave this as is.

PR #864, opened 2023-08-07
[catmem] Mutex locking

iyzhang:
This PR adds a simple mutex to the scheduler that can be used to guarantee serialized access to the data structure. The mutex can either be used by polling it or passing in a yielder that will yield back to the scheduler if the mutex is currently locked. This PR is a fix to issue https://github.com/microsoft/demikernel/issues/842, although all the producers must be in a single process and all consumers as well.


src/rust/catmem/ring/push_ring.rs
```diff
@@ -69,16 +74,33 @@ impl PushRing {
...
-    /// Try to send a byte through the shared memory ring. If there is no space, return [false], otherwise, return [true] if successfully enqueued.
+    /// Try to send a byte through the shared memory ring. If there is no space or another thread is writing to this
+    /// ring, return [false], otherwise, return [true] if successfully enqueued.
     pub fn try_push(&mut self, byte: &u8) -> Result<bool, Fail> {
         let x: u16 = (*byte & 0xff) as u16;
-        Ok(self.buffer.try_enqueue(x).is_ok())
+        // Try to lock the ring buffer.
+        if !self.mutex.try_lock() {
+            return Ok(false);
+        }
```
ppenna, line 97: Should we follow the same semantics of `try_close()` and fail with EAGAIN?

src/rust/catmem/ring/pop_ring.rs
```diff
@@ -55,7 +60,12 @@ impl PopRing {
         if self.state_machine.is_closed() {
             return Err(Fail::new(libc::ECONNRESET, "queue was closed"));
         };
-        if let Some(bytes) = self.buffer.try_dequeue() {
+        if !self.mutex.try_lock() {
+            return Ok((None, false));
+        }
```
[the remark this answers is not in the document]
ppenna: Nevermind, we can't because the protocol keeps retrying. Adding a wan here, if it is too often looks like we are facing contention.

src/rust/catmem/ring/push_ring.rs
ppenna: Same here..

src/rust/scheduler/mutex.rs
```diff
@@ -0,0 +1,130 @@
...
+//======================================================================================================================
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use ::anyhow::Result;
+
+    #[test]
+    fn test_mutex_aquire_release() -> Result<()> {
```
anandbonde, line 97: Typo in the test function names (acquire).

src/rust/scheduler/mutex.rs
```diff
@@ -0,0 +1,130 @@
...
+            },
+        }
+    }
+}
+
+//======================================================================================================================
+// Tests
+//======================================================================================================================
+
+#[cfg(test)]
```
anandbonde, line 90: It would be good to have a test that directly attempts to `release` without acquiring.

anandbonde (approved):
Looks good.

src/rust/scheduler/mutex.rs
ppenna: Sure.

ppenna (approved):

Issue #842, opened 2023-07-11
[catloop] Ring Buffer Should Support Multiple Producers


ppenna:
## Description

Catloop LibOS assumes that `RingBuffer` supports multiple concurrent producers, but it does not.

We currently assume this on the connection establishment protocol.

## Code Snippets

In the code snippet bellow, we do not support multiple producers, because there is a race condition when updating the back pointer:

https://github.com/microsoft/demikernel/blob/d9798065dfd8c600be44bf79e49d68b797195ba4/src/rust/collections/ring.rs#L192-L211

## Expected Behavior

We either:

- Fix Catlop LibOS to not rely on this assumption
- Fix `RingBuffer` implementation to support multiple producers.

iyzhang:
Do we want to support multiple producers within a single process or across multiple processes?

iyzhang:
Here is my suggestion for this PR and #859: We build in support for variable length buffers at the shared_ring level to minimize the number of atomic operations that we need to do. Doing so will also enable us to place framing around each object because we will be able to amortize the header cost over the entire buffer and we will be able to do a single copy of the buffer into the ring. I understand that we originally did not want to do this because we expected the shared_ring to only carry meta-data but if we have a solid implementation at this level, we can re-use the same code for a RDMA shared ring buffer. 

I took a look at Section 3.1 of this paper and believe the following implementation would work for RDMA as well. 
1. We keep a front and back pointer and use fetch_add instructions to atomically update them, producers share the back pointer and consumers share the front pointer.
2. We place a 32-bit header between each variable sized message that contains a magic number followed by a length. 
3. To enqueue, we fetch_add the back pointer to reserve space, then write the buffer first and then atomically write the header. The header indicates that the buffer is now ready to read.
4. On dequeue, we poll the header at the front pointer and wait until the value is non-zero there. Since the header contains a magic number and a length, both should be non-zero. To dequeue, we zero the header and move the front pointer.

Having a per-message header lets us avoid keeping another atomic pointer that we have to check and is more robust to miscalculations. Having variable length messages lets us perform I/O more efficiently and amortize the cost of the header over the whole buffer.

ppenna:
> Do we want to support multiple producers within a single process or across multiple processes?

Currently no, eventually yes.

PR #865, opened 2023-08-07
[collections]  Concurrent Ring Buffer

iyzhang:
This PR implements the concurrent ring buffer detailed in issue #842 . It closes #818 and #842 . The file details the invariants that are preserved by the data structure, while allowing lock-free concurrent consumers and producers.

### DIFF
--- a/src/rust/collections/concurrent_ring.rs
+++ b/src/rust/collections/concurrent_ring.rs
@@ -129,7 +129,7 @@ impl ConcurrentRingBuffer {
         }
         // reserve_space will allocate space for the header.
         if let Some(push_offset) = self.reserve_space(len) {
-            debug_assert!(push_offset % 2 == 0);
+            debug_assert!(push_offset % HEADER_SIZE == 0);
             // Push first part of buffer. If longer than the capacity of the ring, wrap around.
             let first_offset: usize = push_offset + HEADER_SIZE;
             let first_len: usize = if push_offset + len + HEADER_SIZE > self.capacity() {
--- a/src/rust/collections/concurrent_ring.rs
+++ b/src/rust/collections/concurrent_ring.rs
@@ -1,0 +1,759 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+//======================================================================================================================
+// Imports
+//======================================================================================================================
+
+use crate::{
+    collections::{
+        raw_array,
+        ring::Ring,
+    },
+    runtime::fail::Fail,
+};
+use ::core::{
+    alloc::Layout,
+    mem,
+    sync::atomic::{
+        self,
+        AtomicU16,
+        AtomicUsize,
+    },
+};
+use ::std::{
+    alloc,
+    ptr::copy,
+};
+
+//======================================================================================================================
+// Constants
+//======================================================================================================================
+
+/// Size of header in bytes = 16-bit buffer length.
+const HEADER_SIZE: usize = 2;
+
+//======================================================================================================================
+// Structures
+//======================================================================================================================
+
+/// |HEADER = LEN(PAYLOAD) | PAYLOAD | possibly 1 byte padding |
+
+/// A lock-free, multi-reader, multi-writer, variable-sized messaging queue. It uses `push_offset` to indicate where
+/// to write incoming messages and `pop_offset` for where to read incoming messages. Additionally each message contains
+/// a [HEADER_SIZE] header that indicates the length of the data (and as a result, pointing to the next valid buffer in
+/// the ring plus some alignment). This header also serves as a locking mechanism: on push, the writer atomically sets
+/// the header to indicated valid data that is ready to be read; on pop, the reader atomically sets the header to 0 in
+/// order to "claim"/lock the message from other concurrent readers.
+///
+/// For correctness, the following invariants
+/// must hold:
+/// 0. push_offset and pop_offset must monotonically increase, except when wrapping around, and must always be aligned
+///    with [HEADER_SIZE].
+/// 1. push_offset == pop_offset only if queue is empty.
+/// 2. push_offset == pop_offset - [HEADER_SIZE] only if queue is full. We only utilize capacity - [HEADER_SIZE] of the
+///    ring buffer so that we can distinguish between these two states.
+/// 3. First [HEADER_SIZE] bytes of every buffer is a header indicating the length of valid data in the .
+/// 4. If the message header is non-zero, the data in the payload must be valid.
+/// 5. If the message is valid (i.e., the message header is non-zero), it must exist between the push_offset and
+///    pop_offset. As a result, the pop_offset can never overtake the push_offset.
+/// 6. If the length header is zero, the data in the buffer may be valid but if it is valid, it is locked and should
+/// not be modified or read.
+/// 7. If the 16 bytes pointed to by pop_offset are zero, then there is either another ongoing pop or there is no valid
+/// data in the buffer.
+pub struct ConcurrentRingBuffer {
+    // Indexes the first empty byte where buffers can be enqueued.
+    push_offset: *mut usize,
+    // Indexes the first buffer that can be popped.
+    pop_offset: *mut usize,
+    // Underlying buffer.
+    buffer: raw_array::RawArray<u8>,
+    /// Is the underlying memory managed by this module?
+    is_managed: bool,
+}
+
+//======================================================================================================================
+// Associated Functions
+//======================================================================================================================
+
+/// Associated functions.
+impl ConcurrentRingBuffer {
+    /// Creates a ring buffer.
+    #[allow(unused)]
+    pub fn new(capacity: usize) -> Result<Self, Fail> {
+        // Check if capacity is invalid.
+        let layout: Layout = Layout::new::<usize>();
+
+        let push_offset: *mut usize = unsafe {
+            let ptr: *mut usize = alloc::alloc(layout) as *mut usize;
+            if ptr.is_null() {
+                alloc::handle_alloc_error(layout);
+            }
+            *ptr = 0;
+            ptr
+        };
+
+        let pop_offset: *mut usize = unsafe {
+            let ptr: *mut usize = alloc::alloc(layout) as *mut usize;
+            if ptr.is_null() {
+                alloc::handle_alloc_error(layout);
+            }
+            *ptr = 0;
+            ptr
+        };
+
+        let me: Self = Self {
+            pop_offset,
+            push_offset,
+            buffer: raw_array::RawArray::<u8>::new(capacity)?,
+            is_managed: true,
+        };
+        // Intialize the first header to 0.
+        me.write_header(0, 0);
+
+        Ok(me)
+    }
+
+    /// Returns the effective capacity of the target ring buffer in bytes.
+    #[allow(unused)]
+    pub fn capacity(&self) -> usize {
+        self.buffer.capacity()
+    }
+
+    #[allow(unused)]
+    pub fn remaining_capacity(&self) -> usize {
+        let push_offset: usize = peek(self.push_offset);
+        let pop_offset: usize = peek(self.pop_offset);
+        //println!("{:?} {:?}", push_offset, pop_offset);
+        self.available_space(push_offset, pop_offset)
+    }
+
+    /// Attempts to insert a buffer of [len] bytes into the ring buffer.
+    pub fn try_push(&self, buf: &[u8]) -> Result<usize, Fail> {
+        let len: usize = buf.len();
+        if len == 0 {
+            return Err(Fail::new(libc::EINVAL, "Buffer must be non-zero length"));
+        }
+        // reserve_space will allocate space for the header.
+        if let Some(push_offset) = self.reserve_space(len) {
+            debug_assert!(push_offset % 2 == 0);
+            // Push first part of buffer. If longer than the capacity of the ring, wrap around.
+            let first_offset: usize = push_offset + HEADER_SIZE;
+            let first_len: usize = if push_offset + len + HEADER_SIZE > self.capacity() {
+                self.capacity() - first_offset
+            } else {
+                len
+            };
+            let buf_ptr: *const u8 = buf.as_ptr();
+            let ring_ptr: *mut u8 = unsafe { self.buffer.get_mut().as_mut_ptr() };
+            // Copy the data into the ring buffer.
+            unsafe {
+                copy(buf_ptr, ring_ptr.add(first_offset), first_len);
+            }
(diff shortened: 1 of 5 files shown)


src/rust/collections/concurrent_ring.rs
```diff
@@ -0,0 +1,509 @@
...
+/// Associated functions.
+impl ConcurrentRingBuffer {
+    /// Creates a ring buffer.
+    #[allow(unused)]
+    pub fn new(capacity: usize) -> Result<Self, Fail> {
+        // Check if capacity is invalid.
+        let layout: Layout = Layout::new::<usize>();
+
+        let push_offset: *mut usize = unsafe {
+            let ptr: *mut usize = alloc::alloc(layout) as *mut usize;
```
anandbonde, line 82: Can this block be extracted into a separate function? It's also repeated below.

src/rust/collections/concurrent_ring.rs
```diff
@@ -0,0 +1,509 @@
...
+//======================================================================================================================
+// Associated Functions
+//======================================================================================================================
+
+/// Associated functions.
+impl ConcurrentRingBuffer {
+    /// Creates a ring buffer.
+    #[allow(unused)]
+    pub fn new(capacity: usize) -> Result<Self, Fail> {
+        // Check if capacity is invalid.
```
anandbonde, line 78: Where is the capacity check done?

src/rust/collections/concurrent_ring.rs
```diff
@@ -0,0 +1,509 @@
...
+        }
+    }
+
+    /// Frees [len] bytes from the ring buffer.
+    fn free_space(&self, current_offset: usize, len: usize) {
+        let len_: usize = Self::align_header(len);
+        let new_offset: usize = current_offset + len_ % self.capacity();
+        // Ensure that the old pop_offset was what we expected.
+        let old: usize = match check_and_set(self.pop_offset, current_offset, new_offset) {
+            Ok(old) => old,
```
anandbonde, line 288: If we always need the old value, do we need a match here?

src/rust/collections/concurrent_ring.rs
```diff
@@ -0,0 +1,509 @@
...
+        let mut result: Result<()> = Ok(());
+        thread::scope(|s| {
+            let mut writers: Vec<thread::ScopedJoinHandle<()>> = Vec::<thread::ScopedJoinHandle<()>>::new();
+            let mut readers: Vec<thread::ScopedJoinHandle<Result<()>>> =
+                Vec::<thread::ScopedJoinHandle<Result<()>>>::new();
+
+            for _ in 0..CONCURRENT_CLIENTS {
+                let writer: thread::ScopedJoinHandle<()> = s.spawn(|| {
+                    let data: [u8; ITERATIONS] = [1; ITERATIONS];
+                    let mut elements: usize = 0;
```
anandbonde, line 478: Apparently all threads would add the same data in the buffers. May be a test that writes thread-specific data would be interesting to observe that there's no overwriting etc. happening...?

Not a blocker, we can do this later...

anandbonde (approved):

src/rust/collections/concurrent_ring.rs
ppenna: Indeed, I added a simple echeck here.

src/rust/collections/concurrent_ring.rs
ppenna: Sure, but this block is the C's equivalent for:

```c
unsinged *ptr = malloc(sizeof(unsigned));
if (ptr == NULL) {
    exit(1);
}
```

Thus you are asking for a `alloc_unsigned()` function. Is that it?

src/rust/collections/concurrent_ring.rs
```diff
@@ -0,0 +1,509 @@
...
+    fn allocate_space(&self, len: usize) -> Option<usize> {
+        let len_: usize = Self::align_header(len);
+        let push_offset: usize = peek(self.push_offset);
+        let pop_offset: usize = peek(self.pop_offset);
+
+        if len_ > self.available_space(push_offset, pop_offset) {
+            return None;
+        }
+        let new_offset: usize = (push_offset + len_) % self.capacity();
+        assert_ne!(new_offset, pop_offset);
```
ppenna, line 274: Should this be turned in a debug assert?

src/rust/collections/concurrent_ring.rs
```diff
@@ -0,0 +1,509 @@
...
+        if let Some(push_offset) = self.allocate_space(len + HEADER_SIZE) {
+            let buf_ptr: *const u8 = buf.as_ptr();
+            unsafe {
+                // Pointer to where to start writing data for this message.
+                let push_ptr: *mut u8 = self.buffer.get_mut().as_mut_ptr().add(push_offset + HEADER_SIZE);
+                // Copy the data into the ring buffer.
+                copy(buf_ptr, push_ptr, len);
+            }
+            // Commit the write by atomically writing the header to release the buffer. The overwritten header MUST be 0.
+            assert_eq!(self.write_header(len, push_offset), 0);
```
ppenna, line 176: Should this be turned into a `debug_assert!()` since it is not allowed by construction?

src/rust/collections/concurrent_ring.rs
```diff
@@ -0,0 +1,509 @@
...
+                return Err(Fail::new(libc::EINVAL, "Buffer is too small to hold next message"));
+            },
+        };
+
+        let buf_ptr: *mut u8 = buf.as_mut_ptr();
+        unsafe {
+            // Pointer to beginning of valid data for this popped message.
+            let pop_ptr: *const u8 = self.buffer.get()[pop_offset + HEADER_SIZE] as *const u8;
+            // Copy the data from the ring buffer.
+            copy(pop_ptr, buf_ptr, pop_len - HEADER_SIZE)
```
ppenna, line 220: I may have missed something, but how does this address a wrap-around?

src/rust/collections/concurrent_ring.rs
ppenna, line 176: Should this be turned into a `debug_assert!()` since it is not allowed by construction?

src/rust/collections/concurrent_ring.rs
ppenna, line 274: This is multi-process, so isn't it possible that some interleaved pops() happen while retrieving push and pop offsets? How is that coupling with concurrency?

iyzhang: Because we write 0 into the header and do not move the pop_offset, that effectively locks the pop end of the ring because all other pops will block on the 0 length, which looks like there is nothing to pop.

src/rust/collections/concurrent_ring.rs
iyzhang: Sure

ppenna:
@iyzhang and @anandbonde please review this when you get a chance.

src/rust/collections/ring.rs
```diff
@@ -305,13 +247,85 @@ impl<T> Drop for RingBuffer<T> {
...
+        let back_ptr: *mut usize = ptr as *mut usize;
+        unsafe { ptr = ptr.add(SIZE_OF_USIZE) };
+        let buffer_ptr: *mut u8 = {
+            let padding: usize = ptr.align_offset(size_of_t);
+            size_of_ring += padding;
+            unsafe { ptr.add(padding) }
+        };
+
+        // Check if memory region is big enough.
+        if size < (size_of_ring + size_of_t) {
```
anandbonde, line 287: Is <= (less than or equal to) ok here?

src/rust/collections/shared_ring.rs
```diff
@@ -53,8 +53,8 @@ impl<T: Copy> SharedRingBuffer<T> {
 //======================================================================================================================
 
 /// Dereference trait implementation for shared ring buffers.
-impl<T: Copy> Deref for SharedRingBuffer<T> {
-    type Target = RingBuffer<T>;
+impl<T: Ring> Deref for SharedRingBuffer<T> {
+    type Target = T;
```
anandbonde, line 57: I guess it's possible to get rid of this `Target` now and just use use T?

src/rust/collections/shared_ring.rs
```diff
@@ -86,10 +87,11 @@ mod test {
...
     fn ring_buffer_on_shm_sequential() -> Result<()> {
         let shm_name: String = "shm-test-ring-buffer-serial".to_string();
-        let ring: SharedRingBuffer<u8> = match SharedRingBuffer::<u8>::create(&shm_name, RING_BUFFER_CAPACITY) {
-            Ok(ring) => ring,
-            Err(_) => anyhow::bail!("creating a shared ring buffer should be possible"),
-        };
+        let ring: SharedRingBuffer<RingBuffer<u8>> =
+            match SharedRingBuffer::<RingBuffer<u8>>::create(&shm_name, RING_BUFFER_CAPACITY) {
+                Ok(ring) => ring,
+                Err(_) => anyhow::bail!("creating a shared ring buffer should be possible"),
```
anandbonde, line 93: Should we log the actual error hear instead?

anandbonde (approved):
Overall looks good. Left some comments...

src/rust/collections/concurrent_ring.rs
ppenna: This is fixed.

src/rust/collections/concurrent_ring.rs
ppenna: Test has ben re-written.

src/rust/collections/concurrent_ring.rs
ppenna: Done.

src/rust/collections/ring.rs
ppenna: `size` can be only equal to or greater than `size_of_ring + size_of_t` (it must be big enough). It must fail otherwise, as the test suggests.

src/rust/collections/shared_ring.rs
```diff
@@ -121,10 +123,11 @@ mod test {
...
         thread::scope(|s| {
             let writer: ScopedJoinHandle<Result<()>> = s.spawn(|| {
-                let ring: SharedRingBuffer<u8> = match SharedRingBuffer::<u8>::create(&shm_name, RING_BUFFER_CAPACITY) {
-                    Ok(ring) => ring,
-                    Err(_) => anyhow::bail!("creating a shared ring buffer should be possible"),
-                };
+                let ring: SharedRingBuffer<RingBuffer<u8>> =
+                    match SharedRingBuffer::<RingBuffer<u8>>::create(&shm_name, RING_BUFFER_CAPACITY) {
+                        Ok(ring) => ring,
+                        Err(_) => anyhow::bail!("creating a shared ring buffer should be possible"),
```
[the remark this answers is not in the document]
ppenna: Yes.

src/rust/collections/shared_ring.rs
```diff
@@ -141,10 +144,11 @@ mod test {
...
                 barrier.wait();
 
-                let ring: SharedRingBuffer<u8> = match SharedRingBuffer::<u8>::open(&shm_name, RING_BUFFER_CAPACITY) {
-                    Ok(ring) => ring,
-                    Err(_) => anyhow::bail!("openining a shared ring buffer should be possible"),
-                };
+                let ring: SharedRingBuffer<RingBuffer<u8>> =
+                    match SharedRingBuffer::<RingBuffer<u8>>::open(&shm_name, RING_BUFFER_CAPACITY) {
+                        Ok(ring) => ring,
+                        Err(_) => anyhow::bail!("opening a shared ring buffer should be possible"),
```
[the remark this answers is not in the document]
ppenna: Yes.

src/rust/collections/shared_ring.rs
ppenna: Yes.

src/rust/collections/shared_ring.rs
ppenna: No, the target Type is associated type of `Deref`. Any structure that implements `Deref` should specify what the associated type is.

src/rust/collections/ring.rs
anandbonde: Ok, sounds good!

ppenna (approved):

src/rust/collections/ring.rs
```diff
@@ -39,6 +39,14 @@ pub struct RingBuffer<T> {
     is_managed: bool,
 }
 
+//======================================================================================================================
+// Traits
+//======================================================================================================================
+
+pub trait Ring: Sized {
```
iyzhang, line 46: I wasn't sure if this was correct when I put it in. Do we know the size ahead of time and is this ok?

src/rust/collections/concurrent_ring.rs
```diff
@@ -0,0 +1,765 @@
...
+
+    /// Attempts to insert a buffer of [len] bytes into the ring buffer.
+    pub fn try_push(&self, buf: &[u8]) -> Result<usize, Fail> {
+        let len: usize = buf.len();
+        if len == 0 {
+            return Err(Fail::new(libc::EINVAL, "Buffer must be non-zero length"));
+        }
+        // reserve_space will allocate space for the header.
+        if let Some(push_offset) = self.reserve_space(len) {
+            debug_assert!(push_offset % 2 == 0);
```
iyzhang, line 132: We might want to replace 2 with HEADER_SIZE

src/rust/collections/concurrent_ring.rs
```diff
@@ -0,0 +1,765 @@
...
+    /// Attempts to remove next message from the ring buffer up to [len] bytes and copies into [buf]. This function
+    /// does not block.
+    pub fn try_pop(&self, buf: &mut [u8]) -> Result<usize, Fail> {
+        let len: usize = buf.len();
+        if len == 0 {
+            return Err(Fail::new(libc::EINVAL, "Buffer must be non-zero length"));
+        }
+
+        // Is this safe or do we need an atomic indirect read?
+        let pop_offset: usize = peek(self.pop_offset);
```
iyzhang, line 184: Having thought about this more, I am fairly sure that this might be an issue with concurrent writers. Let me think of a fix

src/rust/collections/concurrent_ring.rs
ppenna: Indeed.

src/rust/collections/ring.rs
ppenna: You did it correct, because this should be implemented by a concrete types, and in rust all concrete types have a size.

src/rust/collections/concurrent_ring.rs
ppenna: Let me know what you suggest for this.

PR #867, opened 2023-08-09
[catnap] Encapsulate scheduler in runtime

iyzhang:
This PR removes the public visibility of the scheduler in the runtime and adds functions for interacting with it instead. This is a first step towards having a unified runtime and scheduler across libOSes that is created at the Demikernel library level and passed into the constructor of every libOS.

### DIFF
--- a/src/rust/catnap/runtime.rs
+++ b/src/rust/catnap/runtime.rs
@@ -7,10 +7,21 @@
 
 use crate::{
     runtime::{
+        fail::Fail,
         memory::MemoryRuntime,
+        Operation,
+        OperationTask,
+        QToken,
         Runtime,
     },
-    scheduler::scheduler::Scheduler,
+    scheduler::{
+        scheduler::Scheduler,
+        TaskHandle,
+    },
+};
+use ::std::{
+    boxed::Box,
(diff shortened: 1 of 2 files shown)


src/rust/catnap/runtime.rs
```diff
@@ -35,6 +46,35 @@ impl PosixRuntime {
...
+    /// This function inserts a given [coroutine] into the scheduler with [task_id] to identify it.
+    pub fn insert_coroutine(&self, task_id: String, coroutine: Pin<Box<Operation>>) -> Result<TaskHandle, Fail> {
+        let task: OperationTask = OperationTask::new(task_id, coroutine);
+        match self.scheduler.insert(task) {
+            Some(handle) => Ok(handle),
+            None => Err(Fail::new(libc::EAGAIN, "cannot schedule coroutine")),
+        }
+    }
+
+    pub fn remove_coroutine(&self, handle: &TaskHandle) -> OperationTask {
```
ppenna, line 59: Missing doc?

ppenna (approved):

anandbonde (approved):

PR #868, opened 2023-08-09
[runtime] Move scheduler to shared runtime

iyzhang:
This PR encapsulates the scheduler in the runtime (removes its public visibility) and creates a shared DemiRuntime abstraction to hold it. The DemiRuntime is allocated in the Demikernel library level and passed into each libOS. The DemiRuntime implements the Runtime trait and the MemoryRuntime trait, so it can be used for both coroutine scheduling and memory allocation. It is cloneable and the scheduler is currently using dynamic runtime checks for borrowing, so libOSes can safely share as long as only one libOS is using it at a time.


ppenna (approved):
Thanks for this.

anandbonde (approved):

PR #875, opened 2023-08-14
[catloop] Reorganizing Catloop

ppenna:
## Description

This PR isolates queue state on Catloop LibOs.

### DIFF
--- a/src/rust/catloop/mod.rs
+++ b/src/rust/catloop/mod.rs
@@ -175,12 +175,12 @@ impl CatloopLibOS {
             runtime.alloc_ephemeral_port(Some(local.port()))?;
         }
 
-        // Check if queue descriptor is valid and get a mutable reference.
+        // Check if queue descriptor is valid.
         let queue: &mut CatloopQueue = runtime.get_queue(qd)?;
 
         // Check that the socket associated with the queue is not listening.
-        if let Socket::Passive(_) = queue.get_socket() {
-            let cause: String = format!("Cannot bind a listening queue (qd={:?})", qd);
+        if let Socket::Passive(addr) = queue.get_socket() {
+            let cause: String = format!("Cannot bind a listening queue (qd={:?}, addr={:?})", qd, addr);
             error!("bind(): {}", &cause);
             return Err(Fail::new(libc::EBADF, &cause));
         }
@@ -222,8 +222,11 @@ impl CatloopLibOS {
                 error!("listen(): {}", &cause);
                 Err(Fail::new(libc::EOPNOTSUPP, &cause))
             },
-            Socket::Passive(_) => {
-                let cause: String = format!("cannot call listen on an already listening socket (qd={:?})", qd);
+            Socket::Passive(addr) => {
+                let cause: String = format!(
+                    "cannot call listen on an already listening socket (qd={:?}, addr={:?})",
+                    qd, addr
+                );
                 error!("listen(): {}", &cause);
                 Err(Fail::new(libc::EBADF, &cause))
             },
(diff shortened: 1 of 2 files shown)


iyzhang:
Note that in this PR I have to take a mutable borrow of CatloopQueue but I fix it in the reorg once CatloopQueue is cloneable.

iyzhang (approved):
Looks great, thanks for making the fixes.

src/rust/catloop/mod.rs
```diff
@@ -513,7 +513,7 @@ impl CatloopLibOS {
         let runtime: Ref<CatloopRuntime> = self.runtime.borrow();
 
         // Check if the queue token came from the Catloop LibOS.
-        if let Some((ref opcode, _)) = runtime.get_catloop_qt(qt) {
+        if let Some((ref opcode, _)) = runtime.get_catloop_qd(qt) {
```
iyzhang, line 547: This should be get_catloop_qt, right? Because it's a queue token that we are using to get the result?

ppenna: @iyzhang I have changed this because if you look on the signature of the function it returns `Result<(demi_opcode_t, QDesc), Fail>`. Am reading it wrong?

iyzhang: Ahh, ok you are right. Sorry! Maybe we should name this get opcode? We don't seem to use the qd.

src/rust/catloop/mod.rs
```diff
@@ -778,29 +752,33 @@ impl CatloopLibOS {
...
-            }
-            if let Socket::Active(Some(addr)) | Socket::Passive(addr) = queue.get_socket() {
-                if EphemeralPorts::is_private(addr.port()) {
-                    if self.ephemeral_ports.borrow_mut().free(addr.port()).is_err() {
-                        warn!("close(): leaking ephemeral port (port={})", addr.port());
-                    }
-                }
-            }
-        }
+        // This is a stateless object. All resources are managed by underlying components.
```
anandbonde, line 781: Do we need to implement the `drop()` if this struct is stateless?

src/rust/catloop/runtime.rs
```diff
@@ -0,0 +1,171 @@
...
+        self.catloop_qts.insert(qt, (opcode, qd));
+    }
+
+    /// Inserts a given `qt` into the Catmem queue token table.
+    pub fn insert_catmem_qt(&mut self, qt: QToken, opcode: demi_opcode_t, qd: QDesc) {
+        self.catmem_qts.insert(qt, (opcode, qd));
+    }
+
+    /// Gets the [QDesc] associated with `qt` in the Catloop queue token table.
+    pub fn get_catloop_qd(&self, qt: QToken) -> Option<&(demi_opcode_t, QDesc)> {
```
anandbonde, line 99: Could probably be named as: get_catloop_opcode_and_qd?

Also, the comment needs to be corrected.

src/rust/catloop/runtime.rs
```diff
@@ -0,0 +1,171 @@
...
+        self.catloop_qts.get(&qt)
+    }
+
+    /// Gets the [QDesc] associated with `qt` in the Catmem queue token table.
+    pub fn get_catmem_qd(&self, qt: QToken) -> Option<&(demi_opcode_t, QDesc)> {
```
anandbonde, line 104: Could probably be named as: get_catmem_opcode_and_qd?

Also, the comment needs to be corrected.

anandbonde (approved):
Looks good. Left a few comments.

src/rust/catloop/mod.rs
iyzhang: Removed!

PR #876, opened 2023-08-15
[catmem] Move to catnap arch

iyzhang:
This PR moves the catmem libOS closer to the Catnap architecture. It moves all state machine manipulation into the CatmemQueue and pushes some into the Ring data structure. It also moves to the unified Operation and OperationTask data types from the runtime.

### DIFF
--- a/src/rust/catmem/queue.rs
+++ b/src/rust/catmem/queue.rs
@@ -113,12 +113,19 @@ impl CatmemQueue {
         {
             let mut ring: RefMut<Ring> = self.ring.borrow_mut();
             ring.prepare_close()?;
-            ring.close()?;
-            ring.commit();
-            ring.prepare_closed()?;
-            ring.commit();
+            match ring.close() {
+                Ok(()) => {
+                    ring.commit();
+                },
+                Err(e) => {
+                    ring.abort();
+                    return Err(e);
+                },
+            }
         }
+        self.ring.borrow_mut().prepare_closed()?;
         self.cancel_pending_ops(Fail::new(libc::ECANCELED, "this queue was closed"));
+        self.ring.borrow_mut().commit();
         Ok(())
     }
 
--- a/src/rust/catmem/queue.rs
+++ b/src/rust/catmem/queue.rs
@@ -5,16 +5,20 @@
 // Imports
 //======================================================================================================================
 
-use super::{
-    ring::Ring,
-    QMode,
-};
 use crate::{
+    catmem::{
+        ring::{
+            Ring,
+            MAX_RETRIES_PUSH_EOF,
+        },
+        QMode,
+    },
     runtime::{
         fail::Fail,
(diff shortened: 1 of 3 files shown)


src/rust/catmem/queue.rs
```diff
@@ -94,71 +94,48 @@ impl CatmemQueue {
...
     }
 
-    /// This function aborts a pending operation.
-    pub fn abort(&self) {
-        match &mut *self.ring.borrow_mut() {
-            Ring::PushOnly(ring) => ring.abort(),
-            Ring::PopOnly(_) => warn!("abort() called on a pop-only queue"),
+    /// This function closes a ring endpoint.
+    /// TODO merge this with async_close().
+    pub fn close(&mut self) -> Result<(), Fail> {
```
anandbonde, line 112: Shutdown and close are identical, perhaps one of them one of them could just call the other?

src/rust/catmem/ring/mod.rs
```diff
@@ -27,6 +27,9 @@ use crate::runtime::fail::Fail;
 /// padding. Still, this is intentionally set so as the effective capacity is large enough to hold 16 KB of data.
 const RING_BUFFER_CAPACITY: usize = 65536;
 
+/// Maximum number of retries for pushing a EoF signal.
+pub const MAX_RETRIES_PUSH_EOF: u32 = 16;
```
anandbonde, line 31: Is the pub needed?

anandbonde (approved):

src/rust/catmem/ring/mod.rs
iyzhang: Yes, we use it in CatmemQueue as well.

src/rust/catmem/queue.rs
iyzhang: Shutdown doesn't call close, it just sets the queue to the Closed state and cancels pending ops.

anandbonde: Ah ok, yes, I misread as they are very similar :). Please ignore my comment.

src/rust/catmem/ring/mod.rs
anandbonde: Ok, great, then it looks good.

src/rust/catmem/queue.rs
```diff
@@ -113,12 +113,19 @@ impl CatmemQueue {
         {
             let mut ring: RefMut<Ring> = self.ring.borrow_mut();
             ring.prepare_close()?;
-            ring.close()?;
-            ring.commit();
-            ring.prepare_closed()?;
-            ring.commit();
```
ppenna: We should abort if we fail on `self.close()`, otherwise the state machine may end up have an invalid `next` state.

ppenna (approved):

PR #877, opened 2023-08-15
[network] Use PAL constants in state

anandbonde:
These contants are important when building on Windows.


iyzhang (changes requested):
Could you switch this to use Demikernel queue types instead?

iyzhang (approved):

ppenna (approved):
Thanks.

PR #881, opened 2023-08-17
[catmem] Unify scheduler and memory runtime

iyzhang:
This PR moves Catmem onto the same scheduler as Catloop. It also unifies their memory runtimes into the DemiRuntime abstraction.


ppenna (approved):
Thanks for this.

anandbonde (approved):
Looks good!

PR #882, opened 2023-08-18
[catmem] Duplex Ring

ppenna:
## Description

This PR introduces duplex communication capabilities for Catmem rings.


src/rust/demikernel/libos/memory.rs
```diff
@@ -54,7 +52,7 @@ impl MemoryLibOS {
     pub fn create_pipe_for_push(&mut self, name: &str) -> Result<QDesc, Fail> {
```
iyzhang, line 52: Can you delete this function and update the pipe examples in our test suite?

iyzhang (changes requested):

iyzhang (approved):

examples/rust/pipe-ping-pong.rs
```diff
@@ -222,58 +206,38 @@ impl PipeServer {
...
+    pipeqd: QDesc,
 }
 
 // Implementation of the pipe client.
 impl PipeClient {
     pub fn new(mut libos: LibOS, pipe_name: &str) -> Result<Self> {
-        // Open the Tx pipe. This is inverted from the server's perspsective, because the server reads from it's
-        // pipeqd_rx, but the clients writes to it.
-        let pipeqd_tx = match libos.open_pipe(&format!("{}:rx", pipe_name)) {
+        // Open the Tx pipe. This is inverted from the server's perspective, because the server reads from it's
```
anandbonde, line 224: I believe we don't need this comment now, since there is only one pipeqd...?

src/rust/demikernel/libos/memory.rs
ppenna: Done.

examples/rust/pipe-ping-pong.rs
ppenna: Sure.

examples/rust/pipe-ping-pong.rs
```diff
@@ -294,8 +258,8 @@ impl PipeClient {
 // Drop implementation for the pipe client.
 impl Drop for PipeClient {
     fn drop(&mut self) {
-        self.close(self.pipeqd_rx);
-        self.close(self.pipeqd_tx);
+        self.close(self.pipeqd);
+        self.close(self.pipeqd);
```
anandbonde, line 262: pipeqd is closed twice here. Is this desirable?

anandbonde (approved):
Added a couple of comments. The `close()` issue shoud be addressed before merging.

examples/rust/pipe-ping-pong.rs
ppenna: No, fixed.

PR #891, opened 2023-08-22
[catloop] Enhancement: Schedule all coroutines from Catloop

iyzhang:
This PR moves Catloop to implementing its own async functions for push, pop and close, so that it no longer needs to distinguish between catloop and catmem ops. The new async functions directly call the catmem coroutines for each function instead of scheduling a coroutine and returning a queue token.


src/rust/catloop/mod.rs
```diff
@@ -530,6 +519,44 @@ fn pack_result(result: OperationResult, qd: QDesc, qt: u64) -> demi_qresult_t {
                 qr_value: unsafe { mem::zeroed() },
             }
         },
-        _ => panic!("Should be forwarded on to catmem"),
+        OperationResult::Push => demi_qresult_t {
```
anandbonde, line 522: There are 2 Push blocks in the match, how are they differentiated?

src/rust/catloop/socket.rs
```diff
@@ -283,58 +289,91 @@ impl Socket {
...
             }
         } else {
-            Ok((
-                self.do_generic_sync_control_path_call(coroutine, yielder)?
-                    .get_task_id()
-                    .into(),
-                true,
-            ))
+            // We know that the queue descriptor will be replaced so it doesn't matter what we put here.
+            Ok((QDesc::from(0), OperationResult::Close))
```
ppenna, line 322: Let's create a constant for this and change it to the maximum value then :) It is better to read in the logs that way.

src/rust/catloop/mod.rs
ppenna: They don't....

ppenna (approved):

anandbonde (approved):
Looks good.

PR #892, opened 2023-08-22
[test] Correct doc and usage of pushto

anandbonde:
This PR fixes the doc and usage of `demi_pushto()`.

### DIFF
--- a/examples/c/tcp-push-pop.c
+++ b/examples/c/tcp-push-pop.c
@@ -84,7 +84,7 @@ static int connect_get_sockqd(const struct sockaddr_in *const addr)
     return sockqd;
 }
 
-static int push_get_sent_nbytes(const struct sockaddr_in *const addr, const int sockqd, size_t data_size)
+static int push_get_sent_nbytes(const int sockqd, size_t data_size)
 {
     demi_qtoken_t tok = -1;
     demi_qresult_t res = {0};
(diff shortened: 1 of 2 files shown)


man/demi_pushto.md
```diff
@@ -16,8 +16,8 @@ int demi_pushto(demi_qtoken_t *qt_out, int sockqd, const demi_sgarray_t *sga, co
 
 ## Description
 
-`demi_pushto()` asynchronously pushes a scatter-gather array to a socket I/O queue. It only works on connection-mode
-sockets.
+`demi_pushto()` asynchronously pushes a scatter-gather array to a socket I/O queue. It only works on connectionless
+sockets (e.g. UDP).
```
ppenna, line 20: Since we are improving this, could we change it to match [sendto()](https://pubs.opengroup.org/onlinepubs/9699919799/functions/sendto.html)?

> If the socket is a connectionless-mode socket, the message shall be sent to the address specified by dest_addr if no pre-specified peer address has been set. If a peer address has been pre-specified, either the message shall be sent to the address specified by dest_addr (overriding the pre-specified peer address), or the function shall return -1 and set errno to [EISCONN].

anandbonde: Sure, will update the doc to match to sendto  .

Yeah, UDP was just 1 example. But if it works with both connection-mode and connectionless sockets then I'll not mention any example here.

examples/c/tcp-push-pop.c
```diff
@@ -93,7 +93,7 @@ static int push_get_sent_nbytes(const struct sockaddr_in *const addr, const int
 
     assert(sga.sga_segs != 0);
     memset(sga.sga_segs[0].sgaseg_buf, 1, data_size);
-    assert(demi_pushto(&tok, sockqd, &sga, (const struct sockaddr *)addr, sizeof(struct sockaddr_in)) == 0);
+    assert(demi_push(&tok, sockqd, &sga) == 0);
```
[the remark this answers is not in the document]
anandbonde: Will do.

man/demi_pushto.md
anandbonde: Done, updated the doc and added FIXME in code.

iyzhang (approved):
Thanks!

ppenna (approved):
Looks good.

PR #897, opened 2023-08-23
[catnap] Move coroutines to standalone functions

iyzhang:
This PR moves Catnap towards our most recent architecture. It does the following: 
1. Removes moves to avoid implicit moves of variables on the synchronous path to another closure/async context.
2. Removes references to self in coroutines.
3. Explicitly passes all state into async coroutines, including the CatnapQueue, IoQueueTable (and eventually DemiRuntime)


anandbonde (approved):
Looks good.

PR #898, opened 2023-08-23
[runtime] Add IoQueueTable to DemiRuntime

### LINKED ISSUE #495: [runtime] Move IoQueueTable into runtime
## Context

Currently, each libOS holds its own IoQueueTable and manages it. In order to support more than one libOS concurrently, we need to move the IoQueueTable to a shared, centralized place and have the Demikernel layer manage it. 

## Proposed Solution

Create a runtime trait that holds all of the shared resources, including the network runtime, the scheduler and the memory manager. The queue table would fit into that trait as a default feature.

iyzhang:
This PR closes #495

### DIFF
--- a/src/rust/catnap/mod.rs
+++ b/src/rust/catnap/mod.rs
@@ -24,7 +24,7 @@ use crate::{
             MemoryRuntime,
         },
         queue::{
-            IoQueueTable,
+            NetworkQueue,
             Operation,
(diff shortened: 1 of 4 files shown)


ppenna (approved):
LGTM

anandbonde (approved):
Looks good.

PR #901, opened 2023-08-23
[catloop] Don't Allow `accept()` On A Socket That Is Already Accepting

ppenna:
## Description

This PR closes https://github.com/microsoft/demikernel/issues/829.

### DIFF
--- a/src/rust/catloop/queue.rs
+++ b/src/rust/catloop/queue.rs
@@ -88,9 +88,6 @@ impl CatloopQueue {
     where
         F: FnOnce(Yielder) -> Result<TaskHandle, Fail>,
     {
-        // Check whether we are accepting on this queue.
-        self.socket.borrow_mut().prepare_accept()?;
-
         let yielder: Yielder = Yielder::new();
         let yielder_handle: YielderHandle = yielder.get_handle();
 
--- a/src/rust/catloop/queue.rs
+++ b/src/rust/catloop/queue.rs
@@ -88,6 +88,9 @@ impl CatloopQueue {
     where
         F: FnOnce(Yielder) -> Result<TaskHandle, Fail>,
     {
+        // Check whether we are accepting on this queue.
+        self.socket.borrow_mut().prepare_accept()?;
+
         let yielder: Yielder = Yielder::new();
         let yielder_handle: YielderHandle = yielder.get_handle();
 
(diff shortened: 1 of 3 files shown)


src/rust/catloop/socket.rs
```diff
@@ -147,10 +147,14 @@ impl Socket {
     where
         F: FnOnce(Yielder) -> Result<TaskHandle, Fail>,
     {
-        self.state.prepare(SocketOp::Accept)?;
         self.do_generic_sync_control_path_call(coroutine, yielder)
     }
 
+    /// Begins the accept operation.
```
iyzhang, line 153: Is there a reason to pull this out into a one line function? I removed these because the calling functions are also very small and because they do not need to be public.

src/rust/catloop/queue.rs
```diff
@@ -88,6 +88,9 @@ impl CatloopQueue {
     where
         F: FnOnce(Yielder) -> Result<TaskHandle, Fail>,
     {
+        // Check whether we are accepting on this queue.
+        self.socket.borrow_mut().prepare_accept()?;
```
iyzhang, line 92: Is there a reason to do this here instead of inside the socket? I'd like the queue to not have to interact with the socket state.

iyzhang (changes requested):

src/rust/catloop/socket.rs
iyzhang: Also if we do it this way, then we should do it like this for all async coroutines (e.g., connect and close)

ppenna: > Is there a reason to pull this out into a one line function? I removed these because the calling functions are also very small and because they do not need to be public.

Indeed, specifically for `connect()` in Catloop LibOS there is no reason for that. I was mirroring the same code structure that we have in Catnap LibOs. It sounds that we also should fix how responsibilities are layered there as well, right?

https://github.com/microsoft/demikernel/blob/8d7eaf586167386dc45ec1e29536e9b1305cb20b/src/rust/catnap/queue.rs#L112-L118

> Also if we do it this way, then we should do it like this for all async coroutines (e.g., connect and close)

The way that the state machine is driven in Catloop LibOS is mostly buggy and fixes around that will be required. However, I cannot confirm in advance that we will or will not have to introduce such functions, in order to address this problem.

iyzhang (approved):

anandbonde (approved):

PR #902, opened 2023-08-24
[catnap] Moving to Standalone Async Functions

ppenna:
## Description

This is the first commit in a series that attempt to move async functions to standalone ones. This also enables us to move state machine transitions to the sockets level.

### DIFF
--- a/src/rust/catnap/queue.rs
+++ b/src/rust/catnap/queue.rs
@@ -78,7 +78,6 @@ impl CatnapQueue {
     /// Binds the target queue to `local` address.
     pub fn bind(&self, local: SocketAddrV4) -> Result<(), Fail> {
         let mut socket: RefMut<Socket> = self.socket.borrow_mut();
-        socket.prepare_bind()?;
         match socket.bind(local) {
             Ok(_) => {
                 socket.commit();
@@ -94,7 +93,6 @@ impl CatnapQueue {
     /// Sets the target queue to listen for incoming connections.
     pub fn listen(&self, backlog: usize) -> Result<(), Fail> {
         let mut socket: RefMut<Socket> = self.socket.borrow_mut();
-        socket.prepare_listen()?;
         match socket.listen(backlog) {
             Ok(_) => {
                 socket.commit();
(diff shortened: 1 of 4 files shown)


iyzhang:
I have made the following changes:
1. Removed some warnings in CatnapLibOS
2. Moved all state machine transitions to Socket, except for a few rollback operations that must be performed at the CatnapQueue level because Socket does not know about yielder. We will likely change this in the future, especially once we move to poll.
3. Added a new state machine check for connect and accept to ensure that the socket was not closed while we are trying to connect or accept on it. Because we do not cancel all pending operations immediately on async_close, the socket may transition to the closing state while an accept or connect coroutine are still running (similar to push or pop). I added new functions may_accept and may_connect to perform this check. 
4. Added a new test for that check.

iyzhang (approved):

src/rust/catnap/queue.rs
```diff
@@ -252,7 +221,7 @@ impl CatnapQueue {
                     }
                 },
                 Err(e) => {
-                    socket.rollback();
+                    self.socket.borrow_mut().rollback();
```
ppenna, line 224: Why should we keep some rollback logic on the queue level?

ppenna:
Well, I cannot approve my own PR. This looks good to me, I left just a small comment for clarification.

src/rust/catnap/queue.rs
iyzhang: Because I didn't want to move the yielder into the socket level, so I had to handle the cancel error there in the queue level.

ppenna: I understand, but then what is the point to have the responsibility of driving the state machine spread across the socket and queue level? After all, this is what motived this PR initially. I understand that the state machine is something independent of the transport layer (ie: Linux Sockets vs Shared Memory Rings), so for me it makes sense to have all that in the highets level possible. Eventually we want the queue mechanism be shared as much as possible.

iyzhang: Yes, I think we should move the yielder into the socket level but maybe in another PR?

ppenna: > Yes, I think we should move the yielder into the socket level but maybe in another PR?

Sure we can.

anandbonde (approved):
Looks good.

PR #909, opened 2023-08-27
[catcollar] Move to async fns

iyzhang:
This PR begins to move Catcollar to async functions. It removes all futures and all implicit move closures.


ppenna (approved):

anandbonde (approved):
Looks good.

PR #918, opened 2023-08-29
[catnapw] Merge the Catnapw and Catnap LibOSes

anandbonde:
This PR merges the CatnapW and Catnap LibOSes. This PR also ensures that the tcp (Rust/C) and udp ping-pong examples work on Windows. Once the Windows CI is in place, the build and all the tests will start running for Windows as well.


examples/c/udp-ping-pong.c
```diff
@@ -161,7 +161,14 @@ static void client(int argc,
...
 {
     int sockqd = -1;
+
+    #ifdef _WIN32
+    char expected_buf[DATA_SIZE];
+    #endif
+
+    #ifdef  __linux__
     char expected_buf[data_size];
+    #endif
```
ppenna, line 171: What is the rationale for making this different?

examples/c/udp-push-pop.c
```diff
@@ -58,7 +58,14 @@ static void server(int argc,
...
 {
     int sockqd = -1;
+
+    #ifdef _WIN32
+    char expected_buf[DATA_SIZE];
+    #endif
+
+    #ifdef  __linux__
     char expected_buf[data_size];
+    #endif
```
ppenna, line 68: What is the rationale for making this different?

src/rust/catnap/mod.rs
```diff
@@ -519,7 +529,10 @@ fn pack_result(rt: &DemiRuntime, result: OperationResult, qd: QDesc, qt: u64) ->
             qr_value: unsafe { mem::zeroed() },
         },
         OperationResult::Accept((new_qd, addr)) => {
+            #[cfg(target_os = "linux")]
             let saddr: SockAddr = linux::socketaddrv4_to_sockaddr(&addr);
+            #[cfg(target_os = "windows")]
+            let saddr: SockAddr = convert_socket_addr_v4_to_sockaddr(&addr);
```
ppenna, line 535: Could you around this logic around a function so that we don't need to have these checks outside the `pal` component?

src/rust/catnap/socket.rs
ppenna, line 1: At the sockets layer you have mostly copy-paste with few modifications here and there to handle scoping. Since that is the first approach for integrating the two libOses, could you rather have two sockets.rs files? This provides same semantics, while making sure that we did no miss any change in Linux. Later, gradually, we should plan to actually merge the two logics as much as possible, potentially introducing additional abstractions for that.

examples/c/udp-ping-pong.c
anandbonde: The MSVC compiler does not allow using non-constant array length specifications, so I used the constant.

examples/c/udp-push-pop.c
anandbonde: The MSVC compiler does not allow using non-constant array length specifications, so I used the constant.

src/rust/catnap/socket.rs
anandbonde: I actually started with this approach. :) However, in a subsequent discussion with Irene, she suggested that we do it this way, so that it would be easier to switch to MIO (as applicable) in future.

src/rust/catnap/mod.rs
anandbonde: Sure, good idea. Will update.

src/rust/catnap/mod.rs
```diff
@@ -543,9 +556,14 @@ fn pack_result(rt: &DemiRuntime, result: OperationResult, qd: QDesc, qt: u64) ->
...
         OperationResult::Pop(addr, bytes) => match rt.into_sgarray(bytes) {
             Ok(mut sga) => {
+                #[cfg(target_os = "linux")]
                 if let Some(addr) = addr {
                     sga.sga_addr = linux::socketaddrv4_to_sockaddr(&addr);
                 }
+                #[cfg(target_os = "windows")]
+                if let Some(addr) = addr {
+                    sga.sga_addr = convert_socket_addr_v4_to_sockaddr(&addr);
+                }
```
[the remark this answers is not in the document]
anandbonde: Sure, good point. Will update.

examples/c/udp-push-pop.c
ppenna: Oh, I see. You mean that VLA is not supported by MSVC (even in c99 mode?) Anyways so Let's fix linux code so that it matches windows for this :)

src/rust/catnap/socket.rs
ppenna: I see.

@iyzhang I would recommend that we do it in two steps. If you don't agree, I would ask that this change is split into two, this way it is easier to check that we are not missing any recent change on catnap::socket.

examples/c/udp-push-pop.c
anandbonde: MSVC does not fully support C99 mode. It's mainly a C++ compiler which supports C as well. We can change the Linux code to match this, but that will take away the ability to change the `data_size` based on the program arg. Does that sound ok to you folks?

src/rust/catnap/socket.rs
iyzhang: @ppenna I thought that would introduce more conditional windows tags into queue.rs. If there is a way to transparently do it (e.g., with a single include) then we could consider it, but it seemed easier to keep all of the Windows specific code in Catnap inside the socket.rs file. I think keeping it in the same file will also force us to make sure that we apply the same changes to Windows when we make changes to catnap::socket.

Perhaps a compromise would be pulling the Windows and Linux code out into functions, rather than leaving them inline?

iyzhang (approved):
Thanks for doing this Anand

ppenna (approved):
My only concern was that I could miss anything from the diff on the sockets layer, but @iyzhang has double checked that.

src/rust/catnap/socket.rs
anandbonde: We discussed offline and decided to continue with the current change.

PR #922, opened 2023-08-30
[contrib] Updating Rust Toolchain

ppenna:
## Description

This PR updates our Rust toolchain to the latest version.

Note that I had to fix the way that we were accessing the socket address data structure. The previous code was generating misaligned accessed and could lead to UB.


iyzhang (approved):
LGTM

anandbonde (approved):
Looks good.

PR #926, opened 2023-09-01
[runtime] New shared object for shared state

iyzhang:
This PR introduces a ref counted shared object that can be used for shared state without dynamic borrowing. The object lives as long as there are references (as there is an Rc inside); however it allows both multiple read and write references to the shared object. This object can be safely used as long as the code among which it is shared (e.g., different libOSes, different coroutines within the same libOS) never run concurrently. As Demikernel is currently single threaded and our scheduler can only run one coroutine at a time, this abstraction is safe to use throughout the code for shared references to libOSes, queues and runtimes. This data structure will continue to be safe if we use a multi-kernel architecture for multi-cores and run one instance of the libOSes per core and do not share across cores. If we wanted to share state across cores, a more complex data structure with locking would be required anyway.

### DIFF
--- a/src/rust/catloop/mod.rs
+++ b/src/rust/catloop/mod.rs
@@ -244,7 +244,7 @@ impl CatloopLibOS {
     /// the accept succeeds or fails.
     async fn accept_coroutine(
         qd: QDesc,
-        runtime: DemiRuntime,
+        mut runtime: DemiRuntime,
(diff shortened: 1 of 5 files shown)


ppenna (approved):

anandbonde (approved):
Looks good.

PR #927, opened 2023-09-01
[catloop] Ensure Accepting in `accept`

ppenna:
## Description

This PR introduces a check for the accepting state on the `accept()` async path.


anandbonde (approved):
Looks good.

PR #929, opened 2023-09-02
[catnap] Remove all Rc<RefCell> and replace with SharedObject

iyzhang:
This PR removes all of the dynamic borrows in Catnap and replaces them with our new SharedObject abstraction. We rename the shared state objects to SharedCatnapLibOS and SharedCatnapQueue to indicate that they are shared across coroutines.


anandbonde (approved):
Looks good.

PR #930, opened 2023-09-06
[scheduler] Removing the `deschedule()` function

### LINKED ISSUE #886: Remove deprecated function `deschedule()`
## Description
Warnings due to deprecated functions was causing issues in Bing's build integration with Demikernel.

## Temporary solution until this issue is fixed
Deprecated tag removed.
FIXME comment added in place.

Partially fixes: #878

carvalhof:
This PR closes #886, replacing the `deschedule()` calls to `scheduler.remove()`. For this, we update some functions of TCP in `inetstack` (998cecfa93557a8548259a5fea3401f753b56446, 4155df8cbaf43d0fb5d191368c5c141b0e1da5c6, and b613d1b79aa2ffc6a848c5759d954cbb54850f3c) and in `set_socket` function (eb3073828e07abc21e1366d10d243962a9ba45d6) to avoid memory leak.

Finally, in fd00700151f27867c0e0a60aebc2bb721f810e89, we remove the `deschedule()` function.

### DIFF
--- a/src/rust/inetstack/protocols/tcp/queue.rs
+++ b/src/rust/inetstack/protocols/tcp/queue.rs
@@ -10,7 +10,13 @@ use crate::runtime::{
     queue::IoQueue,
     QType,
 };
-use ::std::any::Any;
+use ::std::{
+    any::Any,
+    mem::{
+        swap,
+        forget,
+    },
+};
 
 //======================================================================================================================
 // Structures
@@ -43,8 +49,11 @@ impl<const N: usize> TcpQueue<N> {
     }
 
     /// Set underlying TCP socket data structure.
-    pub fn set_socket(&mut self, s: Socket<N>) {
-        self.socket = s;
+    pub fn set_socket(&mut self, mut s: Socket<N>) {
+        swap(&mut s, &mut self.socket);
+
+        // Don't run the Socket destructor at this moment.
+        forget(s);
     }
 }
 
--- a/src/rust/inetstack/protocols/tcp/active_open.rs
+++ b/src/rust/inetstack/protocols/tcp/active_open.rs
@@ -330,6 +330,10 @@ impl<const N: usize> ActiveOpenSocket<N> {
 
 impl<const N: usize> Drop for ActiveOpenSocket<N> {
     fn drop(&mut self) {
-        self.handle.deschedule();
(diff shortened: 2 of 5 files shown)


src/rust/inetstack/protocols/tcp/queue.rs
```diff
@@ -43,8 +49,11 @@ impl<const N: usize> TcpQueue<N> {
     }
 
     /// Set underlying TCP socket data structure.
-    pub fn set_socket(&mut self, s: Socket<N>) {
-        self.socket = s;
+    pub fn set_socket(&mut self, mut s: Socket<N>) {
+        swap(&mut s, &mut self.socket);
+
+        // Don't run the Socket destructor at this moment.
+        forget(s);
```
iyzhang, line 56: Instead of doing this, maybe we should more explicitly remove the Task from the scheduler, instead of in the drop function? It makes it hard to debug when there are implicit drops that manipulate the scheduler because it's hard to find where all references are dropped.

src/rust/inetstack/protocols/tcp/queue.rs
```diff
@@ -43,8 +49,11 @@ impl<const N: usize> TcpQueue<N> {
     }
 
     /// Set underlying TCP socket data structure.
-    pub fn set_socket(&mut self, s: Socket<N>) {
-        self.socket = s;
+    pub fn set_socket(&mut self, mut s: Socket<N>) {
+        swap(&mut s, &mut self.socket);
```
iyzhang, line 53: What's the reason for using swap here?

iyzhang (approved):
Thanks a ton for doing this! It is great and I'm so happy to get rid of that function. Just a few questions inline

src/rust/inetstack/protocols/tcp/queue.rs
ppenna: If I understood correctly, he is doing this because we explicitly want to skip the destructor on the next statement.

src/rust/inetstack/protocols/tcp/queue.rs
ppenna: Adding more to this... Why do we want to skip the destructor?

iyzhang:
Closing this PR because it is no longer necessary after PR #977

PR #933, opened 2023-09-12
[demikernel] Enhancement: generalize libOS APIs and C bindings from SocketAddrV4 to SocketAddr

kyleholohan:
This PR is a largely mechanical change for libOS APIs and C bindings to generalize from `SocketAddrV4` to `SocketAddr`, as described in #204 and elaborated in other places such as #413. The non-mechanical portions of this PR include:
- Updating bindings.rs to use socket2 to convert between C `sockaddr`s and `SocketAddr`; see new method `sockaddr_to_socketaddr` to replace `sockaddr_to_socketaddr4` (note latest socket2 has updated this code to be more performant)
- Adding a test case to the bindings.rs test
- Adding the method `crate::runtime::network::unwrap_socketaddr`, as well as call sites in all libOSes, to provide standard error semantics when expecting IPv4 (as all libOSes currently do)
- Removing unused platform abstractions
- Adding the TCP test `tcp_bad_bind`, to test IPv6 binding, as well as a few other cases.
- Updating examples to parse full SocketAddrs (which will fail for now if an IPv6 is parsed)
- Adding .gitignore for Visual Studio (because some habits die hard)

Issue #204, opened 2022-03-29
[ip] Rely on `SocketAddr` `enum` Rather Than `SocketAddrV4` `struct`


ppenna:
Description
========

Catnip uses the ~~`Ipv4Endpoint`~~ `SocketAddrV4` structure to refer to remote and local endpoints. While this is fine when supporting IPv4 addresses only, once we add support to IPv6 it is make the stack a mess. Therefore, we should change the stack so as to refer to local and remote endpoints using a more generic type, like ~~`IpEndpoint`~~`SocketAddr`.

ppenna:
@anandbonde are we done with this?

PR #936, opened 2023-09-19
[demikernel] Enhancement: generalize libOS APIs and C bindings from SocketAddrV4 to SocketAddr

kyleholohan:
This PR is a largely mechanical change for libOS APIs and C bindings to generalize from `SocketAddrV4` to `SocketAddr`, as described in #204 and elaborated in other places such as #413. This PR superseded #933. The non-mechanical portions of this PR include:
- Updating bindings.rs to use socket2 to convert between C `sockaddr`s and `SocketAddr`; see new method `sockaddr_to_socketaddr` to replace `sockaddr_to_socketaddr4` (note latest socket2 has updated this code to be more performant)
- Adding a test case to the bindings.rs test
- Adding the method `crate::runtime::network::unwrap_socketaddr`, as well as call sites in all libOSes, to provide standard error semantics when expecting IPv4 (as all libOSes currently do)
- Removing unused platform abstractions
- Adding the TCP test `tcp_bad_bind`, to test IPv6 binding, as well as a few other cases.
- Updating examples to parse full SocketAddrs (which will fail for now if an IPv6 is parsed)
- Adding .gitignore for Visual Studio (because some habits die hard)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ libc = "0.2.147"
 log = "0.4.20"
 rand = { version = "0.8.5", features = ["small_rng"] }
 slab = "0.4.9"
+socket2 = "0.4.7"
 yaml-rust = "0.4.5"
 x86 = "0.52.0"
 
@@ -53,8 +54,6 @@ windows = { version = "0.44.0", features = [
     "Win32_Foundation",
     "Win32_Networking_WinSock",
 ] }
-# Provides the Rust socket API for Windows.
-socket2 = "0.4.7"
 
 #=======================================================================================================================
 # Targets
--- a/src/rust/demikernel/bindings.rs
+++ b/src/rust/demikernel/bindings.rs
@@ -10,23 +10,7 @@ use crate::{
         name::LibOSName,
         LibOS,
     },
-    pal::{
-        constants::{
-            AF_INET,
-            AF_INET6,
-        },
-        data_structures::{
-            AddressFamily,
-            SockAddrIn,
-            SockAddrIn6,
-            Socklen,
-        },
-        functions::{
-            get_addr_from_sock_addr_in,
-            get_addr_from_sock_addr_in6,
-            get_scope_id_from_sock_addr_in6,
-        }
-    },
+    pal::data_structures::Socklen,
     runtime::{
         fail::Fail,
         logging,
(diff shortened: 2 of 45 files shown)


ppenna (approved):
LGTM. Thanks for working on this.

ppenna:
Adding @anandbonde as second reviewer.

src/rust/demikernel/bindings.rs
```diff
@@ -813,45 +809,101 @@ fn do_syscall<T>(f: impl FnOnce(&mut LibOS) -> T) -> Result<T, Fail> {
...
 
-    // SocketAddrV4: 127.0.0.1:80
-    #[cfg(target_os = "linux")]
-    let saddr: crate::pal::data_structures::SockAddr = {
-        sockaddr {
-            sa_family: crate::pal::constants::AF_INET_VALUE as u16,
-            sa_data: [0, 80, 127, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0],
-        }
+    // Read up to size bytes from saddr into a SockAddrStorage, the type which socket2 can use.
+    let mut storage: mem::MaybeUninit<libc::sockaddr_storage> = mem::MaybeUninit::<SockAddrStorage>::zeroed();
```
anandbonde, line 825: Should this datastructure come from the PAL instead? This would probably not compile on Windows.

Cargo.toml
```diff
@@ -38,6 +38,7 @@ libc = "0.2.147"
 log = "0.4.20"
 rand = { version = "0.8.5", features = ["small_rng"] }
 slab = "0.4.9"
+socket2 = "0.4.7"
```
anandbonde, line 41: Currently, this crate is only used on Linux. Placing it here would download it while building on Linux as well.

anandbonde (changes requested):
Thanks for doing this change! I left a few comments.

src/rust/demikernel/bindings.rs
kyleholohan: Good catch; thanks. I let VSCode fill this in, and I guess it "sees through" the typedef.

Cargo.toml
kyleholohan: Yes, that's intentional. I'm using the SockAddr -> SocketAddr functionality provided by socket2. Please advise if you want something different.

anandbonde: Oh, I see. In that case, this looks good to me. Please proceed with this change.

anandbonde (approved):
Looks good.

PR #937, opened 2023-09-25
[demikernel] Localize packing compiler directives for Windows builds

anandbonde:
Localize packing compiler directives for Windows builds. If this is not done, the packing directives will remain active for other build targets as well. So, to avoid this in integration builds (e.g., Bing), localizing the directives is the best option.


ppenna (approved):
LGTM

iyzhang (approved):

PR #942, opened 2023-10-03
[catloop] Move to shared object

iyzhang:
This PR moves catloop to our new shared object abstraction. It also removes the need for a CatloopRuntime to store the state (just the ephemeral port allocator) because we can now mutate CatloopLibOS directly.


src/rust/catloop/mod.rs
```diff
@@ -564,6 +559,39 @@ impl CatloopLibOS {
...
+        if let Some(port) = port {
+            self.ephemeral_ports.alloc_port(port)?;
+            Ok(None)
+        } else {
+            Ok(Some(self.ephemeral_ports.alloc_any()?))
+        }
+    }
+
+    /// Releases an ephemeral `port`.
+    pub fn free_ephemeral_port(&mut self, port: u16) -> Result<(), Fail> {
```
ppenna, line 574: Do we need this to be public?

src/rust/catloop/mod.rs
```diff
@@ -101,13 +104,22 @@ impl CatloopLibOS {
     pub fn new(config: &Config, runtime: SharedDemiRuntime) -> Self {
         #[cfg(feature = "profiler")]
         timer!("catloop::new");
+        let mut rng: SmallRng = SmallRng::from_entropy();
```
ppenna, line 107: On debug mode, this should be set to a fixed number, so we have some determinism?

ppenna (approved):
Added some changes. Please review before merging.

PR #949, opened 2023-10-03
[catmem] Enhancement: Move to shared object queues

iyzhang:
This PR moves CatmemQueue to the shared object abstraction


ppenna (approved):
LGTM

PR #955, opened 2023-10-05
[catloop] Warn Backlog Length Listen

### LINKED ISSUE #697: [catloop] Backlog Length Is Completely Ignored
## Description

Catloop LibOS currently ignores the backlog informed on `listen()`.

## Expected Behavior

The backlog length informed on `listen` should be considered on the establishment connection protocol. If the queue of pending connections grows larger than this value, new connection requests should be dropped.

## Code Snippet

Note that backlog length is never read / stored.

https://github.com/demikernel/demikernel/blob/c4fbd4247b3046dd62e10693753090bfe832884d/src/rust/catloop/mod.rs#L214-L242

ppenna:
## Description

This PR closes #697.

We currently ignore the backlog length and this PR improves existing code by keeping track of pending connections and triggering logging if a passive socket is too busy.

Ideally we would send `ECONNREFUSED` back to clients, but the current connection handshake protocol does not support that. It will keep queueing requests anyways.

### DIFF
--- a/src/rust/catloop/socket.rs
+++ b/src/rust/catloop/socket.rs
@@ -7,6 +7,7 @@
 
 use crate::{
     catmem::SharedCatmemLibOS,
+    pal,
     runtime::{
         fail::Fail,
         memory::DemiBuffer,
@@ -72,6 +73,10 @@ pub struct Socket {
     local: Option<SocketAddrV4>,
     /// The remote address to which the socket is connected.
     remote: Option<SocketAddrV4>,
+    /// Maximum backlog length for passive sockets.
+    backlog: usize,
+    /// Pending connect requests for passive sockets.
+    pending: usize,
 }
 
 //======================================================================================================================
@@ -87,6 +92,8 @@ impl Socket {
             catmem_qd: None,
             local: None,
             remote: None,
+            backlog: 1,
+            pending: 0,
         })
     }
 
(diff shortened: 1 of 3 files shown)


ppenna:
@iyzhang you may keep this issue in mind if you were to redesign the connection establishment protocol.

src/rust/catloop/socket.rs
```diff
@@ -87,6 +92,8 @@ impl Socket {
             catmem_qd: None,
             local: None,
             remote: None,
+            backlog: 1,
```
iyzhang, line 95: Does it make sense for this to default to 1? Is that the default on Linux or something?

src/rust/catloop/socket.rs
```diff
@@ -149,16 +162,34 @@ impl Socket {
         let catmem_qd: QDesc = self.catmem_qd.expect("Must be a catmem queue in this state.");
         loop {
             self.state.may_accept()?;
+
+            // Check if backlog is full.
+            if self.pending == self.backlog {
```
iyzhang, line 167: This should probably be a greater than since we will keep queueing anyway.

iyzhang (approved):
LGTM

iyzhang (changes requested):
Bumping the pending number on accept is not correct because the server is going to dequeue that connection request, not enqueue it, so the pending number does not currently reflect the backlog. In order to reflect the backlog, we would have to increase the pending number when the request arrives in the queue.

iyzhang (approved):
Thanks for explaining the backlog length to me!

PR #960, opened 2023-10-09
[inetstack] Don't Panic if ARP Query Fail

ppenna:
## Description

This PR closes https://github.com/microsoft/demikernel/issues/220.

Instead of panicking, if ARP query fails, we fail gracefully.

### DIFF
--- a/src/rust/inetstack/protocols/tcp/peer.rs
+++ b/src/rust/inetstack/protocols/tcp/peer.rs
@@ -655,24 +655,36 @@ impl<const N: usize> Inner<N> {
         Ok(())
     }
 
-    pub fn send_rst(&self, local: &SocketAddrV4, remote: &SocketAddrV4) -> Result<(), Fail> {
-        // TODO: Make this work pending on ARP resolution if needed.
-        let remote_link_addr = self
-            .arp
-            .try_query(remote.ip().clone())
-            .ok_or(Fail::new(libc::EINVAL, "detination not in ARP cache"))?;
-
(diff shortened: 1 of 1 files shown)


src/rust/inetstack/protocols/tcp/peer.rs
```diff
@@ -655,24 +655,36 @@ impl<const N: usize> Inner<N> {
...
-        self.rt.transmit(Box::new(segment));
+
+        // Create a RST segment.
+        let segment: TcpSegment = {
+            let mut tcp_hdr: TcpHeader = TcpHeader::new(local.port(), remote.port());
+            tcp_hdr.rst = true;
+            TcpSegment {
+                ethernet2_hdr: Ethernet2Header::new(dst_link_addr, self.local_link_addr, EtherType2::Ipv4),
+                ipv4_hdr: Ipv4Header::new(local.ip().clone(), remote.ip().clone(), IpProtocol::TCP),
+                tcp_hdr,
```
iyzhang, line 679: For consistency, probably just stick TcpHeader::new in here too instead of using a separate variable. Otherwise, create variables for all of them.

iyzhang (approved):

src/rust/inetstack/protocols/tcp/peer.rs
ppenna: It is not possible, because we need to change the RST, SeqNum and AckNum fields of the CP header.

PR #961, opened 2023-10-10
[catloop] Enhancement: dedup requests

### LINKED ISSUE #947: [catloop] Server should deduplicate requests on connecting
## Context

After a timeout, clients will resend connection requests. The server has no way of deduplicating these requests so sometimes accepts a connection twice, even though the client is only waiting on a single connection. This can wedge the server because it is waiting for two response back from the client.

## Proposed Solution

Add a request ID to each connect request, which the server will send back to the client indicating the completion of the request. The client should only send one request at a time and the server will retry if it does not get a response from the client.

iyzhang:
This PR closes #947

### DIFF
--- a/src/rust/catloop/socket.rs
+++ b/src/rust/catloop/socket.rs
@@ -31,7 +31,13 @@ use crate::{
         Yielder,
     },
 };
+use ::rand::{
+    rngs::SmallRng,
+    RngCore,
+    SeedableRng,
+};
 use ::std::{
+    collections::HashSet,
     mem,
     net::{
         Ipv4Addr,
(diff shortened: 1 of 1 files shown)


src/rust/catloop/socket.rs
```diff
@@ -47,15 +53,7 @@ use ::std::{
...
-
-/// Magic payload used to identify connect requests.  It must be a single
-/// byte to ensure atomicity while keeping the connection establishment
-/// protocol. The rationale for this lies on the fact that a pipe in Catmem
-/// LibOS operates atomically on bytes. If we used a longer byte sequence,
-/// we would need to introduce additional logic to make sure that
-/// concurrent processes would not be enabled to establish a connection, if
-/// they sent connection bytes in an interleaved, but legit order.
-pub const MAGIC_CONNECT: u8 = 0x1b;
+const SEED: u64 = 95;
```
ppenna, line 56: Could we have a more detailed named for this?

src/rust/catloop/socket.rs
```diff
@@ -424,13 +427,14 @@ impl Socket {
...
-    // Issue pop. Each magic connect represents a separate connection request, so we always bound the pop.
-    let qt: QToken = catmem.pop(catmem_qd, Some(mem::size_of_val(&MAGIC_CONNECT)))?;
+// Gets the next connection request.
+async fn pop_request_id(
+    catmem: &mut SharedCatmemLibOS,
+    catmem_qd: QDesc,
+    yielder: &Yielder,
+) -> Result<RequestId, Fail> {
+    // Issue pop. No need to bound the pop because we've quantized it already in the concurrent ring buffer.
+    let qt: QToken = catmem.pop(catmem_qd, None)?;
```
ppenna, line 437: We still need a bounded pop, right?

ppenna (approved):
LGTM

PR #962, opened 2023-10-10
[inetstack] SeqNum and AckNum for RST Segments

ppenna:
## Description

This PR partially addresses https://github.com/microsoft/demikernel/issues/235.

### DIFF
--- a/src/rust/inetstack/protocols/tcp/peer.rs
+++ b/src/rust/inetstack/protocols/tcp/peer.rs
@@ -615,26 +615,36 @@ impl<const N: usize> Inner<N> {
         }
     }
 
+    /// Processes an incoming TCP segment.
     fn receive(&self, ip_hdr: &Ipv4Header, buf: DemiBuffer) -> Result<(), Fail> {
-        let (mut tcp_hdr, data) = TcpHeader::parse(ip_hdr, buf, self.tcp_config.get_rx_checksum_offload())?;
+        let (mut tcp_hdr, data): (TcpHeader, DemiBuffer) =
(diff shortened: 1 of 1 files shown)


src/rust/inetstack/protocols/tcp/peer.rs
```diff
@@ -596,26 +596,36 @@ impl<const N: usize> Inner<N> {
...
         }
 
-        // grab the queue descriptor based on the incoming.
+        // Retrieve the queue descriptor based on the incoming segment.
         let &qd: &QDesc = match self.addresses.get(&SocketId::Active(local, remote)) {
             Some(qdesc) => qdesc,
             None => match self.addresses.get(&SocketId::Passive(local)) {
                 Some(qdesc) => qdesc,
-                None => return Err(Fail::new(libc::EBADF, "Socket not bound")),
+                None => {
```
iyzhang, line 618: Should we also send a reset here? It seems like this is a similar case to inactive or closed?

iyzhang (approved):
LGTM, maybe put in some more details about which scenario in the issue that you have covered.

src/rust/inetstack/protocols/tcp/peer.rs
ppenna: I am not sure which case you are talking about.

ppenna: Not really. This is the case where we have received a segment for a connection that is not established or not in its way to be established. We just ignore the packet and remote will re-transmit if that is a packet that arrived before the SYN.

PR #963, opened 2023-10-10
[inetstack] Add Checks for Bind

ppenna:
## Description

Add missing checks for wildcard and port number for the inetstack.

## Related Issues

- https://github.com/microsoft/demikernel/issues/189
- https://github.com/microsoft/demikernel/issues/582

### DIFF
--- a/src/rust/inetstack/protocols/tcp/peer.rs
+++ b/src/rust/inetstack/protocols/tcp/peer.rs
@@ -170,31 +170,38 @@ impl<const N: usize> TcpPeer<N> {
         Ok(new_qd)
     }
 
-    pub fn bind(&self, qd: QDesc, mut addr: SocketAddrV4) -> Result<(), Fail> {
+    pub fn bind(&self, qd: QDesc, local: SocketAddrV4) -> Result<(), Fail> {
+        // Check if we are binding to the wildcard address.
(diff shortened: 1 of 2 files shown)


src/rust/inetstack/protocols/tcp/peer.rs
```diff
@@ -170,31 +170,38 @@ impl<const N: usize> TcpPeer<N> {
...
-        // Check if address is already bound.
-        for (socket_id, _) in &inner.addresses {
-            match socket_id {
-                SocketId::Passive(local) | SocketId::Active(local, _) if *local == addr => {
-                    return Err(Fail::new(libc::EADDRINUSE, "address already in use"))
-                },
-                _ => (),
-            }
+        // Check if we are binding to the wildcard port.
+        // FIXME: https://github.com/demikernel/demikernel/issues/582
```
iyzhang, line 183: What is the reason for not allocating an ephemeral port here?

iyzhang (approved):

src/rust/inetstack/protocols/tcp/peer.rs
ppenna: To match same logic of other LibOSes. Eventually it will be unified.

PR #956, opened 2023-10-05
[runtime] Unify Retry Test

ppenna:
## Description

This PR closes https://github.com/microsoft/demikernel/issues/954.

PR #967, opened 2023-10-11
[demikernel] Fix Windows build errors

anandbonde:
The Windows build broke after a recent commit to the dev branch. This PR fixes it.


src/rust/catnap/socket.rs
```diff
@@ -28,13 +27,16 @@ use crate::{
...
+use crate::{
+    pal::{
+        data_structures::{
+            SockAddr,
+            SockAddrIn,
+            Socklen,
+        },
+        linux,
     },
-    linux,
```
iyzhang, line 37: Why does only the linux code need the DemiRuntime?

iyzhang (approved):
LGTM

src/rust/catnap/socket.rs
anandbonde: It was a recent change that was checked in: [[runtime] Unify Retry Test by ppenna · Pull Request #956 · microsoft/demikernel (github.com)](https://github.com/microsoft/demikernel/pull/956). We should do this for Windows as well. We will handle this as a separate change.

PR #977, opened 2023-10-19
[inetstack] Move from futures to async functions

iyzhang:
This PR adds a new data structure, an AsyncQueue, which will yield to the Demikernel scheduler if there is nothing in the queue and wake blocked coroutines when there is something to read.

The PR then uses this queue to accept incoming sockets.


src/rust/collections/async_queue.rs
```diff
@@ -0,0 +1,63 @@
...
+        YielderHandle,
+    },
+};
+use ::std::collections::VecDeque;
+
+//======================================================================================================================
+// Structures
+//======================================================================================================================
+
+/// This data structure implements an unbounded asynchronous queue that is hooked into the Demikernel scheduler. On pop, if the queue is empty, the coroutine will yield until there is data to be read.
```
ppenna, line 20: Missing line break?

src/rust/collections/async_queue.rs
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+//======================================================================================================================
+// Imports
+//======================================================================================================================
+use crate::{
```
ppenna, line 7: Missing space.

src/rust/collections/async_queue.rs
```diff
@@ -0,0 +1,63 @@
...
+/// This data structure implements an unbounded asynchronous queue that is hooked into the Demikernel scheduler. On pop, if the queue is empty, the coroutine will yield until there is data to be read.
+pub struct AsyncQueue<T> {
+    queue: VecDeque<T>,
+    waiters: VecDeque<YielderHandle>,
+}
+
+//======================================================================================================================
+// Associate Functions
+//======================================================================================================================
+impl<T> AsyncQueue<T> {
```
ppenna, line 29: Missing space

src/rust/collections/async_queue.rs
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    pub async fn pop(&mut self, yielder: Yielder) -> Result<T, Fail> {
+        match self.queue.pop_front() {
+            Some(item) => Ok(item),
+            None => {
+                let handle: YielderHandle = yielder.get_handle();
+                self.waiters.push_back(handle);
+                match yielder.yield_until_wake().await {
+                    Ok(()) => Ok(self.queue.pop_front().expect("Should only be woken with data")),
```
ppenna, line 44: Could we fail instead of panicking?

ppenna (approved):
Thanks for working on this.

src/rust/collections/async_queue.rs
iyzhang: Yup, done

ppenna (changes requested):
Taking out my approval for another review round.

src/rust/runtime/mod.rs
```diff
@@ -146,6 +146,16 @@ impl DemiRuntime {
...
 
+    /// Removes the background `coroutine` associated with `handle`. Since background coroutines do not return a result
+    /// there is no need to cast it.
+    pub fn remove_background_coroutine(&mut self, handle: &TaskHandle) {
+        // 1. Remove Task from scheduler.
+        let _: Box<dyn Task> = self
+            .scheduler
+            .remove(handle)
+            .expect("Removing task that does not exist (either was previously removed or never inserted");
+    }
```
ppenna, line 157: I understand that we panic here if and only if something is badly broken. It would be nice to have extra information just in case if that happens.

I understand that we have two ways for approaching this:
- Change the function signature to Fail, instead of panicking, and then dump the execution context in the caller's functions.
- Keep the function signature as it is, and print some information based on the `TaskHandle`.

I personally prefer the first option.

src/rust/collections/async_value.rs
```diff
@@ -0,0 +1,74 @@
...
+        match self.value.take() {
+            Some(item) => Ok(item),
+            None => {
+                let handle: YielderHandle = yielder.get_handle();
+                self.waiter = Some(handle);
+                match yielder.yield_until_wake().await {
+                    Ok(()) => match self.value.take() {
+                        Some(item) => Ok(item),
+                        None => {
+                            let cause = "Spurious wake up!";
```
ppenna, line 50: Missing type annotation here.

src/rust/collections/async_value.rs
```diff
@@ -0,0 +1,74 @@
...
+                            warn!("async_queue::pop(): {}", cause);
```
ppenna, line 51: Should this be `warn!("get(): {:?}", cause)`?

src/rust/collections/async_queue.rs
```diff
@@ -0,0 +1,80 @@
...
+            Some(item) => Ok(item),
+            None => {
+                let handle: YielderHandle = yielder.get_handle();
+                self.waiters.push_back(handle);
+                match yielder.yield_until_wake().await {
+                    Ok(()) => match self.queue.pop_front() {
+                        Some(item) => Ok(item),
+                        None => {
+                            let cause = "Spurious wake up!";
+                            warn!("async_queue::pop(): {}", cause);
```
ppenna, line 58: Should this be `warn!("get(): {:?}", cause)`? Note that the module name already gets printted in the log message.

src/rust/inetstack/protocols/tcp/established/ctrlblk.rs
```diff
@@ -115,44 +112,40 @@ struct Receiver {
...
+    pub async fn pop(&mut self, size: Option<usize>, yielder: Yielder) -> Result<DemiBuffer, Fail> {
         let buf: DemiBuffer = if let Some(size) = size {
-            let buf: &mut DemiBuffer = self.recv_queue.front_mut().expect("receive queue cannot be empty");
+            let mut buf: DemiBuffer = self.recv_queue.pop(yielder).await?;
             // Split the buffer if it's too big.
             if buf.len() > size {
                 buf.split_front(size)?
             } else {
-                self.recv_queue.pop_front().expect("receive queue cannot be empty")
+                // I think this was a bug before.
```
ppenna, line 134: I agree, but I don't we just add a debug-level message here? If we notice something bad we keep track of this.

src/rust/inetstack/test_helpers/runtime.rs
```diff
@@ -132,9 +154,11 @@ impl SharedTestRuntime {
 //==============================================================================
 
 impl<const N: usize> NetworkRuntime<N> for SharedTestRuntime {
+    /// Send a
```
ppenna, line 157: Missing comment?

src/rust/inetstack/protocols/tcp/tests/established.rs
```diff
@@ -53,60 +55,79 @@ fn cook_buffer(size: usize, stamp: Option<u8>) -> DemiBuffer {
...
+            frame.clone(),
+            sender.get_test_rig().get_link_addr(),
+            receiver.get_test_rig().get_link_addr(),
+            sender.get_test_rig().get_ip_addr(),
+            receiver.get_test_rig().get_ip_addr(),
+            window_size,
+            seq_no,
+            ack_num,
+        )?;
+        // Just drop retranmissions.
```
ppenna, line 101: Typo?

ppenna (approved):
LGTM

PR #980, opened 2023-10-25
[runtime] Remove all Refcell from scheduler

iyzhang:
This PR removes all refcells from the scheduler, as the scheduler is always part of a shared object now and no longer needs to be cloneable.

### DIFF
--- a/src/rust/scheduler/scheduler.rs
+++ b/src/rust/scheduler/scheduler.rs
@@ -31,16 +31,10 @@ use ::rand::{
     SeedableRng,
 };
 use ::std::{
-    cell::{
-        Ref,
-        RefCell,
-        RefMut,
-    },
     collections::HashMap,
     future::Future,
     pin::Pin,
     ptr::NonNull,
-    rc::Rc,
     task::{
         Context,
         Poll,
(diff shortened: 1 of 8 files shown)


src/rust/scheduler/scheduler.rs
```diff
@@ -116,59 +107,53 @@ impl Scheduler {
...
     /// Insert a new task into our scheduler returning a handle corresponding to it.
-    pub fn insert<F: Task>(&self, future: F) -> Option<TaskHandle> {
+    pub fn insert<F: Task>(&mut self, future: F) -> Option<TaskHandle> {
         self.panic_if_too_many_tasks();
 
         let task_name: String = future.get_name();
         // The pin slab index can be reverse-computed in a page index and an offset within the page.
-        let pin_slab_index: usize = self.tasks.borrow_mut().insert(Box::new(future))?;
+        let pin_slab_index: usize = self.tasks.insert(Box::new(future))?;
+        let task_id = self.get_new_task_id(pin_slab_index);
```
ppenna, line 131: Missing type annotation.

ppenna (approved):

Issue #884, opened 2023-08-20
Remove deprecated function `remove_dropped_tasks()`


anandbonde:
## Description
Warnings due to deprecated functions was causing issues in Bing's build integration with Demikernel.

## Temporary solution until this issue is fixed
Deprecated tag removed.
FIXME comment added in place.

Partially fixes: #878

ppenna:
This issue was fixed.

PR #981, opened 2023-10-26
[runtime] Remove drop bits from scheduler

### LINKED ISSUE #723: [runtime] Reorganize scheduler and re-evaluated needed bits in Waker pages
## Context

Currently, our Waker pages have 3 bits per Task -- notified, completed, dropped -- which are not all necessary. For example, the Task also keeps a completed flag, which could be checked instead of using a Waker bit. There is also one extra bit that we could reuse for an active flag.

## Proposed Solution

1. Figure out whether the completed flag is strictly necessary or an optimization for checking if the coroutine in a Task has completed.
2. Add an active flag, so we can stop using the slab offset to indicate which waker bits are in use. We would then need to use a linear search to find a free waker bit and store the offset somewhere (since we can't use the slab offset any more). That would also allow us to get rid of the PinSlab data structure for a HashMap<Pin<>> instead.

iyzhang:
We no longer drop running coroutines through the task handle but always explicitly remove them from the scheduler instead. This PR closes #723 and #884

### DIFF
--- a/src/rust/scheduler/scheduler.rs
+++ b/src/rust/scheduler/scheduler.rs
@@ -96,10 +96,6 @@ impl Scheduler {
             let (waker_page_index, waker_page_offset) = self.get_waker_page_index_and_offset(pin_slab_index);
             (&waker_page_refs[waker_page_index], waker_page_offset)
         };
-        assert!(
-            !waker_page_ref.was_dropped(waker_page_offset),
-            "Task was previously dropped"
-        );
         waker_page_ref.clear(waker_page_offset);
         if let Some(task) = self.tasks.borrow_mut().remove_unpin(pin_slab_index) {
             trace!(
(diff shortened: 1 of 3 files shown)


src/rust/scheduler/scheduler.rs
```diff
@@ -212,9 +208,8 @@ impl Scheduler {
     pub fn poll(&self) {
         let num_waker_pages = self.get_num_waker_pages();
         for waker_page_index in 0..num_waker_pages {
-            let (notified_offsets, dropped_offsets) = self.get_offsets_for_ready_tasks(waker_page_index);
+            let notified_offsets = self.get_offsets_for_ready_tasks(waker_page_index);
             self.poll_notified_tasks(waker_page_index, notified_offsets);
```
ppenna, line 212: Missing type annotation.

ppenna (approved):

PR #983, opened 2023-10-27
[collections] Enhancement: refactor RingBuffer to operate more like a pipe

kyleholohan:
This PR includes the following enhancements:
- Refactors the `RingBuffer` class to operate more like a pipe: once instantiated, it may be `open`ed to produce a `(RingProducer, RingConsumer)` pair. These classes have the respective enqueue/dequeue operations exposed.
- Adds a type class constraint for the size pointers in `RingBuffer`. As a compelling example, XDP requires ring buffers to use 32-bit sizes.
- Creates the trait `IntSize` which allows `RingBuffer` to juggle various size types (no wider than usize).
- Changes the memory ordering to release-acquire semantics, which are fully sufficient for a SPSC model.
- Re-exposes the ring as a public export from collections mod.

### DIFF
--- a/src/rust/collections/mod.rs
+++ b/src/rust/collections/mod.rs
@@ -4,8 +4,6 @@
 pub mod async_queue;
 pub mod async_value;
 pub mod intrusive;
-mod raw_array;
-pub mod ring;
 
 cfg_if! {
     if #[cfg(feature = "catmem-libos")] {
--- a/src/rust/collections/mod.rs
+++ b/src/rust/collections/mod.rs
@@ -4,6 +4,8 @@
 pub mod async_queue;
 pub mod async_value;
 pub mod intrusive;
+mod raw_array;
+pub mod ring;
 
 cfg_if! {
     if #[cfg(feature = "catmem-libos")] {
--- a/src/rust/collections/ring.rs
+++ b/src/rust/collections/ring.rs
@@ -206,6 +206,7 @@ where
 
     /// Open the ring buffer for production and consumption. This method returns a pair of objects which allow the
     /// circular buffer to be interacted with like a single-producer-single-consumer pipe.
+    #[allow(unused)]
     pub fn open<'a>(&'a mut self) -> (RingProducer<'a, T, S>, RingConsumer<'a, T, S>) {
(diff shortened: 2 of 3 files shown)


src/rust/collections/mod.rs
```diff
@@ -4,6 +4,8 @@
 pub mod async_queue;
 pub mod async_value;
 pub mod intrusive;
+mod raw_array;
+pub mod ring;
```
ppenna, line 8: If you move this to all LibOses, you should remove it from the if statement that follows.

src/rust/collections/ring.rs
ppenna, line 1: For now, you may want all unused functions with `#[allow(unused)]`.

src/rust/collections/ring.rs
```diff
@@ -141,130 +204,226 @@ where
...
-        #[cfg(feature = "profiler")]
-        timer!("collections::ring::enqueue");
-        loop {
-            if self.try_enqueue(item).is_ok() {
-                break;
-            }
-        }
+    fn new(ring: &'a RingBuffer<T, S>) -> RingProducer<'a, T, S> {
+        let mut ret: Self = Self {
+            ring: ring,
```
ppenna, line 269: You can simply using 'ring'.

src/rust/collections/ring.rs
```diff
@@ -141,130 +204,226 @@ where
...
-        };
+    /// Start an enqueue operation, consuming self and returning a new RingEnqueue. If there are not enough free slots
+    /// in the ring buffer to enqueue `count` `T`'s, the operation will return an error.
+    pub fn try_enqueue(&mut self, val: T) -> Result<(), Fail> {
+        self.validate_enqueue(S::from(1u8))?;
 
-        // Commit read.
-        self.set_front((front_cached + 1) & self.mask);
+        let mask = self.ring.mask;
+        let start = S::to_usize(self.cached_back & mask);
```
ppenna, line 309: Missing type annotations.

ppenna (approved):
Looks great. Thanks for working on this.

I left some minor suggestions.

PR #984, opened 2023-10-27
[runtime] Enhancement: Move timer to explicit yielding and remove Rc construct

iyzhang:
This PR makes two updates to the Timer data structure. First, it adds support for the SharedObject abstraction and removes the Inner construct. Next, it adds yielders, so that coroutines blocked on timers can be woken and cancelled (for example if the associated queue is closed). It also replaces the pairing heap with a Rust binary heap. Since the pairing heap was unsafe, made it a little more complicated to use with the new SharedObject abstraction, so I chose to replace it with the standard Rust BinaryTree and remove the pairing heap from our codebase.


src/rust/inetstack/protocols/arp/peer.rs
```diff
@@ -172,7 +173,12 @@ impl<const N: usize> SharedArpPeer<N> {
                 // TODO: re-enable eviction once TCP/IP stack is fully functional.
                 // cache.clear();
             }
-            self.clock.wait(self.clock.clone(), Duration::from_secs(1)).await;
+            let yielder: Yielder = Yielder::new();
+
+            match self.clock.clone().wait(Duration::from_secs(1), yielder).await {
```
ppenna, line 178: I just spooted this hard-coded constant. Let's add it a name ;)

src/rust/inetstack/protocols/tcp/passive_open.rs
```diff
@@ -330,7 +330,12 @@ impl<const N: usize> SharedPassiveSocket<N> {
                 tx_checksum_offload: self.tcp_config.get_rx_checksum_offload(),
             };
             self.transport.transmit(Box::new(segment));
-            self.clock.wait(self.clock.clone(), handshake_timeout).await;
+            let clock_ref: SharedTimer = self.clock.clone();
+            let yielder: Yielder = Yielder::new();
+            if let Err(e) = clock_ref.wait(handshake_timeout, yielder).await {
+                self.ready.push_err(e);
+                return;
+            }
```
ppenna, line 338: This code is replicated everwhere. Should we make it a utility standalone function?

src/rust/runtime/timer.rs
```diff
@@ -5,148 +5,111 @@
...
+pub struct Timer {
     now: Instant,
-    heap: PairingHeap<TimerQueueEntry>,
-}
-
-pub struct Timer<P: TimerPtr> {
-    inner: RefCell<TimerInner>,
-    _marker: PhantomData<P>,
+    // Use a reverse to get a min heap.
+    heap: BinaryHeap<Reverse<TimerQueueEntry>>,
```
ppenna, line 44: Awesome

src/rust/runtime/timer.rs
```diff
@@ -5,148 +5,111 @@
...
-                }
-                inner.heap.remove(entry);
+    pub fn advance_clock(&mut self, now: Instant) {
+        assert!(self.now <= now);
+
+        while let Some(Reverse(entry)) = self.heap.peek() {
+            if now < entry.expiry {
+                break;
             }
+            let mut entry = self
```
ppenna, line 69: Missing type annotation

src/rust/runtime/timer.rs
```diff
@@ -264,18 +161,22 @@ mod tests {
         let mut ctx = Context::from_waker(noop_waker_ref());
         let mut now = Instant::now();
 
-        let timer = TimerRc(Rc::new(Timer::new(now)));
+        let mut timer = SharedTimer::new(now);
+        let timer_ref = timer.clone();
+        let yielder: Yielder = Yielder::new();
```
ppenna, line 166: Missing type annotations.

ppenna (approved):

src/rust/inetstack/protocols/tcp/passive_open.rs
iyzhang: Yes, I'm going to move the clock to the DemiRuntime and I will move this code as well. The yielder will need to be stored with the queue so we can cancel the coroutine more gracefully if the socket closes and there's cleanup work in the background thread.

PR #985, opened 2023-10-28
[catnip] Remove inner construct from memory management runtime

iyzhang:
This PR removes the remaining inner construct from Catnip. It was originally used to share the DPDK memory manager in Catnip but I have moved the entire DPDKRuntime to a shared object instead.

### DIFF
--- a/src/rust/catnip/runtime/memory/manager.rs
+++ b/src/rust/catnip/runtime/memory/manager.rs
@@ -34,7 +34,6 @@ use ::std::{
         self,
         NonNull,
     },
-    rc::Rc,
 };
 
 //==============================================================================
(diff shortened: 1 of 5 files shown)


src/rust/catnip/runtime/memory/manager.rs
```diff
@@ -74,10 +69,30 @@ pub struct MemoryManager {
...
+            config.get_max_body_size(),
+            config.get_body_pool_size(),
+            config.get_cache_size(),
+        )?;
 
         Ok(Self {
-            inner: Rc::new(Inner::new(memory_config)?),
+            config,
+            header_pool: header_pool,
+            body_pool: body_pool,
```
ppenna, line 95: Use shorthand structure initialization.

src/rust/catnip/runtime/mod.rs
```diff
@@ -103,12 +106,14 @@ pub struct DPDKRuntime {
     udp_config: UdpConfig,
 }
 
+#[derive(Clone)]
+pub struct SharedDPDKRuntime(SharedObject<DPDKRuntime>);
```
ppenna, line 110: Missing space.

ppenna (approved):

PR #986, opened 2023-10-28
[runtime] Move watched value to current architecture

iyzhang:
This PR does the following updates for watched value:
1. Moves to async functions instead of futures
2. Add support for our shared object abstraction
3. Removes the use of our intrusive linked list and replaces it with a VecDeque


src/rust/inetstack/protocols/tcp/established/mod.rs
```diff
@@ -93,10 +93,11 @@ impl<const N: usize> EstablishedSocket<N> {
...
 //======================================================================================================================
 
-impl<const N: usize> Drop for EstablishedSocket<N> {
-    fn drop(&mut self) {
-        if let Err(e) = self.runtime.remove_background_coroutine(&self.background) {
-            panic!("Failed to drop established socket (error={})", e);
-        }
-    }
-}
+// TODO: Uncomment this once we have proper resource clean up on asynchronous close.
```
ppenna, line 96: May we have an issue linked here?

ppenna (approved):
LGTM

src/rust/runtime/watched.rs
```diff
@@ -5,181 +5,98 @@
...
-
 pub struct WatchedValue<T> {
-    inner: RefCell<WatchedValueInner<T>>,
-}
-
-pub struct WatchFutureInner<'a, T> {
-    watch: &'a WatchedValue<T>,
-    wait_node: ListNode<WatchEntry>,
+    value: T,
+    waiters: VecDeque<YielderHandle>,
```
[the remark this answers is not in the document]
iyzhang: Hm yes I'm not sure why I thought that was needed. I will change to a Vec

src/rust/inetstack/protocols/tcp/established/mod.rs
iyzhang: Done

PR #989, opened 2023-11-01
[collections] Enhancement: Remove unnecessary linked list

iyzhang:
It is not necessary to use a linked list for this data structure either.


ppenna (approved):
Thanks

PR #990, opened 2023-11-02
[scheduler] Remove redundant waker page offset calculation

anandbonde:
The waker page offset does not need to be computed twice in this case.


iyzhang (approved):
Thanks for catching this!

PR #991, opened 2023-11-02
[catcollar] Move to SharedDemiRuntime

iyzhang:
This PR does some clean up inside the DemiRuntime needed to move the inetstack to a more object-oriented architecture. 
1. Move to a default constructor. This will let us add more data structures to the shared runtime, like the ephemeral port allocator without adding a lot of new code.
2. Move Catcollar to the shared runtime. This will let us get rid of the references to the io queue table and scheduler that do not go through the SharedDemiRuntime

### DIFF
--- a/src/rust/catcollar/mod.rs
+++ b/src/rust/catcollar/mod.rs
@@ -42,11 +42,9 @@ use crate::{
         network::unwrap_socketaddr,
         queue::{
             downcast_queue_ptr,
-            IoQueueTable,
             NetworkQueue,
             Operation,
             OperationResult,
-            OperationTask,
             QDesc,
             QToken,
             QType,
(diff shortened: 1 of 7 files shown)


src/rust/catcollar/mod.rs
```diff
@@ -86,10 +80,10 @@ use ::std::{
 
 /// Catcollar LibOS
 pub struct CatcollarLibOS {
-    /// Table of queue descriptors.
-    qtable: Rc<RefCell<IoQueueTable>>, // TODO: Move this into runtime module.
+    /// S
```
anandbonde, line 83: I think some text may have been accidentally deleted here.

anandbonde (approved):
Looks good.

src/rust/catcollar/mod.rs
iyzhang: Yup, thanks! I removed the TODO as well.

PR #992, opened 2023-11-02
[runtime] Enhancement: Move ephemeral port allocator into shared runtime

iyzhang:
This PR moves the ephemeral port allocator into the SharedDemiRuntime to avoid collisions when running more than one libOS. It also removes a lot of extraneous code and extra dependencies because the ephemeral port allocator is not in the inetstack any longer.

### DIFF
--- a/src/rust/runtime/network/ephemeral.rs
+++ b/src/rust/runtime/network/ephemeral.rs
@@ -3,28 +3,31 @@
 
 use crate::runtime::fail::Fail;
 use ::rand::prelude::{
+    SeedableRng,
     SliceRandom,
     SmallRng,
 };
 
-//==============================================================================
+//======================================================================================================================
 // Constants
-//==============================================================================
+//======================================================================================================================
 
 const FIRST_PRIVATE_PORT: u16 = 49152;
 const LAST_PRIVATE_PORT: u16 = 65535;
+/// Seed number of ephemeral port allocator.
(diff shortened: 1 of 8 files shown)


src/rust/catcollar/mod.rs
```diff
@@ -86,10 +80,10 @@ use ::std::{
 
 /// Catcollar LibOS
 pub struct CatcollarLibOS {
-    /// Table of queue descriptors.
-    qtable: Rc<RefCell<IoQueueTable>>, // TODO: Move this into runtime module.
+    /// S
```
anandbonde, line 83: Some text appears to be deleted here.

src/rust/runtime/network/ephemeral.rs
```diff
@@ -0,0 +1,246 @@
...
+
+        Ok(())
+    }
+}
+
+//======================================================================================================================
+// Trait Implementations
+//======================================================================================================================
+
+impl Default for EphemeralPorts {
```
anandbonde, line 95: There are multiple identical implementations for EphemeralPorts. Is it possible to reuse?

anandbonde (changes requested):

src/rust/runtime/network/ephemeral.rs
iyzhang: Ah, I've deleted it now.

anandbonde (approved):
Looks good.

PR #994, opened 2023-11-05
[runtime] Enhancement: Move clock to runtime

iyzhang:
This PR moves the Timer to the SharedDemiRuntime. This ensures that all libOSes share a single clock that tracks time. Note that only the inetstack currently uses this feature as the other libOSes do not use timeouts. However, we should move things like the Catloop retry mechanism to the Timer in the future. We should also remember to advance the clock on poll for all libOSes.


anandbonde (approved):
Looks good.

PR #995, opened 2023-11-06
[runtime] Move to shared demux table

iyzhang:
This PR moves the table that tracks address to queue descriptor mappings from each libOS to the shared DemiRuntime. This ensures that addresses are not reused across libOSes and provides a centralized way to demultiplex an incoming request from the transport into the appropriate queue, regardless of whether we are getting an IP address and port or a file descriptor.


src/rust/catloop/mod.rs
```diff
@@ -249,6 +259,7 @@ impl SharedCatloopLibOS {
         match result {
             Ok(new_queue) => {
                 let new_qd: QDesc = self.runtime.alloc_queue::<SharedCatloopQueue>(new_queue);
+                // TODO: insert into socket id to queue descriptor table?
```
anandbonde, line 262: Which scenario would be prevented if this is done? Duplicate entries, like on line 201?

src/rust/catloop/mod.rs
```diff
@@ -304,6 +315,7 @@ impl SharedCatloopLibOS {
 
         // Wait for connect operation to complete.
         match queue.do_connect(remote, &yielder).await {
+            // TODO: insert into socket id to queue descriptor table?
```
anandbonde, line 318: Which scenario would be prevented if this is done? Duplicate entries, like on line 201?

anandbonde (approved):
Looks good.

src/rust/catloop/mod.rs
iyzhang: If the application tries to later bind to an ephemeral port or some other port that has already been used for accepting or connecting, we should not allow that. However, for Catloop, I don't think connections have address pairs, they just use the server-side address and port (e.g., on the client, the queue only has a remote address and on the server, the queue only has a local address), so it's unclear what the right behavior is here.

PR #996, opened 2023-11-07
[runtime] Remove file descriptor from socket id data structure

iyzhang:
I added a file descriptor to the socket id data structure for demuxing from incoming I/O but it is not platform-agnostic.


anandbonde (approved):
Looks good.

PR #998, opened 2023-11-08
[inetstack] Remove spurious wake ups

iyzhang:
Because all of the yielders in a single coroutine share a waker bit, we are prone to spurious wake ups when a coroutine becomes runnable. Specifically, we cannot distinguish between timeouts and watched values changing, causing us to send acks and retransmits far too frequently.

### DIFF
--- a/src/rust/inetstack/protocols/tcp/established/background/acknowledger.rs
+++ b/src/rust/inetstack/protocols/tcp/established/background/acknowledger.rs
@@ -17,7 +17,7 @@ use ::futures::future::{
 };
 use ::std::time::Instant;
 
-pub async fn acknowledger<const N: usize>(mut cb: SharedControlBlock<N>) -> Result<!, Fail> {
+pub async fn acknowledger<const N: usize>(mut cb: SharedControlBlock<N>, yielder: Yielder) -> Result<!, Fail> {
     loop {
         // TODO: Implement TCP delayed ACKs, subject to restrictions from RFC 1122
         // - TCP should implement a delayed ACK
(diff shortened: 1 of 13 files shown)


src/rust/inetstack/protocols/tcp/established/background/acknowledger.rs
```diff
@@ -33,16 +33,20 @@ pub async fn acknowledger<const N: usize>(mut cb: SharedControlBlock<N>) -> Resu
...
         };
         futures::pin_mut!(ack_future);
 
         futures::select_biased! {
             _ = ack_deadline_changed => continue,
             _ = ack_future => {
+                match cb.get_ack_deadline().get() {
+                    Some(timeout) if timeout > cb.get_now() => continue,
+                    None => continue,
+                    _ => {},
```
anandbonde, line 48: Is there a possible `_` condition if we have already covered `Some` and `None`?

src/rust/inetstack/protocols/tcp/established/background/retransmitter.rs
```diff
@@ -52,11 +51,17 @@ pub async fn retransmitter<const N: usize>(mut cb: SharedControlBlock<N>) -> Res
...
+        // Since these futures all share a single waker bit, they are all woken whenever one of them triggers.
         futures::select_biased! {
             _ = rtx_deadline_changed => continue,
             _ = rtx_fast_retransmit_changed => continue,
             _ = rtx_future => {
-                trace!("Retransmission Timer Expired");
+                match cb.get_retransmit_deadline() {
+                    Some(timeout) if timeout > cb.get_now() => continue,
+                    None => continue,
+                    _ => {},
```
anandbonde, line 62: Is there a possible `_` condition if we have already covered `Some` and `None`?

anandbonde (approved):
Looks good.

src/rust/inetstack/protocols/tcp/established/background/acknowledger.rs
iyzhang: Yes if timeout is less than cb.get_now(), then we actually want to send the ack.

anandbonde: Oh yes, I missed the condition. This loosk good. Thanks.

PR #999, opened 2023-11-09
[runtime] Enhancement: Remove a warning

iyzhang:
I added a constant that is only used for debug builds, so we are getting an unused code warning on release builds.


anandbonde (approved):
Looks good.

Issue #224, opened 2021-11-30
[inetstack] `close()` Does not Close a Connection


ppenna:
Description
========

Currently, when we close a TCP connection using the `close()` system call, we do not `wait()` for the connection FIN packet to arrive, thereby leaving the connection in an unknown state.

Related Issues
----------------

- https://github.com/demikernel/catnip/issues/48

PR #1001, opened 2023-11-09
[inetstack] Move TCP stack to objects

iyzhang:
This PR moves the TCP portion of the inetstack to a more object oriented architecture. This move will enable us to eventually shared code between the libOSes for queues and the socket state machine.

### DIFF
--- a/src/rust/inetstack/protocols/tcp/established/mod.rs
+++ b/src/rust/inetstack/protocols/tcp/established/mod.rs
@@ -7,17 +7,24 @@ mod ctrlblk;
 mod rto;
 mod sender;
 
-pub use self::ctrlblk::{
-    SharedControlBlock,
-    State,
-};
-
 use crate::{
-    inetstack::protocols::tcp::segment::TcpHeader,
+    inetstack::{
+        protocols::tcp::{
+            congestion_control::CongestionControlConstructor,
+            established::ctrlblk::SharedControlBlock,
+            segment::TcpHeader,
+            SeqNumber,
+        },
+        MacAddress,
+        SharedArpPeer,
+        TcpConfig,
+    },
     runtime::{
         fail::Fail,
         memory::DemiBuffer,
+        network::NetworkRuntime,
         QDesc,
+        SharedBox,
         SharedDemiRuntime,
     },
     scheduler::{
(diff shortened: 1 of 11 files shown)


src/rust/inetstack/protocols/tcp/established/background/mod.rs
```diff
@@ -13,32 +13,32 @@ use self::{
...
     FutureExt,
 };
 
-pub async fn background<const N: usize>(
-    cb: SharedControlBlock<N>,
-    fd: QDesc,
-    _dead_socket_tx: mpsc::UnboundedSender<QDesc>,
-) {
-    let acknowledger = acknowledger(cb.clone()).fuse();
+pub async fn background<const N: usize>(cb: SharedControlBlock<N>, _dead_socket_tx: mpsc::UnboundedSender<QDesc>) {
```
anandbonde, line 23: `_dead_socket_tx` is not used, then why is it passed into this function?

src/rust/inetstack/protocols/tcp/established/mod.rs
```diff
@@ -45,15 +52,48 @@ pub struct EstablishedSocket<const N: usize> {
...
+        ack_delay_timeout: Duration,
+        receiver_window_size: u32,
+        receiver_window_scale: u32,
+        sender_seq_no: SeqNumber,
+        sender_window_size: u32,
+        sender_window_scale: u8,
+        sender_mss: usize,
+        cc_constructor: CongestionControlConstructor,
+        congestion_control_options: Option<congestion_control::Options>,
+        dead_socket_tx: mpsc::UnboundedSender<QDesc>,
```
anandbonde, line 72: This is a large number of arguments. We should possible give it a thought about grouping/encapsulating these in logical data structures. Note: This is not an urgent requirement, but something to think about in the long run based on how the EstablishedSocket data structure evolves/is used in future.

anandbonde (approved):
Looks good.

src/rust/inetstack/protocols/tcp/established/background/mod.rs
iyzhang: I don't know. :) I thought perhaps we would use it someday because it was included in the first place but I'm not actually sure of the purpose.

src/rust/inetstack/protocols/tcp/established/mod.rs
iyzhang: Yes, definitely. I think we could certainly do a sort of clone or something of an existing socket and then set up the socket with everything that is different but I'm not sure that is a very common Rust thing to do. I think we probably want to have a socket state data structure that keeps pointers to all of these options and things, which is separate from the EstablishedSocket object but that is also not very object-oriented.

src/rust/inetstack/protocols/tcp/peer.rs
```diff
@@ -210,351 +179,193 @@ impl<const N: usize> SharedTcpPeer<N> {
...
+                        Ok(endpoints) => endpoints,
+                        Err(e) => return (qd, OperationResult::Failed(e)),
+                    };
+                    // Wait for connect to complete.
+                    if let Some(existing_qd) =
+                        runtime.insert_socket_id_to_qd(SocketId::Active(endpoints.0, endpoints.1), new_qd)
+                    {
+                        // We should panic here because the ephemeral port allocator should not allocate the same port more than
+                        // once.
+                        panic!(
```
ppenna, line 222: Shouldn't this be `unreachable!()`?

src/rust/inetstack/protocols/tcp/peer.rs
```diff
@@ -210,351 +179,193 @@ impl<const N: usize> SharedTcpPeer<N> {
...
+        let mut peer: SharedTcpPeer<N> = self.clone();
+        Ok(Box::pin(async move {
+            // Wait for connect to complete.
+            if let Some(existing_qd) = peer
+                .runtime
+                .insert_socket_id_to_qd(SocketId::Active(local, remote.clone()), qd)
+            {
+                // We should panic here because the ephemeral port allocator should not allocate the same port more than
+                // once.
+                panic!(
```
ppenna, line 253: Shouldn't this be `unreachable!()`?

src/rust/inetstack/protocols/tcp/peer.rs
```diff
@@ -210,351 +179,193 @@ impl<const N: usize> SharedTcpPeer<N> {
...
+        // if let Some(socket_id) = queue.close()? {
+        //     match self.runtime.remove_socket_id_to_qd(&socket_id) {
+        //         Some(existing_qd) if existing_qd == qd => {},
+        //         _ => return Err(Fail::new(libc::EINVAL, "socket id did not map to this qd!")),
+        //     };
+        // }
+        // // Free the queue.
+        // self.runtime
+        //     .free_queue::<SharedTcpQueue<N>>(&qd)
+        //     .expect("queue should exist");
```
ppenna, line 317: Could we have this linked to an issue?

ppenna (approved):
Thanks for working on this. I left some minor comments here and there.

src/rust/inetstack/protocols/tcp/peer.rs
iyzhang: Yup, fixed it!

src/rust/inetstack/protocols/tcp/peer.rs
iyzhang: Yes, linked to #224

PR #1002, opened 2023-11-09
[catloop] Do not remove pending op for Close

anandbonde:
Do nothing for Close because it's not registered. This is done because we don't want to allow canceling the Close op as that may leave dangling resources.

### DIFF
--- a/src/rust/catloop/mod.rs
+++ b/src/rust/catloop/mod.rs
@@ -534,9 +534,15 @@ impl SharedCatloopLibOS {
         let task: OperationTask = self.runtime.remove_coroutine(&handle);
         let (qd, result): (QDesc, OperationResult) = task.get_result().expect("The coroutine has not finished");
 
-        match self.get_queue(&qd) {
-            Ok(mut queue) => queue.remove_pending_op(&handle),
-            Err(_) => debug!("take_result(): this queue was closed (qd={:?})", qd),
+        match result {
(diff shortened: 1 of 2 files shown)


iyzhang (approved):

src/rust/catloop/mod.rs
```diff
@@ -534,9 +534,16 @@ impl SharedCatloopLibOS {
         let task: OperationTask = self.runtime.remove_coroutine(&handle);
         let (qd, result): (QDesc, OperationResult) = task.get_result().expect("The coroutine has not finished");
 
-        match self.get_queue(&qd) {
-            Ok(mut queue) => queue.remove_pending_op(&handle),
-            Err(_) => debug!("take_result(): this queue was closed (qd={:?})", qd),
+        match result {
+            // Do nothing for Close because it's not registered. This is done because we don't want to allow canceling
+            // the Close op as that may leave dangling resources.
```
iyzhang, line 539: Actually we won't be able to do this for close because the queue will have already been freed. If it is gone for some other reason (i.e., line 544), then something else went wrong. Do you want to update the comment and the error? I think we should warn if there is a lingering op that is not a close.

anandbonde: Sure, I'll update the comment and a warn msg.

anandbonde: Done.

PR #1003, opened 2023-11-10
[inetstack] Move UDP stack to objects

iyzhang:
This PR moves the UDP stack towards our new architecture. We have moved functionality into each SharedUDPQueue, including a per-queue background coroutine that sends pending packets. Since most of the time we will be able to send synchronously, it's better to avoid the lookup on each send and just rely on the scheduler to wake the appropriate background coroutine. I also added a clean up for the background coroutine on socket close.

This PR has a few lingering issues: 
1. I moved the UDP queues to the AsyncQueue data structure, which is currently unbounded. We should move it back to being bounded and possibly using the mpsc abstractions from the futures crate. We should decide what kind of blocking behavior we want when the receive or send queues fill up.
2. I added a SharedAsyncQueue data structure but wound up not using it. I left it in case we need it somewhere else.

Other remaining tasks:
1. Add UDP support to the socket state machine and have the SharedUdpQueue check it on each operation.
2. Add pending ops to the SharedUdpQueue and cancel them when the queue is closed.

### DIFF
--- a/src/rust/inetstack/protocols/udp/peer.rs
+++ b/src/rust/inetstack/protocols/udp/peer.rs
@@ -98,11 +98,11 @@ pub struct SharedUdpPeer<const N: usize>(SharedObject<UdpPeer<N>>);
 // Associate Functions
 //======================================================================================================================
 
-/// Associate functions for [UdpPeer].
-impl<const N: usize> UdpPeer<N> {
-    /// Creates a Udp peer.
+/// Associate functions for [SharedUdpPeer].
+
+impl<const N: usize> SharedUdpPeer<N> {
     pub fn new(
-        mut runtime: SharedDemiRuntime,
+        runtime: SharedDemiRuntime,
         transport: SharedBox<dyn NetworkRuntime<N>>,
         local_link_addr: MacAddress,
         local_ipv4_addr: Ipv4Addr,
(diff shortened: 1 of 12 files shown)


ppenna (approved):

src/rust/inetstack/protocols/udp/peer.rs
```diff
@@ -241,55 +155,46 @@ impl<const N: usize> SharedUdpPeer<N> {
...
-    pub fn pushto(&mut self, qd: QDesc, data: DemiBuffer, remote: SocketAddrV4) -> Result<(), Fail> {
+    pub fn pushto(&mut self, qd: QDesc, buf: DemiBuffer, remote: SocketAddrV4) -> Result<Pin<Box<Operation>>, Fail> {
         #[cfg(feature = "profiler")]
         timer!("udp::pushto");
-        // Lookup associated endpoint.
+        trace!("pushto(): qd={:?} remote={:?} bytes={:?}", qd, remote, buf.len());
         let mut queue: SharedUdpQueue<N> = self.get_shared_queue(&qd)?;
-        // What happens if not bound?
-        // TODO: Move to a UDP socket state machine.
+        // TODO: Allocate ephemeral port if not bound.
```
ppenna, line 170: Could you file an issue for this?

ppenna (approved):

src/rust/inetstack/protocols/udp/peer.rs
iyzhang: Whoops, apparently I already filed one! https://github.com/microsoft/demikernel/issues/973
I will add it to a FIXME

PR #1004, opened 2023-11-11
[demikernel] Switch to `runtime::pack_result()` for `catcollar` and `catpowder`

anandbonde:
[demikernel] Switch to `runtime::pack_result()` for `catcollar` and `catpowder`


ppenna (approved):
LGTM

PR #1006, opened 2023-11-14
[ci] Reenable Catnap CI on Windows

ppenna:
## Description

This PR re-enables Catnap CI on Windows.

### DIFF
--- a/.github/workflows/catnapw.yml
+++ b/.github/workflows/catnapw.yml
@@ -6,10 +6,6 @@ concurrency:
   group: catnapw
   cancel-in-progress: false
 
-concurrency:
-  group: azure-catnapw
-  cancel-in-progress: true
-
 on:
   push:
     branches:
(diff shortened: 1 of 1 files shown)


.github/workflows/catnapw.yml
```diff
@@ -6,10 +6,6 @@ concurrency:
   group: catnapw
```
anandbonde, line 6: Just a thought, should we create separate directories per OS instead of naming the files with a `w` prefix? What do you think?

One option is to defer this to a later point when we need more than 1 workflow files for Windows.

anandbonde (approved):
Looks good.

.github/workflows/catnapw.yml
ppenna: The concurrent group is should be an unique identifier (it distinguishes which actions conflict), and we cannot use a directory based structure to achieve what you want.

Instead we can have suffixes like "catnap-linux" and "catnap-windows".

What do you think?

anandbonde: Oh, I was referring to the filenames. e g., `.github/workflows/windows/catnap.yml`. 

Also, your suggestion about group names sounds good as well.

ppenna: Sure, I can rename them, but not sure that we can separate them into directories. See: https://stackoverflow.com/questions/64009546/how-to-run-multiple-github-actions-workflows-from-sub-directories

Issue #1007, opened 2023-11-16
[inetstack] Deduplicate state machine


iyzhang:
## Context

Currently we have one state machine that is used by Catnap and Catloop and then an additional one in the inetstack. This should be deduplicated into one state machine.

## Proposed Solution

Add the ability to store state/sockets in the state machine? I think we should be able to store sockets in the state machine and/or addresses depending on the state that the queue/socket is in. This will eliminate the need for the existing state machine in the TCP stack.

PR #1008, opened 2023-11-16
[inetstack] Move TCP stack to use shared state machine

iyzhang:
This PR moves the TCP portion of the inetstack to using the shared state machine that is in the runtime. A few important things to note:
1. The TCP queue is now maintaining 2 state machines. We should deduplicate this and I have filed an issue to track it #1007.
2. This PR moves the inetstack closer to the same architecture as Catnap with functions passed in that move the state machine forward/rollback when there are errors (e.g., we were not able to insert the coroutine into the scheduler).  
3. This PR has added a list of pending ops to the TCP queue. However, because the packing for each returned result is in the libOS, I cannot remove the operation from the pending list. I think this will be resolved with #888 however.
4.  To align with Catnap, we now return queue tokens on synchronous paths for the task that was inserted to run the coroutine. This required adding another call to the runtime so we can remove coroutines using queue tokens because it is needed for the test engine. This should all be deduplicated once we merge the return path and the testing infrastructure.
5. There was one place where the existing inetstack state machine and the shared runtime state machine diverged on the return value when listening on an already listening socket. The former returns EINVAL and the latter returns EADDRINUSE. I chose to go with the runtime but we should make sure that is the expected behavior because I had to change a test case.
6. I've placed profiler and trace messages in the TCP peer because I suspect the other file will go away. I had started adding some of them anyway when debugging the other test cases and it seemed like we should just have them at a finer granularity.

### DIFF
--- a/src/rust/inetstack/protocols/tcp/peer.rs
+++ b/src/rust/inetstack/protocols/tcp/peer.rs
@@ -26,10 +26,14 @@ use crate::{
             NetworkRuntime,
         },
         queue::NetworkQueue,
-        scheduler::Yielder,
+        scheduler::{
+            TaskHandle,
+            Yielder,
+        },
         Operation,
         OperationResult,
         QDesc,
+        QToken,
         SharedBox,
         SharedDemiRuntime,
         SharedObject,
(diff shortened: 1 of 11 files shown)


src/rust/inetstack/protocols/tcp/peer.rs
```diff
@@ -142,7 +146,6 @@ impl<const N: usize> SharedTcpPeer<N> {
         }
 
         // TODO: Check if we are binding to a non-local address.
-
```
ppenna, line 145: Could you rollback this change? The TODO comment does not concern the code block that follows.

iyzhang: Done

PR #1012, opened 2023-11-16
[inetstack] Bug Fix: Ignore random resets

iyzhang:
We seem to get random packets with the RST bit set, so this PR just ignores them. Otherwise, we would treat it as a hard shutdown of the connection, which is keeping the tests from continuing. We should look into where these packets are coming from because we are definitely not sending them from the client to the server.


iyzhang:
Realized that this is only an issue if switching between Catpowder and Catnip because Linux is sending resets. We should be sure to clearly indicate whether Linux should be receiving packets from the IP address that we are using for the tests.

PR #1013, opened 2023-11-17
[testing] Turn on TCP echo tests for inetstack

iyzhang:
This PR turns on TCP echo tests with a single client for Catnip and Catpowder.


ppenna (approved):
LGTM

PR #1015, opened 2023-11-17
[inetstack] Finish implementing async tcp close

### LINKED ISSUE #224: [inetstack] `close()` Does not Close a Connection

iyzhang:
This PR closes #224

### DIFF
--- a/src/rust/inetstack/protocols/tcp/established/ctrlblk.rs
+++ b/src/rust/inetstack/protocols/tcp/established/ctrlblk.rs
@@ -13,7 +13,10 @@ use super::{
     },
 };
 use crate::{
-    collections::async_queue::AsyncQueue,
+    collections::{
+        async_queue::AsyncQueue,
+        async_value::AsyncValue,
+    },
     inetstack::protocols::{
         arp::SharedArpPeer,
         ethernet2::{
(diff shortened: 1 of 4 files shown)


src/rust/inetstack/protocols/tcp/established/ctrlblk.rs
```diff
@@ -774,20 +784,30 @@ impl<const N: usize> SharedControlBlock<N> {
...
-        debug_assert!((self.state == State::Established) || (self.state == State::CloseWait));
-
         // Send a FIN.
+        // TODO: Set state to FIN-WAIT1 if currently established or set to LASTACK if CloseWait.
+        let new_state: State = match self.state {
+            State::Established => State::FinWait1,
+            State::CloseWait => State::LastAck,
+            state => {
+                let cause: String = format!("Cannot close a socket from any other state (current state={:?},", state);
+                warn!("{}", cause);
```
ppenna, line 794: Shouldn't this be an error instead?

ppenna (approved):
Thanks for working on this. I cant wait to deprecate sync close :D

src/rust/inetstack/protocols/tcp/established/ctrlblk.rs
iyzhang: Sure, since it was replacing a debug assert, I thought it should be a warn, but I will change it.

PR #1016, opened 2023-11-17
[ci] Gather and Upload Performance Statistics

ppenna:
## Description

In this PR I introduce required changes to gather performance statistics from CI runs on release mode, and upload results to Azure Storage.

**This should be considered an initial implementation**. Next, I intend to integrate performance regression on our pipeline.


iyzhang (approved):
Thanks for doing this!

PR #1017, opened 2023-11-18
[runtime] Add extra states to indicate which end is closing

### LINKED ISSUE #369: [inetstack] TCP Should Indicate to PDPIX when a Connection is Closing

iyzhang:
This PR closes #369 and #367. We currently signal close to application by returning a zero-length buffer, per the POSIX specification and issue a cancellation of all other tokens. This PR also lets us distinguish whether we are closing from the remote end or the local end or if both have issued close. Once the application receives a zero-length buffer indicating that the remote has closed, the application should also close its end. If the application closes locally first, then the async close will return when the remote has also indicated close and all resources can be cleaned up.  If we would like another way to indicate close, then we should file another issue (and find an error code because POSIX does not offer one in libc).

### DIFF
--- a/src/rust/inetstack/protocols/tcp/queue.rs
+++ b/src/rust/inetstack/protocols/tcp/queue.rs
@@ -291,7 +291,19 @@ impl<const N: usize> SharedTcpQueue<N> {
     pub async fn pop_coroutine(&mut self, size: Option<usize>, yielder: Yielder) -> Result<DemiBuffer, Fail> {
         self.state_machine.may_pop()?;
         match self.socket {
-            Socket::Established(ref mut socket) => socket.pop(size, yielder).await,
+            Socket::Established(ref mut socket) => {
+                let buf: DemiBuffer = socket.pop(size, yielder).await?;
+                // If this is an EOF on the connection, then close the socket.
+                // TODO: Get rid of this when we deduplicate the state machine.
+                // FIXME:
+                if buf.is_empty() {
+                    self.state_machine.prepare(SocketOp::RemoteClose)?;
+                    self.state_machine.commit();
+                }
+                // We are assuming that the application will call close when it receives a zero-length pop. Any other
+                // pops will be cancelled with an EBADF.
+                Ok(buf)
+            },
             _ => unreachable!("State machine check should ensure that this socket is connected"),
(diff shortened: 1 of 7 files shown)


src/rust/inetstack/protocols/tcp/queue.rs
```diff
@@ -291,7 +291,19 @@ impl<const N: usize> SharedTcpQueue<N> {
     pub async fn pop_coroutine(&mut self, size: Option<usize>, yielder: Yielder) -> Result<DemiBuffer, Fail> {
         self.state_machine.may_pop()?;
         match self.socket {
-            Socket::Established(ref mut socket) => socket.pop(size, yielder).await,
+            Socket::Established(ref mut socket) => {
+                let buf: DemiBuffer = socket.pop(size, yielder).await?;
+                // If this is an EOF on the connection, then close the socket.
+                // TODO: Get rid of this when we deduplicate the state machine.
+                // FIXME:
```
ppenna, line 298: Should we remove the 'FIXME' statement?

src/rust/inetstack/protocols/tcp/queue.rs
```diff
@@ -291,7 +291,19 @@ impl<const N: usize> SharedTcpQueue<N> {
...
+                if buf.is_empty() {
+                    self.state_machine.prepare(SocketOp::RemoteClose)?;
+                    self.state_machine.commit();
+                }
+                // We are assuming that the application will call close when it receives a zero-length pop. Any other
+                // pops will be cancelled with an EBADF.
+                Ok(buf)
+            },
```
ppenna, line 306: This will work for for the happy "remote close" case. For others, we still need to modify `SharedControlBlock::receive()`

iyzhang:
We will need to re-evaluate this with the new transport trait

PR #1019, opened 2023-11-18
[catnap] Enhancement: Move to socket2

iyzhang:
This code is a merge of the Windows and Linux code paths for Catnap into a single socket2 API. It is a first step to supporting epoll and completion ports. There are 3 commits: 
1. Just commenting out some overwhelming debug messages.
2. The bulk of the changes adding a CatnapTransport
3. An extra bug fix. It seems like if we start the next test too soon after the previous one exits, the old server address is still bound for a little while. We should probably set to this to some reasonable number but not really sure what.
4. Remove level of unnecessary abstraction for sockets. Now that we are using Socket2, we can keep all of the per-queue code in either the CatnapQueue or the transport.


src/rust/demikernel/libos/network.rs
```diff
@@ -66,7 +66,7 @@ impl NetworkLibOS {
             #[cfg(feature = "catpowder-libos")]
             NetworkLibOS::Catpowder(libos) => libos.socket(domain, socket_type, protocol),
             #[cfg(all(feature = "catnap-libos"))]
-            NetworkLibOS::Catnap(libos) => libos.socket(domain, socket_type, protocol),
+            NetworkLibOS::Catnap(libos) => libos.socket(domain.into(), socket_type.into(), protocol.into()),
```
ppenna, line 69: Once you move all LibOSes to this, I would sugges us to move this conversion logic up to the PDPIX layer. What do you think?

ppenna:
> An extra bug fix. It seems like if we start the next test too soon after the previous one exits, the old server address is still bound for a little while. We should probably set to this to some reasonable number but not really sure what.

@iyzhang this is due to the "TIME_WAIT" state. There are some per-system solutions that might be explored, such as using the `SO_REUSEADDR`.

tests/rust/tcp-test/bind/mod.rs
```diff
@@ -245,7 +245,7 @@ fn bind_to_non_local_address(libos: &mut LibOS) -> Result<()> {
 
     // Fail to bind socket.
     match libos.bind(sockqd, addr) {
-        Err(e) if e.errno == libc::EADDRNOTAVAIL => (),
+        Err(e) if e.errno == libc::EADDRNOTAVAIL || e.errno == libc::EBADF => (),
```
ppenna, line 248: Where is this bug comming from?

src/rust/catnap/queue.rs
```diff
@@ -293,6 +369,32 @@ impl SharedCatnapQueue {
...
+    /// Generic function for spawning a control-path coroutine on [self].
+    fn do_generic_sync_control_path_call<F>(
+        &mut self,
+        coroutine_constructor: F,
+        yielder: Yielder,
+    ) -> Result<TaskHandle, Fail>
+    where
+        F: FnOnce(Yielder) -> Result<TaskHandle, Fail>,
+    {
+        // Spawn coroutine.
```
anandbonde, line 381: With the new name `coroutine_constructor`, this comment probably needs to be revisited.

anandbonde (approved):
Looks good.

tests/rust/tcp-test/bind/mod.rs
iyzhang: socket2 returns EBADF instead from this call.

iyzhang:
> > An extra bug fix. It seems like if we start the next test too soon after the previous one exits, the old server address is still bound for a little while. We should probably set to this to some reasonable number but not really sure what.
> 
> @iyzhang this is due to the "TIME_WAIT" state. There are some per-system solutions that might be explored, such as using the `SO_REUSEADDR`.

Ah great! Let me know what the right solution is. I can certainly set SO_REUSEADDR on all listening sockets.

src/rust/demikernel/libos/network.rs
iyzhang: Yes, definitely! I would love to get rid of some of this code at the libos/network.rs level

tests/rust/tcp-test/bind/mod.rs
ppenna: Sure, but then this does not comply with expected behavior for sockets. Shouldn't we thus interpose the returned error code?

ppenna:
> > > An extra bug fix. It seems like if we start the next test too soon after the previous one exits, the old server address is still bound for a little while. We should probably set to this to some reasonable number but not really sure what.
> > 
> > 
> > @iyzhang this is due to the "TIME_WAIT" state. There are some per-system solutions that might be explored, such as using the `SO_REUSEADDR`.
> 
> Ah great! Let me know what the right solution is. I can certainly set SO_REUSEADDR on all listening sockets.

Setting `SO_REUSEADDR` in all listening sockets works for Linux, not sure how that behaves in Windows. @kyleholohan any ideas on this?

tests/rust/tcp-test/bind/mod.rs
iyzhang: Oh nooo

src/rust/catnap/queue.rs
iyzhang: Done

tests/rust/tcp-test/bind/mod.rs
iyzhang: I'm not sure how to distinguish it from the socket being closed if I just translate all EBADF to EADDRNOTAVAIL

iyzhang: Let me revisit this if/when we move to TcpStream

PR #1020, opened 2023-11-20
[catnap] Enhancement: preliminary epoll support

iyzhang:
This PR is a proposal for basic epoll support. I'm not sure if we will want to use a single yielder as I did in this PR but it seemed like the easiest to have one per CatnapQueue/socket. I just created this so we would have something to discuss.

### DIFF
--- a/src/rust/catnap/queue.rs
+++ b/src/rust/catnap/queue.rs
@@ -34,7 +34,6 @@ use crate::{
 };
 use ::socket2::{
     Domain,
-    Socket,
     Type,
 };
 use ::std::{
--- a/src/rust/catnap/queue.rs
+++ b/src/rust/catnap/queue.rs
@@ -28,7 +28,6 @@ use crate::{
             Yielder,
             YielderHandle,
         },
-        DemiRuntime,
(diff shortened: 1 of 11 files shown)


src/rust/catnap/queue.rs
```diff
@@ -59,142 +77,190 @@ pub struct SharedCatnapQueue(SharedObject<CatnapQueue>);
...
+                self.state_machine.abort();
+                Err(e)
+            },
+        }
     }
 
     /// Starts a coroutine to begin accepting on this queue. This function contains all of the single-queue,
     /// synchronous functionality necessary to start an accept.
-    pub fn accept<F>(&mut self, insert_coroutine: F) -> Result<QToken, Fail>
+    pub fn accept<F>(&mut self, coroutine_constructor: F) -> Result<QToken, Fail>
```
anandbonde, line 145: It would be good to change the variable names at the call sites too.

anandbonde (approved):
Looks good!

iyzhang:
This PR is superseded by #1029

PR #1026, opened 2023-11-21
[runtime] Move Performance Monitoring to PDPIX Layer

ppenna:
## Description

This PR closes https://github.com/microsoft/demikernel/issues/1025

## Summary of Changes

- Moved performance monitoring of Demikernel system calls to PDPIX layer
- Cleaned up deprecated files from the source tree


anandbonde (approved):
Looks good.

PR #1027, opened 2023-11-22
[test] TCP Simulation Testing

### LINKED ISSUE #453: [test] Support packet drill
## Context

We should support testing our inet stack using packet drill, which seems like a standard UDP/TCP testing tool. 
[Packet drill GitHub repo](https://github.com/google/packetdrill)

## Proposed Solution

Implement an API to the inet stack or catnip to interop with packet drill.

ppenna:
## Description

This PR closes #453


linux.mk
```diff
@@ -31,6 +31,7 @@ export BUILD_DIR := $(CURDIR)/target/release
 ifeq ($(BUILD),dev)
 export BUILD_DIR := $(CURDIR)/target/debug
 endif
+export INPUT_DIR ?= $(CURDIR)/input
```
anandbonde, line 34: Can this directory be moved inside the nettest directory? Currently it's in the Demikernel root dir so it's hard to associate it with where it belongs.

nettest/README.md
```diff
@@ -0,0 +1,26 @@
...
+## Usage Statement
+
+This project is a prototype. As such, we provide no guarantees that it will work
+and you are assuming any risks with using the code. We welcome comments and
+feedback. Please send any questions or comments to one of the following
+maintainers of the project:
+
+- [Pedro Henrique Penna](https://github.com/ppenna) - [ppenna@microsoft.com](mailto:ppenna@microsoft.com)
+
+> By sending feedback, you are consenting that it may be used  in the further
```
anandbonde, line 16: Extra space

nettest/README.md
```diff
@@ -0,0 +1,26 @@
+## Network Test Suite for Demikernel
+
+This is repository contains hosts the network test suite for Demikernel.
```
anandbonde, line 3: Could you elaborate on what you mean by `hosts` here?

nettest/src/glue.rs
```diff
@@ -0,0 +1,206 @@
...
+        }
+        write!(f, "{}", flags)
+    }
+}
+
+pub fn parse_int(s: &str) -> Result<u32, ()> {
+    match s.parse::<u32>() {
+        Ok(val) => Ok(val),
+        Err(_) => {
+            eprintln!("{} cannot be represented as a u64", s);
```
anandbonde, line 202: Should the err msg refer to u32? Maybe I'm not reading the intent correctly here.

nettest/src/main.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    let lines: Vec<String> = read_input_file(filename)?;
+
+    // Process all lines of the source file.
+    for line in lines {
+        if verbose {
+            println!("Line: {:?}", line);
+        }
+        // Process in multiple passes.
+        for pass in ["Lexical Analysis", "Syntax Analysis"] {
```
anandbonde, line 53: Do you anticipate having more passes apart from the Lexical and Syntax analysis?

src/rust/inetstack/protocols/tcp/segment.rs
```diff
@@ -109,6 +110,7 @@ impl TcpOptions2 {
     fn compute_size(&self) -> usize {
         use TcpOptions2::*;
         match self {
+            EndOfOptionsList => 0,
```
anandbonde, line 113: Could you explain why we need changes to the production code for supporting this test?

src/rust/inetstack/protocols/tcp/tests/simulator.rs
```diff
@@ -0,0 +1,827 @@
...
+    remote_addr: SocketAddrV4,
+    local_qd: Option<(u32, QDesc)>,
+    remote_qd: Option<(u32, Option<QDesc>)>,
+    server: SharedEngine<RECEIVE_BATCH_SIZE>,
+    now: Instant,
+    inflight: Option<QToken>,
+    steps: Vec<String>,
+}
+
+impl Simulation {
```
anandbonde, line 173: Suggestion: This file/class and some of the functions are long. Is it possible to split it into smaller groups of functionalities? If we don't expect much change in this class then it's not a big concern.

anandbonde (changes requested):
Overall this looks great. Thanks! Left a few comments.

linux.mk
ppenna: Sure.

nettest/README.md
ppenna: Fixed.

nettest/README.md
ppenna: Fixed.

nettest/src/main.rs
ppenna: It may be possible. For instance it would be nice to add a pre-processing step to substitute compile-time constants. Why the question?

src/rust/inetstack/protocols/tcp/segment.rs
ppenna: It fixes a bug for supporting the initial test case.

src/rust/inetstack/protocols/tcp/tests/simulator.rs
ppenna: I intend to be working on this whole module for the next months. As we agreed, I just pushed an initial version so that it does not break recurrently with all the changes that are being introduced.

nettest/src/main.rs
anandbonde: `Preprocessing` layer sounds good. I was curious about why we need the loop on Ln. 53. But this answers my question.

src/rust/inetstack/protocols/tcp/segment.rs
anandbonde: Awesome!

src/rust/inetstack/protocols/tcp/tests/simulator.rs
anandbonde: Yes, that sounds good.

anandbonde (approved):
Looks good! Thanks for doing this!

PR #1040, opened 2023-11-30
[runtime] Ephemeral Allocator Misses to Allocate Last Port on Private Range

ppenna:
## Description

This PR closes https://github.com/microsoft/demikernel/issues/1039

## Summary of Changes

- Included last port in the allocator
- Changed shuffling to occur only on release mode
- Cleaned up dead code

### DIFF
--- a/src/rust/runtime/network/ephemeral.rs
+++ b/src/rust/runtime/network/ephemeral.rs
@@ -31,16 +31,6 @@ pub struct EphemeralPorts {
 //======================================================================================================================
 
 impl EphemeralPorts {
-    /// Creates a new ephemeral port allocator.
-    pub fn new(rng: &mut SmallRng) -> Self {
-        let mut ports: Vec<u16> = Vec::<u16>::new();
-        for port in FIRST_PRIVATE_PORT..LAST_PRIVATE_PORT {
-            ports.push(port);
-        }
-        ports.shuffle(rng);
-        Self { ports }
-    }
-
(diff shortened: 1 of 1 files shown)


anandbonde (approved):
Looks good.

src/rust/runtime/network/ephemeral.rs
```diff
@@ -12,11 +17,10 @@ use ::rand::prelude::{
 // Constants
 //======================================================================================================================
 
+/// First private port.
```
anandbonde, line 20: It would be good to comment why these values are chosen for these consts.

ppenna: These come from the RFC itself, sure I can link the RFC here.

src/rust/runtime/network/ephemeral.rs
```diff
@@ -96,21 +90,15 @@ impl EphemeralPorts {
...
-                SmallRng::seed_from_u64(EPHEMERAL_PORT_SEED)
-            }
-            #[cfg(not(debug_assertions))]
-            {
-                SmallRng::from_entropy()
-            }
-        };
         let mut ports: Vec<u16> = Vec::<u16>::new();
-        for port in FIRST_PRIVATE_PORT..LAST_PRIVATE_PORT {
+        for port in (FIRST_PRIVATE_PORT..=LAST_PRIVATE_PORT).rev() {
```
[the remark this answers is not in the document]
anandbonde: Sounds good.

PR #1041, opened 2023-11-30
[catnap] Track pending_op for async_close

anandbonde:
Currently, the `async_close()` does not track pending_op, and this change is to enable that tracking.


ppenna (approved):
LGTM

PR #1042, opened 2023-12-01
[test] Adding Blocking Connect Test to Simulator

ppenna:
## Description

This PR adds a blocking connect test to the TCP simulator.


src/rust/runtime/network/ephemeral.rs
```diff
@@ -12,10 +17,12 @@ use ::rand::prelude::{
 // Constants
 //======================================================================================================================
 
+/// First private port.
```
anandbonde, line 20: It would be good to comment why these values were chosen for the consts.

src/rust/runtime/network/ephemeral.rs
```diff
@@ -96,21 +93,15 @@ impl EphemeralPorts {
...
-                SmallRng::seed_from_u64(EPHEMERAL_PORT_SEED)
-            }
-            #[cfg(not(debug_assertions))]
-            {
-                SmallRng::from_entropy()
-            }
-        };
         let mut ports: Vec<u16> = Vec::<u16>::new();
-        for port in FIRST_PRIVATE_PORT..LAST_PRIVATE_PORT {
+        for port in (FIRST_PRIVATE_PORT..=LAST_PRIVATE_PORT).rev() {
```
anandbonde, line 97: Why is the loop `rev`?

anandbonde (approved):
Looks good.

PR #1047, opened 2023-12-05
[inetstack] Remove background coroutine from active connect

iyzhang:
This PR is a step towards having a linear flow for all of our TCP state machines by doing the following:
1. Using the connect coroutine to directly drive retries of SYN send instead of another background coroutine.
2. Directly moving the TcpHeader/packet to the module where it will be processed rather than borrowing it in the receive path. The receive path simply pushes the packet into an AsyncQueue, while the connect coroutine waits on arriving packets. The connect coroutine will then wake up and process the incoming packet.
3. The flow of connect is not changed too much because it only requires waiting on one packet. The utility of this architecture will be clear once we are performing part of the architecture with more steps. Here we are mostly getting rid of the extra coroutine.

### DIFF
--- a/src/rust/inetstack/protocols/tcp/active_open.rs
+++ b/src/rust/inetstack/protocols/tcp/active_open.rs
@@ -6,7 +6,7 @@
 //======================================================================================================================
 
 use crate::{
-    collections::async_value::AsyncValue,
+    collections::async_queue::AsyncQueue,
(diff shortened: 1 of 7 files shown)


ppenna (approved):
Looks good to me.

At first, I was wondering if move semantics would cause any side effects, but borrowing rules just don't let them happen 😄

PR #1049, opened 2023-12-05
[inetstack] Move poll_bg_work to coroutine

iyzhang:
This PR moves the poll_bg_work function which polls the DPDK interface to its own background cororoutine. This moves us toward being able to align the network stack with the rest of the other libOSes.


src/rust/runtime/mod.rs
```diff
@@ -99,6 +99,12 @@ use self::{
     },
 };
 
+//======================================================================================================================
+// Constants
+//======================================================================================================================
+
+const TIMER_RESOLUTION: usize = 64;
```
ppenna, line 106: Could you add a comment here that explains the rationale for setting `TIMER_RESOLUTION` to 64?

src/rust/collections/async_queue.rs
```diff
@@ -93,14 +99,27 @@ impl<T> AsyncQueue<T> {
...
         self.queue.is_empty()
     }
+
+    /// Get an iterator over values
+    #[allow(unused)]
+    pub fn get_values(&self) -> Iter<T> {
+        self.queue.iter()
+    }
+
+    /// Get an interator over mutable values
```
ppenna, line 119: Typo.

ppenna (approved):
Looks good to me.

PR #1050, opened 2023-12-05
[test] Use TCP Async Close in Test Engine

ppenna:
## Description

This PR changes the tcp test engine to use `async_close()` rather than `close()`.


anandbonde (approved):
Looks good.

PR #1055, opened 2023-12-06
[examples] Use `async_close()` in tcp-echo

anandbonde:
This PR:
- Removes synchronous close from the tcp-echo example.
- Moves common code to a helper functions module.

This is a part of the series of PRs aimed towards removing synchronous close() from Demikernel.


anandbonde:
This PR is no longer needed. Another PR https://github.com/microsoft/demikernel/pull/1064 supersedes these change.

PR #1058, opened 2023-12-08
[runtime] Enhancement: Use runtime instead of libos

iyzhang:
This PR removes the following functions from the libOSes and directly calls the runtime:
* poll 
* schedule/from_task_id
* pack_result
* sgaalloc
* sgafree

Note that we have not moved the sgaalloc and free functions for the libOSes that need special memory. I have filed an issue for that: https://github.com/microsoft/demikernel/issues/1057

### DIFF
--- a/src/rust/catmem/mod.rs
+++ b/src/rust/catmem/mod.rs
@@ -237,20 +237,6 @@ impl SharedCatmemLibOS {
         (qd, result)
     }
 
-    pub fn from_task_id(&self, qt: QToken) -> Result<TaskHandle, Fail> {
-        self.runtime.from_task_id(qt.into())
-    }
-
-    /// Allocates a scatter-gather array.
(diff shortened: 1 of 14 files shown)


src/rust/catmem/mod.rs
```diff
@@ -225,7 +225,7 @@ impl SharedCatmemLibOS {
     }
 
     /// Takes out the [OperationResult] associated with the target [TaskHandle].
-    fn take_result(&mut self, handle: TaskHandle) -> (QDesc, OperationResult) {
+    pub fn take_result(&mut self, handle: TaskHandle) -> (QDesc, OperationResult) {
         let task: OperationTask = self.runtime.remove_coroutine(&handle);
```
ppenna, line 229: No need to be `pub`.

ppenna (approved):
LGTM

PR #1059, opened 2023-12-10
[inetstack] Move ARP cache to background polling model

iyzhang:
For the inetstack, we want to unwrap and turn the state machine operations into a linear flow. In order to accommodate this, we need to move away from a polling receive model where we always call the receive function from the top every time. Instead, we will simply insert the new packet into an AsyncQueue and the inetstack will pick up that packet from wherever in the linear flow it is blocked. This also means that the receive function will not return a result because it does not know if the inetstack will drop the packet. 

This change impacts the ARP cache very little since it does not have a state machine and does not really flow through a protocol. It will be more important for TCP, especially during connection set up and close, when we are waiting for very specific next packets.

### DIFF
--- a/src/rust/inetstack/protocols/arp/peer.rs
+++ b/src/rust/inetstack/protocols/arp/peer.rs
@@ -10,6 +10,7 @@ use super::{
     },
 };
 use crate::{
+    collections::async_queue::AsyncQueue,
     inetstack::protocols::ethernet2::{
         EtherType2,
         Ethernet2Header,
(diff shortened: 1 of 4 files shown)


src/rust/inetstack/protocols/arp/peer.rs
```diff
@@ -167,87 +151,105 @@ impl<const N: usize> SharedArpPeer<N> {
         rx.await.expect("Dropped waiter?")
     }
 
-    /// Background task that cleans up the ARP cache from time to time.
-    async fn background(&mut self) {
-        let yielder: Yielder = Yielder::new();
+    pub async fn poll(mut self) {
         loop {
```
ppenna, line 155: No need to be `pub`.

iyzhang: good catch

ppenna (approved):
LGTM

PR #1063, opened 2023-12-11
[inetstack] Enhancement: Move ICMP to polling

iyzhang:
This PR moves our ICMP ping infrastructure to using a background polling request, rather than doing all of the processing in the receive function. It does not change the number of coroutines since we already had a background coroutine.


ppenna (approved):
LGTM

PR #1064, opened 2023-12-11
[demikernel] Remove synchronous close() from catnap,catcollar,catloop liboses

anandbonde:
This PR removes the synchronous close path from catnap,catcollar,catloop liboses.


ppenna (approved):
LGTM

PR #1066, opened 2023-12-12
[catmem] Remove external support for synchronous close()

anandbonde:
This PR removes the ability to call synchronous close() on catmem via the Demikernel API.

Incremental PR based on this PR https://github.com/microsoft/demikernel/pull/1064.


ppenna (approved):
LGTM

PR #1068, opened 2023-12-12
[inetstack] Move TCP to receive polling coroutine

iyzhang:
This PR moves the remainder of the inetstack to AsyncQueues and doing network processing in a background coroutine instead of directly in the receive function. A few things to note:
1. We can no longer tell if the inetstack will drop the packet in the receive path. We need to wait for the stack to pick it up from the AsyncQueue and process it. This makes sense because whether the packet is valid depends on what state the inetstack is in. 
2. We use to check the state of the inetstack (or not check it if we decided it would be too expensive). Now the state of the inetstack will be encoded in the async coroutine and where it is blocked.
3. Because the receive path can no longer tell you if the packet will be dropped, some of our tests no longer work. I have filed an issue for this, but I wanted to wait and see what we want to do with these tests with our new nettest frame work.
4. This PR is just the very first step in removing the states from the inetstack and using the async/await framework to encode them instead. This will allow us to unroll the entire lifetime of a TCP socket into a single linear flow.

### DIFF
--- a/src/rust/collections/async_queue.rs
+++ b/src/rust/collections/async_queue.rs
@@ -39,7 +39,6 @@ pub struct AsyncQueue<T> {
     waiters: Vec<YielderHandle>,
 }
 
-#[derive(Clone)]
 pub struct SharedAsyncQueue<T>(SharedObject<AsyncQueue<T>>);
 
 //======================================================================================================================
@@ -160,5 +159,11 @@ impl<T> Deref for SharedAsyncQueue<T> {
 impl<T> DerefMut for SharedAsyncQueue<T> {
     fn deref_mut(&mut self) -> &mut Self::Target {
         self.0.deref_mut()
+    }
+}
+
+impl<T> Clone for SharedAsyncQueue<T> {
+    fn clone(&self) -> Self {
+        Self(self.0.clone())
     }
 }
--- a/src/rust/inetstack/protocols/tcp/passive_open.rs
+++ b/src/rust/inetstack/protocols/tcp/passive_open.rs
@@ -6,7 +6,10 @@
 //======================================================================================================================
 
 use crate::{
-    collections::async_queue::AsyncQueue,
+    collections::async_queue::{
+        AsyncQueue,
+        SharedAsyncQueue,
+    },
     inetstack::protocols::{
         arp::SharedArpPeer,
         ethernet2::{
@@ -44,15 +47,18 @@ use crate::{
             Yielder,
             YielderHandle,
         },
-        timer::SharedTimer,
         QDesc,
         SharedBox,
         SharedDemiRuntime,
         SharedObject,
     },
 };
-use ::core::panic;
-use ::futures::channel::mpsc;
+use ::futures::{
+    channel::mpsc,
+    pin_mut,
+    select_biased,
+    FutureExt,
+};
 use ::libc::{
     EBADMSG,
     ETIMEDOUT,
@@ -72,18 +78,9 @@ use ::std::{
 // Structures
 //======================================================================================================================
 
-struct InflightAccept {
-    local_isn: SeqNumber,
-    remote_isn: SeqNumber,
-    header_window_size: u16,
-    remote_window_scale: Option<u8>,
-    mss: usize,
-    handle: TaskHandle,
-    yielder_handle: YielderHandle,
-}
-
 pub struct PassiveSocket<const N: usize> {
-    inflight: HashMap<SocketAddrV4, InflightAccept>,
+    connections: HashMap<SocketAddrV4, SharedAsyncQueue<(Ipv4Header, TcpHeader, DemiBuffer)>>,
+    recv_queue: SharedAsyncQueue<(Ipv4Header, TcpHeader, DemiBuffer)>,
     ready: AsyncQueue<Result<EstablishedSocket<N>, Fail>>,
     max_backlog: usize,
     isn_generator: IsnGenerator,
@@ -94,6 +91,8 @@ pub struct PassiveSocket<const N: usize> {
     local_link_addr: MacAddress,
     arp: SharedArpPeer<N>,
     dead_socket_tx: mpsc::UnboundedSender<QDesc>,
+    yielder_handle: YielderHandle,
+    task_handle: Option<TaskHandle>,
 }
 
 #[derive(Clone)]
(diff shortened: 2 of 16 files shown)


src/rust/collections/async_queue.rs
```diff
@@ -39,7 +39,6 @@ pub struct AsyncQueue<T> {
     waiters: Vec<YielderHandle>,
 }
 
-#[derive(Clone)]
```
ppenna, line 42: Why can't we just `derive`?

src/rust/inetstack/protocols/tcp/tests/setup.rs
```diff
@@ -163,10 +162,10 @@ fn test_refuse_connection_early_rst() -> Result<()> {
...
 
     // Server: SYN_RCVD state at T(2).
-    match server.receive(buf) {
-        Err(error) if error.errno == EBADMSG => Ok(()),
-        _ => anyhow::bail!("server receive should have returned an error"),
-    }
+    server.receive(buf)?;
+    // TODO: Assert that we dropped the packet.
+    // FIXME: https://github.com/microsoft/demikernel/issues/1065
+    Ok(())
```
ppenna, line 168: I fix this in https://github.com/microsoft/demikernel/pull/1074.

src/rust/inetstack/protocols/tcp/tests/setup.rs
```diff
@@ -227,10 +226,10 @@ fn test_refuse_connection_early_ack() -> Result<()> {
...
```
ppenna, line 232: I fix this in https://github.com/microsoft/demikernel/pull/1074.

src/rust/inetstack/protocols/tcp/tests/setup.rs
```diff
@@ -301,10 +300,10 @@ fn test_refuse_connection_missing_syn() -> Result<()> {
...
```
ppenna, line 306: I fix this in https://github.com/microsoft/demikernel/pull/1074.

src/rust/inetstack/protocols/tcp/queue.rs
```diff
@@ -418,18 +437,26 @@ impl<const N: usize> SharedTcpQueue<N> {
...
+        }
+
+        // If this is an inactive socket, then generate a RST segment.
+        // Generate the RST segment according to the ACK field.
         // If the incoming segment has an ACK field, the reset takes its
         // sequence number from the ACK field of the segment, otherwise the
         // reset has sequence number zero and the ACK field is set to the sum
         // of the sequence number and segment length of the incoming segment.
         // Reference: https://datatracker.ietf.org/doc/html/rfc793#section-3.4
+
```
ppenna, line 459: Could you remove this line?

src/rust/inetstack/protocols/tcp/passive_open.rs
```diff
@@ -140,44 +148,33 @@ impl<const N: usize> SharedPassiveSocket<N> {
...
-                    },
-                    _ => continue,
-                }
+    async fn poll(mut self, yielder: Yielder) {
+        loop {
+            let (ipv4_hdr, tcp_hdr, buf) = match self.recv_queue.pop(&yielder).await {
+                Ok(result) => result,
+                Err(_) => break,
+            };
+            let remote = SocketAddrV4::new(ipv4_hdr.get_src_addr(), tcp_hdr.src_port);
```
ppenna, line 157: Missing type annotation.

src/rust/inetstack/protocols/tcp/passive_open.rs
```diff
@@ -140,44 +148,33 @@ impl<const N: usize> SharedPassiveSocket<N> {
...
 
-        // If not a SYN, then this packet is not for a new connection and we throw it away.
-        if !header.syn || header.ack || header.rst {
-            return Err(Fail::new(EBADMSG, "invalid flags"));
+            // If not a SYN, then this packet is not for a new connection and we throw it away.
+            if !tcp_hdr.syn || tcp_hdr.ack || tcp_hdr.rst {
+                let cause = "invalid flags";
+                warn!("{}", cause);
+                continue;
+            }
```
ppenna, line 169: We should send a reset on this case, I will change https://github.com/microsoft/demikernel/pull/1074 for doing so.

src/rust/inetstack/protocols/tcp/passive_open.rs
```diff
@@ -186,26 +183,108 @@ impl<const N: usize> SharedPassiveSocket<N> {
...
+            debug!("receive(): sending RST (local={:?}, remote={:?})", self.local, remote);
+            let (seq_num, ack_num): (SeqNumber, Option<SeqNumber>) = if tcp_hdr.ack {
+                (tcp_hdr.ack_num, None)
+            } else {
+                (
+                    SeqNumber::from(0),
+                    Some(tcp_hdr.seq_num + SeqNumber::from(tcp_hdr.compute_size() as u32)),
+                )
+            };
+            let local: SocketAddrV4 = self.local;
```
ppenna, line 195: This is duplicate now.

src/rust/inetstack/protocols/tcp/passive_open.rs
```diff
@@ -186,26 +183,108 @@ impl<const N: usize> SharedPassiveSocket<N> {
...
+            self.send_rst(&local, &remote, seq_num, ack_num);
+            return;
```
ppenna, line 197: No need for return statement here.

src/rust/collections/async_queue.rs
iyzhang: Oh, it seemed to require that T be clonable as well. Not sure why

src/rust/inetstack/protocols/tcp/tests/setup.rs
iyzhang: Great! I'll take a look

src/rust/inetstack/protocols/tcp/tests/setup.rs
iyzhang: Great!

src/rust/inetstack/protocols/tcp/passive_open.rs
iyzhang: Perfect, thanks!

src/rust/inetstack/protocols/tcp/passive_open.rs
iyzhang: ok sounds good.

ppenna (approved):
LGTM

PR #1070, opened 2023-12-13
[catmem] Move pending_ops & pack_result to runtime

anandbonde:
This pr moves the pending_ops & pack_result to runtime.

### DIFF
--- a/src/rust/catmem/queue.rs
+++ b/src/rust/catmem/queue.rs
@@ -18,7 +18,6 @@ use crate::{
         scheduler::{
             TaskHandle,
             Yielder,
-            YielderHandle,
         },
         DemiRuntime,
         QToken,
@@ -28,7 +27,6 @@ use crate::{
 };
 use ::std::{
     any::Any,
-    collections::HashMap,
     ops::{
         Deref,
         DerefMut,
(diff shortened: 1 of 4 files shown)


src/rust/catmem/queue.rs
```diff
@@ -109,7 +102,6 @@ impl SharedCatmemQueue {
             }
         }
         self.ring.prepare_closed()?;
-        self.cancel_pending_ops(Fail::new(libc::ECANCELED, "this queue was closed"));
```
ppenna, line 112: Where is cancelletaion happening now? I missed this :)

ppenna (approved):
LGTM

src/rust/catmem/queue.rs
anandbonde: This now happens at a centralized place in this function: https://github.com/microsoft/demikernel/blob/2fa835968cee1db9025a5e41efe2bc80236f19a1/src/rust/runtime/mod.rs#L228.

This change went in earlier in another PR. :)

ppenna: Yes, that is what I was thinking. Thank you for the reference :)

PR #1074, opened 2023-12-13
[inetstack] Handle Reset Generation Correctly

ppenna:
## Description

This PR closes https://github.com/microsoft/demikernel/issues/235.

### DIFF
--- a/src/rust/inetstack/protocols/tcp/queue.rs
+++ b/src/rust/inetstack/protocols/tcp/queue.rs
@@ -9,20 +9,12 @@ use crate::{
     collections::async_queue::SharedAsyncQueue,
     inetstack::{
         protocols::{
-            ethernet2::{
-                EtherType2,
-                Ethernet2Header,
-            },
-            ip::IpProtocol,
             ipv4::Ipv4Header,
             tcp::{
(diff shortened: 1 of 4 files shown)


src/rust/inetstack/protocols/tcp/queue.rs
```diff
@@ -435,81 +427,12 @@ impl<const N: usize> SharedTcpQueue<N> {
...
-        remote: SocketAddrV4,
-        buf: DemiBuffer,
-    ) {
+    pub fn receive(&mut self, ip_hdr: Ipv4Header, tcp_hdr: TcpHeader, buf: DemiBuffer) {
         // If this queue has an allocated receive queue, then direct the packet there.
         if let Some(recv_queue) = self.recv_queue.as_mut() {
             recv_queue.push((ip_hdr, tcp_hdr, buf));
             return;
         }
-
```
iyzhang: You still need to generate a reset here because we don't have a socket data structure to do further processing. I will eventually just move PassiveOpenSocket to this file as well (or at least squash it all down to one receive queue).

iyzhang (approved):
Great! Thanks for doing this and closing 1065 as well. I'm glad it all merged well.

PR #1075, opened 2023-12-13
[catcollar] Enable tracking of pending_ops

anandbonde:
Enable tracking of pending_ops in catcollar. This is already done for other liboses, so just onboarding catcollar is straighforward. Cleanup (cancellation of pending_ops) happens in pack_result automatically.


ppenna (approved):
LGTM

PR #1076, opened 2023-12-14
[inetstack] Track pending_op for async_close() in udp peer

anandbonde:
This PR enables tracking of pending_ops for the async_close() path in the UDP peer in the intestack.


ppenna (approved):
LGTM

PR #1078, opened 2023-12-14
[test] Modular Unit Tests

ppenna:
## Description

This PR closes https://github.com/microsoft/demikernel/issues/1077


linux.mk
anandbonde, line 1: We should change this for Windows as well.

anandbonde (approved):
Looks good. It would be good to have this change enabled for Windows too (ok at a later point as well).

linux.mk
ppenna: Sure. It is now fixed.

PR #1079, opened 2023-12-14
[test] Drop Unit Tests

ppenna:
## Description

This PR drops the following unit tests, in favor of simulation tests based on network packet simulation:
- `test_good_connect`
- `test_connection_timeout`


iyzhang (approved):
Thanks, it's great to see these tests being removed!

PR #1080, opened 2023-12-14
[test] Support Reset Packages

ppenna:
## Description

This PR adds supports for RST packets in the network simulator engine.


anandbonde (approved):
Looks good.

Issue #988, opened 2023-11-01
[inetstack] Clean up background tasks for established connections on close


iyzhang:
## Context

Currently, we only have a synchronous close in the inetstack which does not give us a chance to clean up resources after the TCP close protocol finishes. In particular, we have lingering background tasks that still may be signaled by arriving packets.

## Proposed Solution

We should move entirely to asynchronous close, wait until the close protocol completes and then clean up all meta-data and background tasks and other resources.

PR #1082, opened 2023-12-15
[inetstack] Move TCP close to a linear flow and remove synchronous close

iyzhang:
This PR moves the TCP close protocol to a linear flow and cleans up background tasks once we move out of the established state, closing #988

### DIFF
--- a/src/rust/inetstack/protocols/tcp/established/ctrlblk.rs
+++ b/src/rust/inetstack/protocols/tcp/established/ctrlblk.rs
@@ -13,9 +13,9 @@ use super::{
     },
 };
 use crate::{
-    collections::{
-        async_queue::AsyncQueue,
-        async_value::AsyncValue,
+    collections::async_queue::{
+        AsyncQueue,
+        SharedAsyncQueue,
     },
     inetstack::protocols::{
         arp::SharedArpPeer,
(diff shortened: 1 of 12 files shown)


src/rust/inetstack/protocols/tcp/established/ctrlblk.rs
```diff
@@ -389,26 +390,105 @@ impl<const N: usize> SharedControlBlock<N> {
...
+
+    // This is the main TCP processing routine.
+    pub async fn poll(&mut self, yielder: Yielder) -> Result<!, Fail> {
+        // Normal data processing in the Established state.
+        loop {
+            let (header, data) = match self.recv_queue.pop(&yielder).await {
+                Ok((_, header, data)) if self.state == State::Established => (header, data),
+                Ok(result) => {
+                    self.recv_queue.push_front(result);
+                    return Err(Fail::new(
```
ppenna, line 405: Log an `error` message here.

src/rust/inetstack/protocols/tcp/established/ctrlblk.rs
```diff
@@ -695,75 +739,13 @@ impl<const N: usize> SharedControlBlock<N> {
...
-            } else {
-                // We already owe our peer an ACK (the timer was already running), so cancel the timer and ACK now.
-                self.ack_deadline.set(None);
-                self.send_ack();
-            }
-        }
+        // We can only legitimately receive data in ESTABLISHED, FIN-WAIT-1, and FIN-WAIT-2.
+        assert!(self.state == State::Established || self.state == State::FinWait1 || self.state == State::FinWait2);
+        self.receive_data(seg_start, data);
+        Ok(())
```
ppenna, line 748: We should still use the match statement. A malfunctioning peer may send us data after the fin, and we don't want to panic, but rather drop the packet.

iyzhang: If you look at the code flow, it is no longer possible to enter this function unless we are in the established state, so I removed the assert but certainly I could put a debug assert back.

ppenna (approved):
Thanks for working on this!

PR #1088, opened 2023-12-15
[runtime] Enhancement: Adding transport trait

iyzhang:
This PR adds a transport trait for high level network transport. In contrast to our network runtime, which abstracts the lowest level network API (i.e., only raw ethernet send and receive), the NetworkTransport trait abstracts a high-level network API including sockets. Currently I have hard coded the SharedCatnapTransport as the only transport used but eventually we should move all of our network libOSes to using this trait and sharing their code at the Cat*LibOS and Cat*Queue level


src/rust/catnap/linux/transport.rs
```diff
@@ -368,39 +362,13 @@ impl SharedSocketData {
...
-            .expect("should be able to insert background coroutine");
-        me
-    }
-
     /// This function registers a handler for incoming and outgoing I/O on the socket. There should only be one of
     /// these per socket.
-    fn register_epoll(&mut self, sd: &SocketDescriptor, events: u32) -> Result<(), Fail> {
+    fn register_epoll(
+        &mut self,
+        sd: &<SharedCatnapTransport as NetworkTransport>::SocketDescriptor,
```
anandbonde, line 369: Is it possible to move the `<SharedCatnapTransport as NetworkTransport>::SocketDescriptor` to the top of the file in a `use` expression?

src/rust/catnap/linux/transport.rs
```diff
@@ -504,8 +476,151 @@ impl SharedCatnapTransport {
...
+        let yielder: Yielder = Yielder::new();
+        let background_task: YielderHandle = yielder.get_handle();
+        let me: Self = Self(SharedObject::new(CatnapTransport {
+            epoll_fd,
+            socket_table: Slab::<SharedSocketData>::new(),
+            background_task,
+        }));
+        let mut me2: Self = me.clone();
+        runtime
+            .insert_background_coroutine(
```
anandbonde, line 614: The background coroutine is woken up with error in the drop() fn and the poll coroutine exits on any error on yielder await. Should the insert_background_coroutine() be paired with remove_background_coroutine() too? (for proper cleanup inside the scheduler)

src/rust/runtime/network/transport.rs
```diff
@@ -0,0 +1,84 @@
...
+    fn socket(&mut self, domain: Domain, typ: Type) -> Result<Self::SocketDescriptor, Fail>;
+
+    /// Bind an address to the socket.
+    fn bind(&mut self, sd: &mut Self::SocketDescriptor, local: SocketAddr) -> Result<(), Fail>;
+
+    /// Listen on this socket in the network transport layer.
+    fn listen(&mut self, sd: &mut Self::SocketDescriptor, backlog: usize) -> Result<(), Fail>;
+
+    /// Forcibly close this socket in the network transport layer. This function should only be used in Drop and other
+    /// internal functions, never exposed to the application.
```
anandbonde, line 42: Can this (not to be exposted outside) be controlled/restricted at compile-time? Maybe be structuring differently (or hierarchy etc.)?

src/rust/catnap/linux/transport.rs
iyzhang: Sure, I couldn't remember if background coroutines were always removed once they exited. But I still want to wake the background coroutine so that it can do any clean up needed.

src/rust/runtime/network/transport.rs
iyzhang: Yes, I believe  since it is not exposed through the demikernel libos layer, then external apps cannot use it. I don't know if we can enforce that we do not expose it in some other way.

src/rust/catnap/linux/transport.rs
iyzhang: Sure, I didn't know that we could do that. Thanks!

anandbonde (approved):
Looks good.

PR #1089, opened 2023-12-18
[demikernel] Consolidate yielder creation for all LibOS coroutines

anandbonde:
This PR removes the Yielder and YielderHandle creation from the LibOS level and now does it at a centralized place in the runtime.

It only focuses on the LibOS level.
- This PR is not applicable for background coroutines.
- Does not cover the complete inetstack.


ppenna (approved):
LGTM

PR #1099, opened 2023-12-19
dpdk-rs Enhancement: update intrinsics flags for bindgen and inline

tylerretzlaff:
* Add the -mavx flag to build of the inline translation unit to match the flag passed to clang for bindgen.

* Add the rtm and cldemote flags for both bindgen and inline to make visible respective intrinsics in the application namespace. This allows build.rs to be robust to DPDK that may have intrinsics option enabled instead of inline asm.


anandbonde (approved):
Looks good.

anandbonde:
PR #1500 supersedes this PR, so closing this one.

PR #1101, opened 2023-12-19
[dpdk-rs] Enhancement: link to dpdk librte_log

tylerretzlaff:
As of the dpdk 23.11 release the logging facilities of dpdk have been split out of librte_eal and placed in its own library. the new library must now be added when linking to avoid link failure.


anandbonde:
Closing this PR now because this change cannot be applied on current version of DPDK that Demikernel uses. We will bring back this PR when we upgrade (if this change is needed).

PR #1103, opened 2023-12-20
[inetstack] Wait ACK on Push

ppenna:
## Description

This PR closes https://github.com/microsoft/demikernel/issues/363.

## Summary of Changes

- Drop deprecated unit-tests in favor of simulation tests
- Introduces an async queue to block push operations until the remote peer ACKs sent data.


src/rust/inetstack/protocols/tcp/passive_open.rs
```diff
@@ -327,6 +330,7 @@ impl SharedPassiveSocket {
                 .clone()
                 .wait_for_ack(
                     recv_queue.clone(),
+                    ack_queue.clone(),
```
iyzhang, line 333: Also do not need a clone here.

src/rust/inetstack/protocols/tcp/passive_open.rs
```diff
@@ -455,6 +460,7 @@ impl SharedPassiveSocket {
             self.runtime.clone(),
             self.transport.clone(),
             recv_queue.clone(),
+            ack_queue.clone(),
```
iyzhang, line 463: Also no need for clone here.

src/rust/inetstack/protocols/tcp/queue.rs
```diff
@@ -235,6 +236,7 @@ impl SharedTcpQueue {
             self.runtime.clone(),
             self.transport.clone(),
             recv_queue.clone(),
+            ack_queue.clone(),
```
iyzhang, line 239: no need to clone here

iyzhang (commented):
Thanks for doing this!

iyzhang (approved):

src/rust/inetstack/protocols/tcp/passive_open.rs
ppenna: Actually it does, because it is in a look and value would be moved.

src/rust/inetstack/protocols/tcp/passive_open.rs
```diff
@@ -201,12 +201,14 @@ impl SharedPassiveSocket {
...
         let recv_queue: SharedAsyncQueue<(Ipv4Header, TcpHeader, DemiBuffer)> =
             SharedAsyncQueue::<(Ipv4Header, TcpHeader, DemiBuffer)>::default();
+        let ack_queue: SharedAsyncQueue<usize> = SharedAsyncQueue::<usize>::default();
         let future = self.clone().send_syn_ack_and_wait_for_ack(
             remote,
             remote_isn,
             local_isn,
             tcp_hdr,
             recv_queue.clone(),
+            ack_queue.clone(),
```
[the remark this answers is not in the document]
ppenna: Sure.

src/rust/inetstack/protocols/tcp/queue.rs
ppenna: Sure.

src/rust/inetstack/protocols/tcp/passive_open.rs
iyzhang: ah good point.

PR #1104, opened 2023-12-20
[demikernel] Remove TaskHandle (and other simplifications)

anandbonde:
This PR removes the TaskHandle from the codebase.

### DIFF
--- a/src/rust/demikernel/libos/network/mod.rs
+++ b/src/rust/demikernel/libos/network/mod.rs
@@ -263,18 +263,18 @@ impl NetworkLibOS {
     }
 
     /// Waits for any operation in an I/O queue.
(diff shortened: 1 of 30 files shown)


src/rust/demikernel/libos/network/mod.rs
```diff
@@ -13,13 +13,13 @@ pub mod queue;
 // Imports
 //======================================================================================================================
 
+#[cfg(all(feature = "catnap-libos"))]
```
iyzhang, line 16: I believe this will cause a warning on all of the other libOSes that have not moved yet

iyzhang (approved):

src/rust/demikernel/libos/network/mod.rs
anandbonde: Added this condition to avoid warning on other LibOSes.

PR #1107, opened 2023-12-20
[dpdk-rs] Workaround: redefine RTE_TOOLCHAIN_XXX for host toolchain

tylerretzlaff:
there is a bug in dpdk headers where the toolchain preprocessor definition used to build dpdk is retained and then incorrectly directs conditional compilation during application builds.

redefine the RTE_TOOLCHAIN_XXX macros as appropriate using the host toolchain predefined macros to get correct conditional evaluation when consuming dpdk headers.

this is just a workaround for now, i will work with the dpdk community to upstream a fix which will eventually allow this change to be removed.

### DIFF
--- a/dpdk-rs/inlined.c
+++ b/dpdk-rs/inlined.c
@@ -3,6 +3,35 @@
  * Licensed under the MIT license.
  */
 
+#include <rte_config.h>
+
+/*
+ * work around bug in dpdk headers where the toolchain preprocessor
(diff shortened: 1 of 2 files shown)


tylerretzlaff:
there is an assumption here that the toolchains used to build demikernel are in the set { gcc, clang, msvc } if there are others the change needs to be enlightened.

tylerretzlaff:
abandoning this pr since RFC patch has been submitted to dpdk

PR #1109, opened 2023-12-21
[inetstack] Move to NetworkTransport trait

### LINKED ISSUE #889: Remove deprecated function `wait2()`
## Description
Warnings due to deprecated functions was causing issues in Bing's build integration with Demikernel.

## Temporary solution until this issue is fixed
Deprecated tag removed.
FIXME comment added in place.

Partially fixes: #878

iyzhang:
This PR moves the inetstack into the network transport architecture. It also closes #889. The main purpose of this PR is to move our Catnip and Catpowder libOSes to an architecture where the inetstack is: 1) a network transport and fulfills the NetworkTransport trait *and* 2) the inetstack is paramaterized on the NetworkRuntime, which represents the lowest level network API. This includes the following changes.

* The NetworkRuntime trait now requires the MemoryRuntime trait because the lowest level network API should dictate how memory is allocated. We have a default implementation in the NetworkRuntime which uses the default MemoryRuntime to allocate memory from the heap.
* We have moved the `new` function for the transport out of the NetworkRuntime trait, so that different transports can take different parameters on construction. The NetworkLibOS now requires a fully constructed transport be passed in instead.
* We have added a get_runtime() function to the transport trait to let the NetworkLibOS and the testing infrastructure access the underlying SharedDemiRuntime. This lets us not have to keep a separate reference to the SharedDemiRuntime but instead get it from the transport as we need.

The last 5 are roughly in this order:
1. Move the inetstack to use the transport trait. This includes adding the NetworkRuntime as a generic to the inetstack and then passing in the appropriate network runtime. We also remove all of the infrastructure around Demikernel queues and states.
2. Move all of the testing infrastructure to the transport trait. The DummyLibOS is now a transport as is the Engine in the unit tests.
3.  Remove the Catnip LibOS. It is now just a parameterized transport on top of the SharedDPDKRuntime.
4. Remove the Catpowder LibOS. It is now just a parameterized transport on top of the LinuxRuntime.
5. Move the libOS level to allocating Catnip and Catpowder as a NetworkLibOS with the SharedInetstack as a transport and the appropriate NetworkRuntime.

The plan in the future is to move Catcollar and Catloop to the transport architecture as well so that we can stop maintaining multiple copies of code at the NetworkLibOS level and the NetworkQueue level. If necessary, we can parameterize the NetworkQueues instead of the NetworkLibOS to let multiple transports co-exist. We should also eventually merge the code in demikernel/libos/network/mod.rs and lib.rs


ppenna (changes requested):
Simulation tests do not pass because the simulator itself is sensitive on when co-routines are polled, when time is advanced and the interaction of these two.

I understand that the end goal is to not rely on this, but at the same time, the current code base is not ready for that change.

I suggest that we have a discussion to decide how to move forward with this PR.

iyzhang:
> Simulation tests do not pass because the simulator itself is sensitive on when co-routines are polled, when time is advanced and the interaction of these two.
> 
> I understand that the end goal is to not rely on this, but at the same time, the current code base is not ready for that change.
> 
> I suggest that we have a discussion to decide how to move forward with this PR.

Yes, I agree I was too ambitious trying to also fix some issues with the scheduler. I have rolled back the last commit, which I believe is stable and we can discuss further when we return in January.

src/rust/inetstack/protocols/tcp/tests/simulator.rs
```diff
@@ -349,6 +342,7 @@ impl Simulation {
             },
         }
 
+        self.engine.poll();
```
ppenna, line 345: Should we add this here?

iyzhang: Because our usual scheduling policy in the libOS level is to call poll immediately after each sys call. Since this PR moves the send to a coroutine, rather than doing it in the synchronous path, to better align with the other libOSes, we need to poll immediately after issuing the push.

iyzhang: Ok, I have added a single poll just to the push system call. It also lets us preserve the push pop trace as before.

ppenna (approved):

PR #1112, opened 2024-01-02
[liburing-rs] Moving `liburing-rs` Into Source tree

ppenna:
## Description

This PR moves the `liburing-rs` crate into Demikernel's source tree.


iyzhang:
Do we have a reason for doing this? I noticed that we've been moving towards a larger and larger single Git repo, and I wasn't sure what the rationale for that is.

ppenna:
> Do we have a reason for doing this? I noticed that we've been moving towards a larger and larger single Git repo, and I wasn't sure what the rationale for that is.

You are correct. We are now moving back to a single repository.

Rationale:
- Updating individual components is unlikely to introduce breaking changes.
- Keep versions on third-party crates unified.
- Easier to manage (maintaining several repos does consume my engineering time)
- microsoft/demikernel is now the official repository for all Demikernel artifacts, and having everything in one place will attract more people.

iyzhang (approved):
LGTM

PR #1117, opened 2024-01-08
[inetstack] Clarifications of Port Scan Attack

ppenna:
## Description

This PR closes https://github.com/microsoft/demikernel/issues/260.


iyzhang (approved):
Thanks for this comment

PR #1118, opened 2024-01-09
[runtime] Move to FusedFuture

iyzhang:
This PR moves the scheduler and tasks to using FusedFutures instead of the basic Futures. Our scheduler notes the first time that a Task returns Ready and marks it as completed and never schedules it again. So if the future is not fused, it'll still never get polled again. [See this line of code](https://github.com/microsoft/demikernel/blob/647d0200e3266196d6adbedac703e7b79e198723/src/rust/runtime/scheduler/page/page.rs#L60).

Furthermore, our TaskWithResult type will have stored the output of the coroutine and will never invoke it again.[See here](https://github.com/microsoft/demikernel/blob/647d0200e3266196d6adbedac703e7b79e198723/src/rust/runtime/scheduler/task.rs#L101).

Thus, we should use FusedFutures to enforce that none of the futures that our scheduler operate on are able to run again after the first time the future returns Ready.

### DIFF
--- a/src/rust/catcollar/mod.rs
+++ b/src/rust/catcollar/mod.rs
@@ -53,6 +53,7 @@ use crate::{
         SharedDemiRuntime,
     },
 };
+use ::futures::FutureExt;
 use ::std::{
     mem,
     net::{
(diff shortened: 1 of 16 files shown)


ppenna (approved):
Thanks for working on this.

Indeed it is a great improvement :)

I wonder if we should deprecate `take_notified()`, else turn it into a sanity check?

PR #1119, opened 2024-01-09
[inetstack] Free Ephemeral Ports Once Connection is Closed

ppenna:
## Description

This PR closes https://github.com/microsoft/demikernel/issues/236

## Summary of Changes

Added required logic to rollback ephemeral port allocation.

### DIFF
--- a/src/rust/demikernel/libos/network/libos.rs
+++ b/src/rust/demikernel/libos/network/libos.rs
@@ -152,7 +152,6 @@ impl<T: NetworkTransport> SharedNetworkLibOS<T> {
 
         // Issue bind operation.
         if let Err(e) = self.get_shared_queue(&qd)?.bind(local) {
-            trace!("error");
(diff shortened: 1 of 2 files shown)


iyzhang (changes requested):
LGTM. Can you add a test for this?

ppenna:
> LGTM. Can you add a test for this?

@iyzhang this is currently not possible because:
- This test requires allocating and releasing 16384 ephemeral ports, to force an overflow on the table.
- This requires 16384 connections to be open and closed using our TCP stack.
- The integration test `tcp-close` exercises this scenario in a small scale.
- It is not all about increasing the scale, because this test is disabled for Catnip and Catpowder.
- `tcp-close` is disabled for Catnip and Catpowder because there are other bugs in the connection termination that should be addressed first.

iyzhang (approved):

PR #1123, opened 2024-01-11
[collections] Enhancement: External to internal ID map

iyzhang:
This PR adds a new data structure that is explicitly for us to enforce ID translations between internal data structures and external ones. Generally this takes the form of translating from something that we may hand to the application (e.g., a queue token) to an internal representation (e.g., an offset into a slab). Since slab offsets are often allocated sequentially, we do not want external applications or even the libOSes to depends on their specific characteristics. 

To deal with this, we introduce an IdMap data structure. It is explicitly used for translating between external and internal identifiers. On debug builds, this data structure randomly generates IDs and uses a hash map to map external to internal identifiers to enforce that there are no weird dependencies on IDs. On release builds, we discard the map and directly use the internal identifier.

Currently, I'm using a simple flag to perform this switch because it seemed easier to read and the compiler will remove the unused code but we could use more explicit compiler blocks later.

These results are from running the benchmark_insert() test from scheduler.rs
| | Debug | Release|
|----|-----|----|
|Indirection|1228 ns|311 ns|
|Direct Mapping|408 ns|141 ns|

This PR also removes the limitation on the number of coroutines that we can schedule since we have completely moved to 64-bit identifiers, there should be no issues with collisions.

### DIFF
--- a/src/rust/collections/id_map.rs
+++ b/src/rust/collections/id_map.rs
@@ -1,0 +1,154 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+//======================================================================================================================
+// Imports
+//======================================================================================================================
+
+use ::rand::{
+    rngs::SmallRng,
+    RngCore,
+    SeedableRng,
+};
+use ::std::{
+    collections::HashMap,
+    convert::From,
+    hash::Hash,
+};
+
+//======================================================================================================================
+// Constants
+//======================================================================================================================
+
+/// Performance note: These two flags were benchmarked with the scheduler insert benchmark on a
+/// release build and have the following impact on performance. The number is the average of 5 runs of the test and
+/// performed on an Azure Standard D16ds v4 (16 vcpus, 64 GiB memory) VM running Linux (ubuntu 22.04).
+/// Direct vs indirect mapping: 152 for direct, indirect is below.
+/// Randomize vs non random: 332ns vs 154ns
+
+/// This flag controls whether we actually use a mapping or just directly expose internal IDs.
+/// We should eventually set this using an environment variable.
+const DIRECT_MAPPING: bool = false;
+/// This flag controls how the ids are allocated, either randomly or in a Fibonacci sequence.
+#[cfg(debug_assertions)]
+const RANDOMIZE: bool = true;
+#[cfg(not(debug_assertions))]
+const RANDOMIZE: bool = false;
+
+/// Arbitrary size chosen to pre-allocate the hashmap. This improves performance by 6ns on average on our scheduler
+/// insert benchmark.
+const DEFAULT_SIZE: usize = 1024;
+
+/// Seed for the random number generator used to generate tokens.
+/// This value was chosen arbitrarily.
+const SCHEDULER_SEED: u64 = 42;
+const MAX_RETRIES_ID_ALLOC: usize = 500;
+
+//======================================================================================================================
+// Structures
+//======================================================================================================================
+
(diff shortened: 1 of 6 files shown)


src/rust/collections/id_map.rs
```diff
@@ -0,0 +1,154 @@
...
+                self.current_id + self.last_id
+            } else {
+                self.last_id - (u64::MAX - self.current_id)
+            };
+            self.last_id = id;
+            let external_id: E = E::from(id);
+            if self.ids.insert(external_id, internal_id).is_some() {
+                panic!("Should not have a previous task with this id");
+            }
+            external_id
```
ppenna, line 134: What are the motivations for using a Fibonnacci sequence in contrast to a simple arithmethic progression with step 1 starting at 0?

src/rust/collections/pin_slab.rs
```diff
@@ -114,6 +120,25 @@ impl<T> PinSlab<T> {
         }
     }
 
+    /// Get a reference to the value at the given slot.
+    #[inline(always)]
+    unsafe fn internal_get(&self, key: usize) -> Option<&T> {
```
ppenna, line 125: Why is this function unsafe? Is it possible to reduce the unsafe scope? Could you add a note on why it is safe to call the unsafe scope?

src/rust/runtime/scheduler/scheduler.rs
```diff
@@ -229,13 +199,15 @@ impl Scheduler {
...
-    pub fn has_completed(&self, task_id: u64) -> Option<bool> {
-        let pin_slab_index: usize = match self.task_ids.get(&task_id) {
-            Some(pin_slab_index) => *pin_slab_index,
+    pub fn has_completed(&self, task_id: TaskId) -> Option<bool> {
+        let pin_slab_index: InternalId = match self.task_ids.get(&task_id) {
+            Some(pin_slab_index) => pin_slab_index,
             None => return None,
         };
+
+        // Add a check here for whether the index is valid.
```
ppenna, line 208: Should this be a TODO comment?

src/rust/runtime/scheduler/scheduler.rs
```diff
@@ -43,31 +40,24 @@ use ::std::{
...
     /// Stores all the tasks that are held by the scheduler.
     tasks: PinSlab<Box<dyn Task>>,
-    /// Maps between externally meaningful ids and the index of the task in the slab.
-    task_ids: HashMap<u64, usize>,
     /// Holds the waker bits for controlling task scheduling.
     waker_page_refs: Vec<WakerPageRef>,
-    /// Small random number generator for tokens.
-    rng: SmallRng,
+    /// Holds the total number of running tasks.
+    total_num_tasks: usize,
```
ppenna, line 60: What is the purpose of this variable? I see that it is used on test code only.

ppenna (approved):
Looks good for me.

@iyzhang, please address my comments before merging.

src/rust/runtime/scheduler/scheduler.rs
iyzhang: Sure, let me take it out of here because it is only needed for direct mapping.

src/rust/collections/pin_slab.rs
iyzhang: I have limited the unsafe block to the only small bit of unsafe code. For symmetry, I have also updated internal_get_mut.

src/rust/runtime/scheduler/scheduler.rs
iyzhang: Sorry this was a lingering comment from before. I have moved it now

PR #1130, opened 2024-01-17
[runtime] Enhancement: Add scheduler groups

iyzhang:
This PR adds the ability to create new task groups and assign tasks to them. Eventually we will likely create one task group for background coroutines and then one for each Demikernel queue. However, currently we only have one default group and we add all tasks to this group.

For our two benchmark tests, here are the changes in performance for a release build. Each run 5 times and averaged.

| | dev | this PR|
|--|--|--|
|benchmark_insert|330ns| 444ns|
|benchmark_poll|33ns|29ns|

Insertion has gone up 35%, however polling has gone down 12%. This seems like a reasonable trade-off as polling is done more frequently. I suspect the polling changes are due to polling the futures in a tighter loop after collecting all of the ready bits. I did not think it was worth digging deeper into the costs, but I believe we can optimize insertion further in the future.

### DIFF
--- a/src/rust/catnap/win/overlapped.rs
+++ b/src/rust/catnap/win/overlapped.rs
@@ -404,8 +404,8 @@ mod tests {
     use crate::{
         ensure_eq,
         runtime::scheduler::{
-            scheduler::TaskId,
             Scheduler,
+            TaskId,
             TaskWithResult,
         },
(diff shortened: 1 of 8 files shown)


ppenna (approved):
OK

Issue #237, opened 2021-07-02
[inetstack] Implement `close()` for Passive Sockets


ppenna:
# Description

We should implement `close()` for passive sockets.

Related Issues
----------------

- https://github.com/demikernel/catnip/issues/82

iyzhang:
This is done

PR #1131, opened 2024-01-17
[inetstack] Close for Passive Sockets

### LINKED ISSUE #206: [inetstack] PassiveSockets are never cleaned up
Description
=========

`PassiveSocket`s are held inside the `TcpPeer`'s `passive` field. These are never removed from this field though. They should probably be removed once `close` is called on the corresponding socket.

Related Issues
==========

- https://github.com/demikernel/catnip/issues/48
- https://github.com/demikernel/catnip/issues/82

ppenna:
## Description

This PR closes #206 and #237.

### DIFF
--- a/src/rust/inetstack/protocols/tcp/passive_open.rs
+++ b/src/rust/inetstack/protocols/tcp/passive_open.rs
@@ -77,7 +77,17 @@ use ::std::{
 // Structures
 //======================================================================================================================
 
+/// States of a passive socket.
+enum State {
+    /// The socket is listening for new connections.
+    Listening,
+    /// The socket is closed.
+    Closed,
+}
+
 pub struct PassiveSocket<N: NetworkRuntime> {
+    // TCP Connection State.
+    state: State,
     connections: HashMap<SocketAddrV4, SharedAsyncQueue<(Ipv4Header, TcpHeader, DemiBuffer)>>,
     recv_queue: SharedAsyncQueue<(Ipv4Header, TcpHeader, DemiBuffer)>,
     ready: AsyncQueue<Result<EstablishedSocket<N>, Fail>>,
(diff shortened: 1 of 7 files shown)


src/rust/inetstack/protocols/tcp/passive_open.rs
```diff
@@ -147,8 +166,32 @@ impl<N: NetworkRuntime> SharedPassiveSocket<N> {
...
+    // Closes the target socket.
+    pub fn close(&mut self) -> Result<(), Fail> {
+        self.state = State::Closed;
+        Ok(())
+    }
+
     async fn poll(mut self, yielder: Yielder) {
         loop {
+            match self.socket_rx.try_next() {
+                Ok(Some(socket)) => match self.connections.remove(&socket) {
```
iyzhang, line 178: This particular check doesn't seem necessary to implement passive close. If the passive socket is closed, we will simply drop all packets headed to that address. So I would suggest that you do not need to introduce a mpsc channel here at all.

iyzhang (commented):
I would suggest that we either get rid of the communication channel between the active socket and the passive socket or switch it to an AsyncQueue and a select_bias there because the passive socket will not be woken if the active socket is closed right now.

src/rust/inetstack/protocols/tcp/passive_open.rs
```diff
@@ -147,8 +166,32 @@ impl<N: NetworkRuntime> SharedPassiveSocket<N> {
...
 
+    // Closes the target socket.
+    pub fn close(&mut self) -> Result<(), Fail> {
+        self.state = State::Closed;
+        Ok(())
+    }
+
     async fn poll(mut self, yielder: Yielder) {
         loop {
+            match self.socket_rx.try_next() {
```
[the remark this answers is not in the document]
ppenna: I believe that I am missing something here. How is this code related to active sockets in the local peer?

src/rust/inetstack/protocols/tcp/passive_open.rs
ppenna: Unfortunately, we need this, because the `recv_queue` is kept on the table and thus its drop logic does not run.

iyzhang (approved):

PR #1132, opened 2024-01-17
[runtime] Enhancement: Scheduler clean up

iyzhang:
This PR just cleans up some things around the scheduler.
1. Remove the handle.rs file. We only have a yielder handle for now and that can be moved to the yielder.rs file
2. Remove use of Rc Refcell from yielder handle in favor of our SharedObject
3. Remove an unnecessarily public function (insert) from the runtime.


ppenna (approved):
LGTM

Issue #1128, opened 2024-01-16
[runtime] Remove poll and then check completion


iyzhang:
## Context

Currently we poll the scheduler and then check for each qtoken whether it has completed. This add a lookup in our hash map that is unnecessary.

## Proposed Solution

We should immediately return all completed tasks from poll and let the runtime decide what to do with them. This also eliminates the need to remove tasks from the scheduler. Instead, we could remove completed tasks and eliminate the need for completion bits. However, it would require us to ensure that there are no lingering yielders for the coroutine as they might set a ready bit for a new task.

PR #1136, opened 2024-01-19
[libos] move wait functions into libos

iyzhang:
This PR moves the wait_* into the network and memory libOSes so that we can eventually get rid of poll and get_result. See #1128


ppenna (approved):
I'm not sure what is the end goal here, but this looks OK.

PR #1137, opened 2024-01-20
[runtime] Remove poll and check completed in scheduler

### LINKED ISSUE #1128: [runtime] Remove poll and then check completion

iyzhang:
This PR closes #1128 . We move away from a polling model to directly return completed tasks after running the scheduler. We also immediately remove completed tasks (since they will never run again) and it is the responsibility of the higher-level runtime to stash them somewhere. Thus, now the runtime contains a wait function that runs the scheduler until the requested queue token has completed. This also moves us towards a model where we can unify timeouts for wait calls and background coroutines.

Note that this PR introduces mandatory time outs for everything in the libOS, ensuring that we are never wedged in Demikernel. However, it does require us to set these defaults, which might need to be adjusted and should probably eventually be moved to a single place.

### DIFF
--- a/src/rust/catloop/socket.rs
+++ b/src/rust/catloop/socket.rs
@@ -46,18 +46,18 @@ use ::std::{
         SocketAddrV4,
     },
     slice,
+    time::Duration,
 };
 
 //======================================================================================================================
 // Constants
 //======================================================================================================================
 
-/// Maximum number of connection attempts.
-/// This was chosen arbitrarily.
-const MAX_ACK_RECEIVED_ATTEMPTS: usize = 1024;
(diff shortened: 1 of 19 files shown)


ppenna (approved):
LGTM

PR #1140, opened 2024-01-22
[catnap] Allow chained DemiBuffers buffers for push and pop

kyleholohan:
This is a draft to discuss approach.

Add a `next` method to DemiBuffer which will instantiate a new DemiBuffer for the next chained buffer (if any such buffer is present). Allow chained buffers in Windows catnap push/pop operations.

There are some issues with the approach as stated:
- There is no way to chain buffers currently.
- Assume we have some `chain` or `append` method on `DemiBuffer`. This method must consume the buffer being chained to be consistent with other support functions. Specifically, `clone`, which will fork the MetaData for a buffer to avoid having more than one mutable view of a DemiBuffer.
- The mutable/reference semantics are not correct. Currently, a user may call `next` more than once on the same DemiBuffer. Since DemiBuffer is a reference to MetaData, this allows mutliple mutable references.

I'm going to reverse my position: DemiBuffer should wrap the entire buffer chain, rather than an individual chunk. This would make solving reference issues much simpler.

Also, I didn't do any testing yet :)

### DIFF
--- a/src/rust/catnap/win/socket.rs
+++ b/src/rust/catnap/win/socket.rs
@@ -102,6 +102,19 @@ const SOCKADDR_BUF_SIZE: usize = std::mem::size_of::<SOCKADDR_STORAGE>() + 16;
 const ACCEPT_BUFFER_LEN: usize = (SOCKADDR_BUF_SIZE) * 2;
 
 //======================================================================================================================
+// Enums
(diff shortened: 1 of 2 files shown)


anandbonde:
Closing this PR for now. Let's reopen it later if needed.

PR #1143, opened 2024-01-29
[runtime] Enhancement: Add condition variable

iyzhang:
This PR is the first of a sequence designed to remove the Yielder abstraction from our code and replace it with a condition variable. Additionally, when we signal the condition variable, we never simply set the waker bit but always check for whether the task is still valid in the scheduler. In order to do this we have made the scheduler a thread local variable to avoid passing a reference to the scheduler or runtime into every data structure. With this change, we no longer need to track yielders or coroutines to ensure that they are either correctly removed from the scheduler or cancelled when resources are freed. Instead, it is up to each coroutine to monitor the variables that it depends on with a condition variable and exit if those resources are gone. 

This PR makes the following changes:
1. Create a condition variable on which coroutines can wait and then signal or broadcast. This abstraction replaces our Yielders.
2. Make the scheduler into a thread local variable but only visible to the runtime.
3. Make the timer into a thread local variable but only visible to the runtime.
4. Add conditional_yield_with_timeout and conditional_yield_until static functions to the runtime for interacting with the condition variables and ensuring that coroutines do not block forever.


ppenna (approved):
LGTM

ppenna (approved):

PR #1156, opened 2024-02-05
[testing] Enhancement: Adding unified error checking for close

iyzhang:
This PR fixed integration testing for our new changes with the condition variable and state machine. It includes:
1. A unified function for network error code checking
2. Checks for timeouts from the operations themselves rather than the wait (because for Catloop, the operations themselves now time out instead of the wait).


src/rust/catloop/socket.rs
```diff
@@ -163,73 +146,57 @@ impl SharedMemorySocket {
...
+                self.pending_request_ids.remove(&request_id);
+                return Ok((new_socket, new_addr.into()));
+            },
+            // Some error.
+            Err(e) => {
+                // Clean up newly allocated duplex pipe.
+                catmem.close(new_qd)?;
+                return Err(e);
+            },
+        };
```
ppenna, line 187: There is no need for return statements here.

src/rust/catloop/transport.rs
```diff
@@ -102,26 +102,47 @@ impl NetworkTransport for SharedCatloopTransport {
...
-        sd.connect(self.runtime.clone(), self.catmem.clone(), remote).await
+        sd.connect(self.catmem.clone(), remote).await
     }
 
     /// Asynchronous code to close a queue. This function returns a coroutine that runs asynchronously to close a queue
     /// and the underlying Catmem queue and performs any necessary multi-queue operations at the libOS-level after
     /// the close succeeds or fails.
     async fn close(&mut self, sd: &mut Self::SocketDescriptor) -> Result<(), Fail> {
-        sd.close(self.catmem.clone()).await
+        sd.close(self.catmem.clone()).await?;
```
ppenna, line 127: If this fails, we leak the ephemeral port. We should not.

src/rust/catloop/transport.rs
```diff
@@ -102,26 +102,47 @@ impl NetworkTransport for SharedCatloopTransport {
...
+                // Allocate ephemeral port from the pool.
+                self.runtime.free_ephemeral_port(local.port())?;
+            }
+        }
+        Ok(())
     }
 
     fn hard_close(&mut self, sd: &mut Self::SocketDescriptor) -> Result<(), Fail> {
-        sd.hard_close(&mut self.catmem)
+        sd.hard_close(&mut self.catmem)?;
```
ppenna, line 138: If this fails, we leak the ephemeral port. We should not.

ppenna (approved):

src/rust/catloop/transport.rs
iyzhang: Actually we already free the emphemeral port in the close path, so I will be removing this. https://github.com/microsoft/demikernel/blob/835d33ab599ecbe107fd312aa033cdbe6e2eb41b/src/rust/demikernel/libos/network/libos.rs#L323

src/rust/catloop/socket.rs
iyzhang: yup, removed

PR #1157, opened 2024-02-05
[catnap] Enhancement: allow configuring Nagle's algorithm for TCP on Windows

kyleholohan:
This PR allows configuring Nagle's algorithm (i.e. TCP_NODELAY socket option) via a "use_nagle" config in config.yaml. I also refactored some of the Catnap/win/config.rs because the `require_option...` contradiction was bothering me.


iyzhang (approved):
Looks good to me. I might suggest leaving Nagel off as the default, this is what we do in Linux.

PR #1158, opened 2024-02-07
[catloop] Enhancement: Remove all calls to wait

iyzhang:
This PR removes all calls to wait and uses of the demi_sgarray_t from the Catloop libOS. Instead it moves it completely to using coroutines and directly calling asynchronous Catmem functions. This is important because it keeps our scheduler from being re-entrant, which could potentially be difficult to debug later.


ppenna (approved):
LGTM

PR #1163, opened 2024-02-21
[catnap] Add support for condition variables in Windows transport

iyzhang:
This PR adds the necessary functions for moving the Windows catnap transport to our new condition variables.


src/rust/collections/pin_slab.rs
```diff
@@ -107,22 +114,23 @@ impl<T> PinSlab<T> {
...
-        self.insert_at(key, val)?;
-        Some(key)
+    #[allow(unused)]
+    /// Gets a read only pinned reference.
+    pub fn get_pin(&self, key: usize) -> Option<Pin<&T>> {
+        let entry: &T = self.internal_get(key)?;
+        // Safety: all storage is pre-allocated in chunks, and each chunk
+        // doesn't move. We only provide mutators to drop the storage through
+        // `remove` (but it doesn't return it).
+        unsafe { Some(Pin::new_unchecked(entry)) }
```
ppenna, line 124: If unused, we can get rid of this.

iyzhang: We may want to use this in some other code, I left it for symmetry since we have get_pin_mut and get (and I didn't want to re-write this function later if we need it).

ppenna (approved):
LGTM

src/rust/catnap/win/overlapped.rs
```diff
@@ -64,37 +56,31 @@ pub struct OverlappedResult {
 
 /// Data required by the I/O completion port processor to process I/O completions.
 #[repr(C)]
-struct OverlappedCompletion {
+struct OverlappedCompletion<S: Unpin> {
```
kyleholohan, line 59: The semantics of Unpin are still a little unclear to me. From the Rust docs:
> Implementing the Unpin trait for T expresses the fact that T is pinning-agnostic: it shall not expose nor rely on any pinning guarantees.

The type `SocketOpState` itself doesn't care about pinning, e.g. it's not self-referential; however, it can't be used for its intended purpose without pinning, because e.g. the address memory pointer on AcceptEx must be pinned during the call. Implementing `Unpin` in this case seems to just make the pinning use more implicit. This has the benefit of making the code easier to write, with fewer unsafe blocks, but the written code is less expressive and therefore more error-prone to future development. Not sure this merits a rewrite of what's here, just trying to build my own understanding.

kyleholohan: I guess in my first attempt, the lack of Unpin was explicit, because every time I referenced the State, I wanted to stop and think whether it was valid, then write a justification for why such an access was valid at that point. But maybe that's an overuse of unsafe?

kyleholohan: Not going to block on this.

kyleholohan (approved):
Left some thoughts on Unpin, but otherwise looks good.

PR #1164, opened 2024-02-29
[examples] Improving Statistics for TCP Echo

ppenna:
## Description

This PR adds P50 and P99 statistics for TCP Echo.


iyzhang (approved):
LGTM

PR #1165, opened 2024-02-29
[build] Enhancement: Bump socket2 version

iyzhang:
The windows build seems to be taking the newest version of socket2 available, so we need to bump the windows-sys crate to match.


ppenna (approved):
LGTM

PR #1168, opened 2024-03-06
[testing] Add multicore support to tcp echo

iyzhang:
This PR add support for multicore execution to our tcp-echo benchmark as an example of how we will handle basic multi-core architectures with RSS support. It currently only works with Catnap because there is more configuration that needs to be done for DPDK support. Since it has a share nothing architecture, it should work fine with Catloop and Catmem but is not currently supported.


iyzhang:
Closing this as it was against a feature branch that no longer exists

PR #1172, opened 2024-03-07
[catpowder] Bug Fix: Doubling time out for tcp push pop

iyzhang:
I believe that Catpowder is timing out on the accept for the release build on the CI. I was not able to reproduce on my machine.


iyzhang:
This was caused by the memory leak, not a time out issue. See #1173

PR #1174, opened 2024-03-08
[libos] Bug Fix: Unify DemiBuffer allocation

### LINKED ISSUE #1173: [libos] Memory Leak on DemiBuffers
## Description

After our recent architectural change, we have been allocating a DemiBuffer on each pop and not freeing it.

## How to Reproduce

1. Run any test for longer than 30 seconds under load and it will be clear that the heap is growing continuously

iyzhang:
This PR closes #1173 by unifying the hierarchy of memory allocation for DemiBuffers. In the past, we were allocating Demibuffers out of the libOS and then freeing them through the shared runtime, which caused a memory leak because those were two different instances of the MemoryRuntime. Now, all DemiBuffer allocations go through the libOS first, then the transport and then the Network(Memory)Runtime. Since not all libOSes have NetworkRuntimes, some of the simply use the default heap allocator in the transport. However, we need to be careful in the future when supporting multiple libOSes with multiple transports to free DemiBuffers back to the right pool. In fact, we should probably have more tags to enable this.

### DIFF
--- a/src/rust/catmem/mod.rs
+++ b/src/rust/catmem/mod.rs
@@ -11,6 +11,7 @@ mod ring;
 use self::queue::SharedCatmemQueue;
 use crate::{
     demikernel::config::Config,
+    pal::linux::socketaddrv4_to_sockaddr,
     runtime::{
         fail::Fail,
         limits,
@@ -19,7 +20,12 @@ use crate::{
             MemoryRuntime,
         },
         queue::downcast_queue,
-        types::demi_sgarray_t,
+        types::{
+            demi_opcode_t,
+            demi_qr_value_t,
+            demi_qresult_t,
+            demi_sgarray_t,
+        },
         Operation,
         OperationResult,
         SharedDemiRuntime,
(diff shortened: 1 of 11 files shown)


src/rust/catmem/mod.rs
```diff
@@ -204,6 +212,73 @@ impl SharedCatmemLibOS {
...
+                demi_qresult_t {
+                    qr_opcode: demi_opcode_t::DEMI_OPC_FAILED,
+                    qr_qd: qd.into(),
+                    qr_qt: qt.into(),
+                    qr_ret: e.errno as i64,
+                    qr_value: unsafe { mem::zeroed() },
+                }
+            },
+        }
+    }
```
ppenna, line 280: Is it possible to share this function across all liboses?

ppenna (approved):

ppenna (approved):
LGTM

ppenna (approved):
LGTM

ppenna (approved):
LGTM

src/rust/catmem/mod.rs
iyzhang: We share it across the network libOSes but since the memory libOS has some unreachable statements, I thought I would leave the two functions separate.

PR #1175, opened 2024-03-08
[libos] Remove copy from transport API

iyzhang:
When we added the transport trait we introduced a copy into almost all of the libOSes because the pop API took a DemiBuffer in and filled it. This removes those copies and lets the transport allocate the DemiBuffer and directly return it.


ppenna (approved):
LGTM

PR #1177, opened 2024-03-11
[runtime] Bug Fix: Add drop function for condition variable

iyzhang:
This PR keeps condition variables from growing in an uncontrolled way by removing waiters when the reference to the waiter is removed.


ppenna:
This change worsens the performance of Demikernel in our benchmark.

[tcp-echo-latency-p50.pdf](https://github.com/microsoft/demikernel/files/14605954/tcp-echo-latency-p50.pdf)
[tcp-echo-latency-p99.pdf](https://github.com/microsoft/demikernel/files/14605955/tcp-echo-latency-p99.pdf)
[tcp-echo-throughput.pdf](https://github.com/microsoft/demikernel/files/14605956/tcp-echo-throughput.pdf)

PR #1183, opened 2024-03-12
[runtime] Enhancement: Create into_buf in the MemoryRuntime

carvalhof:
This PR creates `into_buf` function to convert `demi_sgarray_t` to `DemiBuffer` without clone.

### DIFF
--- a/src/rust/runtime/memory/mod.rs
+++ b/src/rust/runtime/memory/mod.rs
@@ -164,3 +164,56 @@ pub trait MemoryRuntime {
         Ok(clone)
     }
 }
+
+//======================================================================================================================
+// Unit Tests
+//======================================================================================================================
+
+#[cfg(test)]
+mod tests {
+    use crate::{
+        expect_ok,
+        demi_sgarray_t,
+        runtime::memory::MemoryRuntime,
+    };
+    use ::test::{
+        black_box,
+        Bencher,
+    };
+
+    // The buffer size.
+    const BUFSIZE: usize = 1024;
+
+    pub struct DummyRuntime { }
+
+    impl MemoryRuntime for DummyRuntime {}
+
+    #[bench]
+    fn benchmark_clone_sgarray(b: &mut Bencher) {
+        let runtime: DummyRuntime = DummyRuntime { };
+
+        let sga: demi_sgarray_t = match runtime.sgaalloc(BUFSIZE) {
+            Ok(sga) => sga,
+            Err(e) => panic!("failed to allocate sgarray: {:?}", e),
+        };
+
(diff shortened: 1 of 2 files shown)


src/rust/runtime/memory/mod.rs
```diff
@@ -85,45 +107,18 @@ pub trait MemoryRuntime {
...
     fn sgafree(&self, sga: demi_sgarray_t) -> Result<(), Fail> {
-        // Check arguments.
-        // TODO: Drop this check once we support scatter-gather arrays with multiple segments.
-        if sga.sga_numsegs != 1 {
-            return Err(Fail::new(libc::EINVAL, "demi_sgarray_t has invalid segment count"));
-        }
-
-        if sga.sga_buf == ptr::null_mut() {
-            return Err(Fail::new(libc::EINVAL, "demi_sgarray_t has invalid DemiBuffer token"));
+        match self.into_buf(&sga) {
```
iyzhang, line 110: Please do not remove TODO comments. For example, we still do not support multiple sga segments (I think), so we probably shouldn't remove it.

iyzhang (changes requested):

src/rust/runtime/memory/mod.rs
carvalhof: Actually, the TODOs are moved into `into_buf` function.

ppenna (changes requested):

src/rust/runtime/memory/mod.rs
iyzhang: Ah ok, all good then!

iyzhang (approved):

src/rust/runtime/memory/mod.rs
```diff
@@ -52,6 +52,28 @@ pub trait MemoryRuntime {
         })
     }
 
+    /// Converts a scatter-gather array into a buffer.
+    fn into_buf(&self, sga: &demi_sgarray_t) -> Result<DemiBuffer, Fail> {
+        // Check arguments.
+        // TODO: Drop this check once we support scatter-gather arrays with multiple segments.
+        if sga.sga_numsegs != 1 {
+            return Err(Fail::new(libc::EINVAL, "demi_sgarray_t has invalid segment count"));
```
anandbonde, line 60: We could enhance this message to state that multiple segments are not currently supported.

anandbonde (approved):
Looks good.

iyzhang:
Unclear that this is needed for performance right now. Will re-evaluate after we possibly do some restructuring of DemiBuffer

PR #1185, opened 2024-03-12
[perf] Remove Task Name

ppenna:
## Description

This PR removes tasks names from the Task structure to avoid memory allocations in the critical path.

### DIFF
--- a/nettest/input/pop-blocking.pkt
+++ b/nettest/input/pop-blocking.pkt
@@ -22,7 +22,7 @@
 // Receive data packet.
 +.1 < P. seq 1(1000) ack 1 win 65535 <nop>
 // Send ACK packet.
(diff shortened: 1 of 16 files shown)


iyzhang (approved):
Looks fine to me. We put these in in the first place so we could debug more easily. Is there some way to retain that functionality without the overhead?

ppenna:
> Looks fine to me. We put these in in the first place so we could debug more easily. Is there some way to retain that functionality without the overhead?

Sure I fixed this. Task names will be compiled for debug builds.

iyzhang:
Awesome, thanks for doing this.

iyzhang (approved):

ppenna:
Closing this as requested by @iyzhang.

iyzhang:
This was replaced in PR #1191

PR #1186, opened 2024-03-12
[Runtime] Enhancement: support pooled allocation of DemiBuffer

kyleholohan:
Adds a few new classes to support pools of memory for DemiBuffer: `MemoryPool` which is a generic homogeneous pool of memory, and `BufferPool` which integrates `MemoryPool` and `DemiBuffer`. The latter is pretty light-weight: it's mainly to prevent exposing the details of `MetaData` to the user. Still needs some more testing, and a better `DemiBuffer::clone` implementation.

### DIFF
--- a/src/rust/runtime/memory/demibuffer.rs
+++ b/src/rust/runtime/memory/demibuffer.rs
@@ -76,6 +76,7 @@ use ::std::{
     },
     slice,
 };
+use std::rc::Rc;
 
 // Buffer Metadata.
 // This is defined to match a DPDK MBuf (rte_mbuf) in order to potentially use the same code for some DemiBuffer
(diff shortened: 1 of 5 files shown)


ppenna (changes requested):

src/rust/runtime/memory/demibuffer.rs
```diff
@@ -262,37 +345,72 @@ impl DemiBuffer {
...
+
+    /// Create a new DemiBuffer in the specified memory, with relevant configuration values.
+    fn new_from_parts(
+        metadata_buf: &mut MaybeUninit<MetaData>,
+        buf_addr: *mut MaybeUninit<u8>,
+        capacity: u16,
+        pool: Option<Rc<MemoryPool>>,
+    ) -> Self {
+        let buf_addr: *mut u8 = if capacity > 0 {
+            // TODO: casting the MaybeUninit away can cause UB (when deref'd). Change the exposed data type from
```
ppenna, line 390: Should we create an issue to keep track of this?

src/rust/runtime/memory/demibuffer.rs
```diff
@@ -77,12 +88,12 @@ use ::std::{
 #[repr(C)]
 #[repr(align(64))]
```
ppenna, line 89: Since we are not matching DPDK's structure we may want to remove this and let the compiler handle alignments.

PR #1187, opened 2024-03-13
[runtime] Bug Fix: Clean up spurious wake ups in the timer

iyzhang:
This PR removes entries from the ordered heap in the global timer when a coroutine is no longer waiting on it. It also reduces the scope of the global variable that holds the timer since another PR will soon remove the global reference to the scheduler.

### DIFF
--- a/nettest/input/pop-blocking.pkt
+++ b/nettest/input/pop-blocking.pkt
@@ -22,7 +22,7 @@
 // Receive data packet.
 +.1 < P. seq 1(1000) ack 1 win 65535 <nop>
 // Send ACK packet.
-+.6 > . seq 1(0) ack 1001 win 65535 <nop>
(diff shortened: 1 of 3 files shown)


ppenna (commented):
LGTM

ppenna:
This change introduces performance degradation on our benchmark.

[tcp-echo-latency-p50.pdf](https://github.com/microsoft/demikernel/files/14595339/tcp-echo-latency-p50.pdf)
[tcp-echo-latency-p99.pdf](https://github.com/microsoft/demikernel/files/14595340/tcp-echo-latency-p99.pdf)
[tcp-echo-throughput.pdf](https://github.com/microsoft/demikernel/files/14595341/tcp-echo-throughput.pdf)

ppenna (approved):

PR #1191, opened 2024-03-18
[perf] Feature: Adding support for async profiling

iyzhang:
This PR adds a feature to our profiler to measure the time spent in async blocks. It only supports fused (non-reentrant) blocks for now. It works by simply wrapping a Future around the existing code that measures when we enter the async block and stops the clock when we exit with either a Pending or Ready status. 

For now, I have annotated all of our coroutines but nothing more detailed. We should add more annotations, especially around the different async workers that all run in a single background coroutine.

### DIFF
--- a/src/rust/perftools/profiler/mod.rs
+++ b/src/rust/perftools/profiler/mod.rs
@@ -1,13 +1,24 @@
 // Copyright(c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+//! This module provides a small performance profiler for the Demikernel libOSes.
+
 #[cfg(test)]
 mod tests;
 
+use ::futures::future::FusedFuture;
 use std::{
     cell::RefCell,
+    fmt,
+    fmt::Debug,
+    future::Future,
     io,
+    pin::Pin,
     rc::Rc,
+    task::{
+        Context,
+        Poll,
+    },
     time::{
         Duration,
         SystemTime,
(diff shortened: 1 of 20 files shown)


src/rust/perftools/profiler/mod.rs
```diff
@@ -63,11 +41,13 @@ macro_rules! timer {
 /// root nodes. Thus, if you have multiple root nodes and they do not cover
 /// all code that runs in your program, the printed frequencies will be
 /// overestimated.
+
```
ppenna, line 44: Please rollback this change.

src/rust/perftools/profiler/mod.rs
```diff
@@ -63,11 +41,13 @@ macro_rules! timer {
...
 pub fn write<W: io::Write>(out: &mut W, max_depth: Option<usize>) -> io::Result<()> {
     PROFILER.with(|p| p.borrow().write(out, max_depth))
 }
 
 /// Reset profiling information.
+
```
ppenna, line 50: Please rollback this change.

src/rust/perftools/profiler/mod.rs
```diff
@@ -203,6 +229,7 @@ impl Drop for Guard {
 /// Note that there is a global thread-local instance of `Profiler` in
 /// [`PROFILER`](constant.PROFILER.html), so it is not possible to manually
 /// create an instance of `Profiler`.
+
```
ppenna, line 232: Please rollback this change.

ppenna (approved):
LGTM

iyzhang:
Note the statistics at the end of this file:
[tcp-echo-sequential-1-64-128-client-demikernel-vm5.eastus.cloudapp.azure.com.stdout.txt](https://github.com/microsoft/demikernel/files/14641466/tcp-echo-sequential-1-64-128-client-demikernel-vm5.eastus.cloudapp.azure.com.stdout.txt)

PR #1193, opened 2024-03-18
[perf] Bug Fix: Remove extraneous tracing

iyzhang:
Printing at the info level is too much for every time we enter and leave a scope. It is unclear if we even want trace level print outs since we enter and exit the polling coroutines frequently.


ppenna (approved):
I understand that this is a change that was not included in the original PR https://github.com/microsoft/demikernel/pull/1191

PR #1194, opened 2024-03-18
[testing] Add multi-core support to TCP echo

iyzhang:
This PR add support for multicore execution to our tcp-echo benchmark as an example of how we will handle basic multi-core architectures with RSS support. It currently only works with Catnap because there is more configuration that needs to be done for DPDK support. Since it has a share nothing architecture, it should work fine with Catloop and Catmem but is not currently supported.

### DIFF
--- a/examples/tcp-echo/main.rs
+++ b/examples/tcp-echo/main.rs
@@ -284,7 +284,13 @@ fn main() -> Result<()> {
     let mut threads = vec![];
     match args.peer_type.as_str() {
         "server" => {
-            for _ in 0..args.nthreads.unwrap_or(1) {
+            // Multi-threaded server is only supported by Catnap.
+            // TODO: Check/add support for other libOSes.
+            let nthreads: usize = match libos_name {
+                LibOSName::Catnap => args.nthreads.unwrap_or(1),
+                _ => 1,
+            };
+            for _ in 0..nthreads {
                 if let Ok(handle) = start_server_thread(libos_name, args.addr, args.log_interval) {
                     threads.push(handle)
                 }
--- a/examples/tcp-echo/main.rs
+++ b/examples/tcp-echo/main.rs
@@ -25,6 +25,8 @@ use server::TcpEchoServer;
 use std::{
     net::SocketAddr,
     str::FromStr,
+    thread,
+    thread::JoinHandle,
     time::Duration,
 };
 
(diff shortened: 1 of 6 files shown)


src/rust/catnap/linux/transport.rs
```diff
@@ -294,6 +294,7 @@ impl NetworkTransport for SharedCatnapTransport {
                     error!("new(): {}", cause);
                     return Err(Fail::new(get_libc_err(e), &cause));
                 }
+
```
ppenna, line 297: Could you evert this line?

examples/tcp-echo/main.rs
```diff
@@ -110,7 +114,15 @@ impl ProgramArguments {
...
-                    .help("Sets number of clients"),
+                    .help("Sets number of clients (per thread)"),
+            )
+            .arg(
+                Arg::new("nthreads")
+                    .long("nthreads")
+                    .value_parser(clap::value_parser!(usize))
+                    .required(false)
+                    .value_name("NUMBER")
+                    .help("Sets number of threads"),
```
ppenna, line 125: This option should be available only on targets that support this.

src/rust/demikernel/libos/name.rs
```diff
@@ -12,6 +12,7 @@ use ::std::env;
 // Structures
 //======================================================================================================================
 
+#[derive(Clone, Copy)]
```
ppenna, line 15: Why is this change required?

tools/ci/config/ci_map.yaml
```diff
@@ -121,6 +121,18 @@ catnap:
...
+      nclients: 32
+      nrequests: 1024
+      run_mode: concurrent
+      nthreads: 2
+    scenario9:
+      bufsize: 64
+      nclients: 32
+      nrequests: 1024
+      run_mode: concurrent
+      nthreads: 4
```
ppenna, line 135: This requires the CI tool to be updated.

tools/ci/config/ci_map.yaml
ppenna, line 1: Something else is required to make this change work out of the box for the other LibOSes.

iyzhang: Hmm, I thought I turned it off for all other libOSes...

PR #1205, opened 2024-03-25
[perf] Add support for async blocks

iyzhang:
This PR adds support in our profiler for arbitrary async blocks, not just coroutines.


ppenna (approved):

ppenna (approved):
LGTM

PR #1212, opened 2024-03-28
[runtime] Enhancement: Add new 'wait_next_n' binding and add type annotations to bindings

kyleholohan:
This PR adds a new binding API to batch-deque completed operations from the libOS. The semantics are similar to `wait_any` except that the returned QTokens are not filtered, and the method may retrieve multiple results in a single call.

Also added GCC- and MSVC-specific correctness annotations to the C headers.

### DIFF
--- a/src/rust/demikernel/bindings.rs
+++ b/src/rust/demikernel/bindings.rs
@@ -42,6 +42,7 @@ use ::libc::{
     sockaddr,
 };
 use ::socket2::SockAddr;
+use std::mem::MaybeUninit;
 use ::std::{
     cell::RefCell,
     ffi::CStr,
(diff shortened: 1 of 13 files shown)


src/rust/demikernel/bindings.rs
```diff
@@ -42,6 +42,7 @@ use ::libc::{
     sockaddr,
 };
 use ::socket2::SockAddr;
+use std::mem::MaybeUninit;
```
ppenna, line 45: Please, pre-append this with `::`.

ppenna (commented):
@kyleholohan Rust implementation looks good to me. Let me know when bindings are ready to be reviewed.

kyleholohan:
Closing to open on a real branch.

PR #1214, opened 2024-03-29
[runtime] Avoid UB for Zero Length Buffer

ppenna:
## Description

This PR fixes undefined behavior when a dereferencing zero length `demibuffer`.


kyleholohan (approved):
lgtm

PR #1220, opened 2024-04-03
[demikernel] Bug Fix: Don't leak libos when calling init more than once

stolet:
Multiple calls to demi_init would create a new libOS and the old one would leak. This change makes it so that a call to demi_init immediately returns if there is a libOS already initialized.

### DIFF
--- a/src/rust/demikernel/bindings.rs
+++ b/src/rust/demikernel/bindings.rs
@@ -85,6 +85,15 @@ pub extern "C" fn demi_init(argc: c_int, argv: *mut *mut c_char) -> c_int {
         Err(e) => panic!("{:?}", e),
     };
 
+    // Check if demikernel has already been initialized and return
+    match unsafe { DEMIKERNEL.try_borrow() } {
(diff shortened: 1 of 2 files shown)


src/rust/demikernel/bindings.rs
```diff
@@ -82,6 +82,15 @@ pub extern "C" fn demi_init(argc: c_int, argv: *mut *mut c_char) -> c_int {
         Err(e) => panic!("{:?}", e),
     };
 
+    // Check if demikernel has already been initialized and return
+    match unsafe { DEMIKERNEL.try_borrow() } {
+        Ok(libos) => match *libos {
+            Some(_) => return 0,
```
ppenna, line 88: Instead of continuing silently, how about returning `-EEXIST`?

stolet: Sounds good. Updated.

stolet:
@microsoft-github-policy-service agree

________________________________

ppenna (approved):
LGTM

PR #1227, opened 2024-04-08
[runtime] Enhancement: Increase timer resolution

iyzhang:
Currently our timer resolution is 64 scheduler tasks. Assuming each task runs for roughly 100ns, we are checking the clock every 6.4us, which is too frequently to make a sys call. Thus, this PR increases the timer resolution to 1024, which is roughly once every 100us. This should reduce the sys call overhead, while still providing a reasonable timer granularity.


ppenna (approved):
LGTM

PR #1229, opened 2024-04-09
[collections] Overflow on Buffer Copy

ppenna:
## Description

This PR exposes and fixes an overflow bug that was happening when popping data from a concurrent ring buffer.

### DIFF
--- a/src/rust/collections/concurrent_ring.rs
+++ b/src/rust/collections/concurrent_ring.rs
@@ -217,7 +217,7 @@ impl ConcurrentRingBuffer {
         let first_len: usize = if pop_offset + pop_len + HEADER_SIZE > self.capacity() {
             self.capacity() - first_offset
         } else {
-            buf_len
+            pop_len
         };
         let buf_ptr: *mut u8 = buf.as_mut_ptr();
         let ring_ptr: *const u8 = unsafe { self.buffer.get().as_ptr() };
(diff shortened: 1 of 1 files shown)


src/rust/collections/concurrent_ring.rs
```diff
@@ -217,27 +217,29 @@ impl ConcurrentRingBuffer {
...
         let buf_ptr: *mut u8 = buf.as_mut_ptr();
         let ring_ptr: *const u8 = unsafe { self.buffer.get().as_ptr() };
         // Copy the data into the ring buffer.
+        debug_assert!(first_len <= pop_len);
         unsafe {
             copy(ring_ptr.add(first_offset), buf_ptr, first_len);
         }
         // If there is remaining data in the buffer, wrap around.
-        if len > first_len {
+        if buf_len > first_len {
```
iyzhang, line 230: I believe this should be pop_len > first_len because the buffer might be bigger than the amount of data available.

src/rust/collections/concurrent_ring.rs
```diff
@@ -217,27 +217,29 @@ impl ConcurrentRingBuffer {
...
             // Copy the data into the ring buffer.
+            debug_assert!((first_len - first_len) <= pop_len);
```
iyzhang, line 232: I'm not exactly sure what you are testing here but right now it's always 0.

src/rust/collections/concurrent_ring.rs
ppenna: You are correct. I missed this additional bug when producing the PR.

src/rust/collections/concurrent_ring.rs
```diff
@@ -217,27 +217,28 @@ impl ConcurrentRingBuffer {
...
+                copy(ring_ptr, buf_ptr.add(first_len), pop_len - first_len);
             }
         }
 
         // Move to next buffer.
         self.release_space(pop_offset, pop_len);
         trace!(
             "try_push() len={:?} push_offset={:?} pop_offset={:?}",
-            len,
+            buf_len,
```
iyzhang, line 241: I think we might want this to be pop_len as well. We may want to truncate the buffer to the amount of data popped, although we do return the size explicitly.

iyzhang (approved):

PR #1230, opened 2024-04-10
[testing] Bug Fix: Add proper error checking

iyzhang:
Move all of our error testing on integration tests to check_for_network_error. We introduced the function some time ago but there were still some lingering integration tests that did not use it.


ppenna (approved):
LGTM

PR #1232, opened 2024-04-10
[collections] Check for Invalid Push Size of Concurrent Ring Buffer

ppenna:
## Description

This PR fixes invalid push size for concurrent ringle buffers. It also adds more test units to this data structure.

### DIFF
--- a/src/rust/collections/concurrent_ring.rs
+++ b/src/rust/collections/concurrent_ring.rs
@@ -309,10 +309,7 @@ impl ConcurrentRingBuffer {
     #[allow(unused)]
     pub fn is_full(&self) -> bool {
         timer!("collections::concurrent_ring::is_full");
-        let push_offset = peek(self.push_offset);
-        let pop_offset = peek(self.pop_offset);
-
-        self.available_space(push_offset, pop_offset) == 0
+        self.remaining_capacity() == HEADER_SIZE
     }
 
     /// Peeks the target ring buffer and checks if it is empty.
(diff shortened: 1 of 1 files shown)


src/rust/collections/concurrent_ring.rs
```diff
@@ -506,6 +500,58 @@ mod test {
         Ok(ring)
     }
 
+    /// Pushes data to a ring buffer until it is full.
+    fn make_full_ring_buffer() -> Result<ConcurrentRingBuffer> {
```
iyzhang, line 504: could call this `fill_ring_buffer`

src/rust/collections/concurrent_ring.rs
```diff
@@ -506,6 +500,58 @@ mod test {
...
+        let mut npushes: usize = 0;
+        let byte: [u8; 1] = [0xff];
+        let ring: ConcurrentRingBuffer = do_new()?;
+        while ring.remaining_capacity() > HEADER_SIZE {
+            do_sucess_push_bytes(&ring, &byte)?;
```
iyzhang, line 509: Success has 2 Cs

iyzhang (approved):

src/rust/collections/concurrent_ring.rs
ppenna: Sure.

src/rust/collections/concurrent_ring.rs
ppenna: Thanks.

PR #1236, opened 2024-04-12
[pdpix] Demikernel in Thread Local Storage

ppenna:
## Description

This PR moves Demikernel to thread local storage.

### DIFF
--- a/src/rust/demikernel/bindings.rs
+++ b/src/rust/demikernel/bindings.rs
@@ -47,7 +47,7 @@ use ::std::{
     ffi::CStr,
     mem::{
         self,
-        MaybeUninit
+        MaybeUninit,
     },
     net::SocketAddr,
     ptr,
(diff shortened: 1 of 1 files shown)


src/rust/demikernel/bindings.rs
```diff
@@ -86,25 +88,28 @@ pub extern "C" fn demi_init(argc: c_int, argv: *mut *mut c_char) -> c_int {
...
 
     // Check if demikernel has already been initialized and return
-    match unsafe { DEMIKERNEL.try_borrow() } {
-        Ok(libos) => match *libos {
-            Some(_) => return libc::EEXIST,
-            None => (),
-        },
-        Err(e) => panic!("{:?}", e),
+    let ret: i32 = DEMIKERNEL.with(|demikernel| match *demikernel.borrow() {
+        Some(_) => -libc::EEXIST,
```
kyleholohan, line 92: Why are you negating the return code?

kyleholohan: Just double-checked, and libos.h says:
> On failure, a positive error code is returned instead.

kyleholohan (approved):
Approved, pending comment.

src/rust/demikernel/bindings.rs
ppenna: Thanks.

PR #1237, opened 2024-04-13
[catnap] Bug Fix: Check if socket is already nonblocking

stolet:
Currently, if the socket is set to non-blocking when created, set_nonblocking will return an error. Therefore, check if the socket is nonblocking before calling set_nonblocking.


ppenna (approved):
LGTM

PR #1241, opened 2024-04-25
[build] Copy All Libs On Install

ppenna:
## Description

This PR closes https://github.com/microsoft/demikernel/issues/1240


ppenna:
Merging this PR as it concerns the building system.

PR #1244, opened 2024-04-25
[inetstack] Bug Fix: Clamp window scale to 14

### LINKED ISSUE #216: [inetstack] Fail Instead of Panicking on Bad Remote Window Scale
Description
========

In `protocols::tcp::active_open::receive()`, we currently panicking on a bad window scale, instead of failing or rounding down the provided value.

iyzhang:
This closes #216

### DIFF
--- a/src/rust/inetstack/protocols/tcp/active_open.rs
+++ b/src/rust/inetstack/protocols/tcp/active_open.rs
@@ -180,13 +180,24 @@ impl<N: NetworkRuntime> SharedActiveOpenSocket<N> {
             }
         }
 
-        let (local_window_scale, remote_window_scale) = match remote_window_scale {
-            Some(w) => (self.tcp_config.get_window_scale() as u32, w),
+        let (local_window_scale, remote_window_scale): (u32, u8) = match remote_window_scale {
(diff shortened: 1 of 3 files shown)


src/rust/inetstack/protocols/tcp/active_open.rs
```diff
@@ -180,14 +180,25 @@ impl<N: NetworkRuntime> SharedActiveOpenSocket<N> {
...
+                    self.tcp_config.get_window_scale() as u32
+                };
+                let remote: u8 = if remote_window_scale > 14 {
+                    warn!("remote windows scale larger than 14 is incorrect, so setting to 14. See RFC 1323.");
+                    14
+                } else {
+                    remote_window_scale
+                };
+                (local, remote)
+            },
```
ppenna, line 198: Can we have a file-scope constant for `14`? Preferrable this should be placed on the same file with other TCP constants.

iyzhang: yeah sounds good

ppenna (approved):
LGTM

PR #1245, opened 2024-04-25
[crate] Enhancement: Remove use of super

### LINKED ISSUE #168: [build] Fix the super vs. self vs. crate inconsistency
Context
======

The codebase currently uses a mix of super vs. self vs. crate methods to specify references. It would be good to have consistency.

Proposed Solution
=============

Finalize and use the same method across the codebase.

Alternative solutions
=============
This is not a bug, but it would make the code easier to read/consistent if we just use one method everywhere. This is possibly applicable to other repos in Demikernel, as well.

iyzhang:
This closes #168


ppenna (approved):
LGTM

PR #1246, opened 2024-04-25
[testing] Turn timer tests back on

iyzhang:
Turn the timer unit tests back on now that we are not using a global scheduler any more


ppenna (approved):
LGTM

PR #1247, opened 2024-04-25
[testing] Remove references to push2

### LINKED ISSUE #196: [pdpix] Move System Calls that Operate on Raw Data to Demikernel
Description
========

Once we have a common buffer representation across all the libOSes, we should move al system calls that operate on raw data to the Demikernel repo. Currently, these system calls are:

- `pub fn push2(qd: QDesc, data: &[u8]) -> Result<QToken, Fail>`
- `pub fn pushto2(qd: QDesc, data: &[u8], remote: Ipv4Endpoint, ) -> Result<QToken, Fail>`

iyzhang:
We previously removed all raw data calls to Demikernel but this PR removes any references to old push2 interface. Closes #196

### DIFF
--- a/tests/rust/tcp.rs
+++ b/tests/rust/tcp.rs
@@ -947,7 +947,7 @@ mod test {
                 Ok(_) => {
                     // Close socket if not error because this test cannot continue.
                     // FIXME: https://github.com/demikernel/demikernel/issues/633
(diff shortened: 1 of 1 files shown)


ppenna (approved):
LGTM

PR #1249, opened 2024-04-26
[catpowder] Close Raw Socket

### LINKED ISSUE #671: [catpowder] Implement drop for libOS for correct cleanup
## Description

Currently our libOSes do not have any code for cleaning up resources on exit, specifically for components that deal with external resources. For example, Catpowder should ensure that the OS raw socket is closed when the socket data structure is dropped and when Catpowder exits, it should drop everything from the IoQueueTable to ensure that all lingering raw sockets are closed.

## How to Reproduce

1. Run Catpowder
2. Crash on unit test
3. Cannot run again because of lingering open sockets

## Related Issues

- #649 
- #651

iyzhang:
Currently, we do not close the raw socket on exit, which possibly causes a lingering socket in Linux when we crash. Closes #671

### DIFF
--- a/src/rust/catpowder/runtime/mod.rs
+++ b/src/rust/catpowder/runtime/mod.rs
@@ -26,6 +26,7 @@ use crate::{
             types::MacAddress,
         },
         Runtime,
+        SharedObject,
     },
 };
 use ::std::{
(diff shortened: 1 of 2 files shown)


ppenna (approved):
LGTM

PR #1251, opened 2024-04-27
[inetstack] Bug Fix: Check bind to local IP

iyzhang:
This PR checks that IP addresses to bind match the local assigned address (unless it is the wildcard address).


ppenna (commented):
Thank you for introducing this change!

The proposed fix is effective, but it specifically applies to LibOSes that are backed by our network stack.

Currently, all semantic checks for network LibOSes occur within the `SharedNetworkLibOS` structure implementation. Thanks to this construct, all network LibOSes exhibit the same behavior. To maintain consistency, we should move the bind check to the same layer. There are multiple ways to address this issue, and we can have a dedicated discussion about them later.

In the meantime, I’ve made a small improvement by unifying the logic at the `inetstack` layer. This ensures consistent behavior across all network protocols we eventually implement.

If you accept my change, I ask you to do the following:
- Open an issue to keep track of this improvement.
- Change my `TODO` message to a `FIXME` and paste the link of the issue there
- Squash all commits in this PR into one.

ppenna (commented):
Thanks for working on that.

ppenna (approved):
LGTM

PR #1255, opened 2024-04-30
[test] Recursive Test Case Discovery

ppenna:
## Description

This PR introduces several small enhancements to the network simulator:

- Support single test case run
- Recursive find test cases in a directory
- Drop use of absolute time in existing test cases
- Clean up network simulator output


iyzhang (approved):
Thanks for doing this, it will make testing much easier

PR #1258, opened 2024-05-01
[inetstack] Bug Fix: Support timed wait

### LINKED ISSUE #234: [inetstack] MSL for Active Sockets
# Description

We should set to wait for `2*MSL` when closing active sockets.

# Related Issues

#224

### LINKED ISSUE #1083: [inetstack] Support SO_LINGER
## Context

Currently our TCP stack does not properly perform a time wait. We should have an option for setting it once we do perform it.

## Proposed Solution

Add a way for applications to set socket options in our API.

iyzhang:
This PR closes #234 and closes #1083


src/rust/runtime/network/socket/option.rs
```diff
@@ -38,6 +38,9 @@ impl TcpSocketOptions {
 
 impl Default for TcpSocketOptions {
     fn default() -> Self {
-        Self { linger: None }
+        Self {
+            // Setting this to zero by default.
+            linger: Some(Duration::ZERO),
```
ppenna, line 43: Shouldn't be the default MSL?

iyzhang: Yes, it should be None, but I set it to 0 to avoid needing to set it for every single one of our tests. We should have this as a configuration option, I will add a TODO and a link to my PR #1265

ppenna: So it sounds like a fix to #1265 must come first in the commit history, right?

src/rust/catnip/runtime/mod.rs
```diff
@@ -431,12 +355,42 @@ impl DerefMut for SharedDPDKRuntime {
...
+        };
+
+        let (mm, port_id, link_addr) = Self::initialize_dpdk(
+            &config.eal_init_args()?,
+            config.enable_jumbo_frames()?,
+            config.mtu()?,
+            tcp_offload.unwrap_or(false),
+            udp_offload.unwrap_or(false),
+        )
+        .unwrap();
```
ppenna, line 384: Please, fail here instead of panic.

ppenna (approved):
LGTM

PR #1261, opened 2024-05-06
[shim] Correctly Interpose `fcntl()`

### LINKED ISSUE #1260: [shim] Correctly Interpose `fcntl()`
## Description

We are not interposing `fcntl()`, as it is a vardic function.

We should parse the command and var list and then forward the call.

ppenna:
## Description

This PR fixes #1260


iyzhang (approved):
Thanks for fixing this

PR #1264, opened 2024-05-07
[catnap] Don't Fail Silently When Cannot Set Socket Options

ppenna:
## Description

This PR fixes https://github.com/microsoft/demikernel/issues/1263


iyzhang (approved):
Thanks for fixing this

PR #1266, opened 2024-05-08
[test] Simultaneous Close

ppenna:
## Description

This PR adds a simultaneous close test case to our suite.


iyzhang (approved):
This looks great, although my last PR changes our default to linger on and time out of 0. We might want to discuss tomorrow about what the default should be.

ppenna:
> This looks great, although my last PR changes our default to linger on and time out of 0. We might want to discuss tomorrow about what the default should be.

The default timeout should be 2*MSL.

PR #1274, opened 2024-05-13
[libos] Merge Config Files

### LINKED ISSUE #1265: [config] Catnap section of the config.yaml file should be changed to TCP socket defaults
## Context

Currently we have a catnap section in the config file that we mostly use for CatnapW. It sets the defaults for TCP keep alive and linger options on sockets that we open. However, these socket options apply to all of our stack and we should have a way to configure the expected default for these.

## Proposed Solution

We move the Catnap section to a TCP option defaults section and keep the two current options. We add support for this section to every libOS at the shared network queue level.

iyzhang:
This PR closes #1265 and reworks our config to take all settings from the config file with the option to override simple config values from environment variables.

The config file should be
```
demikernel:
  local_ipv4_addr: 10.9.1.12
  local_link_addr: "00:22:48:7bf6:c5"
raw_socket:
  interface_name: "eth1"
dpdk:
  eal_init: ["-c", "0xff", "-n", "4", "-a", "00:02.0","--proc-type=auto", "--vdev=net_vdev_netvsc0,iface=eth1"]
tcp_socket_options:
  keepalive:
    enabled: false
    time_millis: 0
    interval: 0
  linger:
    enabled: true
    time_seconds: 0
  nodelay: true
inetstack_config:
  mtu: 1500
  mss: 1500
  enable_jumbo_frames: false
  udp_checksum_offload: false
  tcp_checksum_offload: false
```


new_config.yaml~
ppenna, line 1: Please, remove this file from the PR.

src/rust/catloop/transport.rs
```diff
@@ -35,6 +35,7 @@ use ::std::{
         DerefMut,
     },
 };
+use std::net::Ipv4Addr;
```
ppenna, line 38: Please, prefix with `::`

src/rust/catnip/runtime/mod.rs
```diff
@@ -131,41 +131,49 @@ pub struct SharedDPDKRuntime(SharedObject<DPDKRuntime>);
...
-            &config.eal_init_args(),
-            config.use_jumbo_frames(),
+            &config.eal_init_args()?,
+            config.enable_jumbo_frames()?,
             config.mtu()?,
-            config.tcp_checksum_offload(),
-            config.udp_checksum_offload(),
+            tcp_offload.unwrap_or(false),
+            udp_offload.unwrap_or(false),
         )
```
ppenna, line 150: Do not proceed silently. If options were not present warn about default behavior.

src/rust/runtime/network/types/macaddr.rs
```diff
@@ -8,6 +8,7 @@
 use crate::runtime::fail::Fail;
 use ::libc::EINVAL;
 use ::std::fmt;
+use std::str::FromStr;
```
ppenna, line 11: Please prefix with `::`

src/rust/demikernel/config.rs
```diff
@@ -40,159 +93,334 @@ pub struct Config(pub Yaml);
...
+    /// Tcp socket option: Reads TCP keepalive settings as a `tcp_keepalive` structure from "tcp_keepalive" subsection.
+    pub fn tcp_keepalive(&self) -> Result<KeepAlive, Fail> {
+        let section: &Yaml = Self::get_subsection(self.get_tcp_socket_options()?, tcp_socket_options::KEEP_ALIVE)?;
+        let onoff: bool = Self::get_bool_option(section, "enabled")?;
 
-        local_interface_name.to_string()
+        #[cfg(target_os = "windows")]
+        let keepalivetime: u32 = Self::get_int_option(section, "time_millis")?;
+        #[cfg(target_os = "windows")]
+        let keepaliveinterval: u32 = Self::get_int_option(section, "interval")?;
```
ppenna, line 182: Can these option names be the same?

src/rust/catnip/runtime/mod.rs
iyzhang: done

ppenna (commented):
⁠There is still some work left:

- Fix instructions in [setup.md](https://github.com/microsoft/demikernel/blob/8bdecb80f98d000c2064e9e0a6aa59c5e48e0bc3/doc/setup.md#5-setup-configuration-file-only-once).
- Fix [demi_getsockopt.md](https://github.com/microsoft/demikernel/blob/8bdecb80f98d000c2064e9e0a6aa59c5e48e0bc3/man/demi_getsockopt.md) to include newly supported options.
- Fix [config.yaml](https://github.com/microsoft/demikernel/blob/8bdecb80f98d000c2064e9e0a6aa59c5e48e0bc3/scripts/config/default.yaml) to include ARP table.

PR #1285, opened 2024-05-16
[runtime] Workaround: Adjust TIMER resolutions to fix perf regression

anandbonde:
Bing observed performance regression between Demikernel builds. The TIMER resolution changes that were introduced between the build were the root cause. So, tuning these values manually to work around this issue for now. Refer to attached files for more details.

[RCA.pdf](https://github.com/microsoft/demikernel/files/15339187/RCA.pdf)
[TestWorkaround.pdf](https://github.com/microsoft/demikernel/files/15339194/TestWorkaround.pdf)


ppenna (approved):
LGTM

PR #1286, opened 2024-05-17
[bugfix] Workaround: Downgrade libc

ppenna:
## Description

This PR is a workaround for the `libc` crate that got updated in a way that broke our build.


anandbonde (approved):
Looks good!

PR #1291, opened 2024-05-23
[shim] Shim race conditions

stolet:
This pull request deals with a couple of issues with the shim library
- Fixes race conditions when initializing the demikernel
- Makes sure every call to the demikernel is wrapped with the DEMI_CALL macro (there were a few calls that we forgot about and that were causing issues)
- Sets errno before returning an error from getpeername

### DIFF
--- a/shim/src/ctrl/getsockopt.c
+++ b/shim/src/ctrl/getsockopt.c
@@ -5,6 +5,7 @@
 #include "../log.h"
 #include "../qman.h"
 #include "../utils.h"
+#include <glue.h>
 #include <demi/libos.h>
 #include <errno.h>
 #include <netinet/tcp.h>
@@ -58,5 +59,5 @@ int __getsockopt(int sockfd, int level, int optname, void *optval, socklen_t *op
         WARN("%s is not supported", "TCP_ULP");
     }
 
-    return (demi_getsockopt(sockfd, level, optname, optval, optlen));
+    return (__demi_getsockopt(sockfd, level, optname, optval, optlen));
 }
--- a/shim/src/ctrl/getsockopt.c
+++ b/shim/src/ctrl/getsockopt.c
@@ -59,5 +59,12 @@ int __getsockopt(int sockfd, int level, int optname, void *optval, socklen_t *op
         WARN("%s is not supported", "TCP_ULP");
     }
 
-    return (__demi_getsockopt(sockfd, level, optname, optval, optlen));
+    ret = __demi_getsockopt(sockfd, level, optname, optval, optlen);
+    if (ret != 0)
+    {
+        errno = ret;
+        return -1;
+    }
+
+    return (ret);
 }
--- a/shim/src/ctrl/setsockopt.c
+++ b/shim/src/ctrl/setsockopt.c
@@ -3,6 +3,7 @@
 
 #include "../log.h"
 #include "../qman.h"
+#include <glue.h>
 #include <demi/libos.h>
 #include <errno.h>
 #include <sys/socket.h>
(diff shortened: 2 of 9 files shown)


shim/src/ctrl/getsockopt.c
```diff
@@ -58,5 +59,5 @@ int __getsockopt(int sockfd, int level, int optname, void *optval, socklen_t *op
         WARN("%s is not supported", "TCP_ULP");
     }
 
-    return (demi_getsockopt(sockfd, level, optname, optval, optlen));
+    return (__demi_getsockopt(sockfd, level, optname, optval, optlen));
```
ppenna, line 62: Shouldn't we do the same that we do in `getpeername()`? That isCheck if return value is different from zero. If it is, we should set errno and return `-1`.

shim/src/ctrl/setsockopt.c
```diff
@@ -58,5 +59,5 @@ int __setsockopt(int sockfd, int level, int optname, const void *optval, socklen
         WARN("%s is not supported", "TCP_ULP");
     }
 
-    return (demi_setsockopt(sockfd, level, optname, optval, optlen));
+    return (__demi_setsockopt(sockfd, level, optname, optval, optlen));
```
ppenna, line 62: Shouldn't we do the same as we do in `getpeername()`? That isCheck if return value is different from zero. If it is, we should set errno and return `-1`.

shim/src/interpose.c
```diff
@@ -94,50 +95,88 @@ static int (*libc_epoll_create1)(int) = NULL;
...
+            }
+            MEM_BARRIER();
+        }
+    }
 }
 
 static void init(void)
 {
-    if (!initialized)
+    int ret;
```
ppenna, line 158: Can we move this to inner scope?

shim/src/interpose.c
```diff
@@ -94,50 +95,88 @@ static int (*libc_epoll_create1)(int) = NULL;
...
+            MEM_BARRIER();
+            initialized_libc = 1;
+        }
+        else
+        {
+            while (initialized_libc == 0)
+            {
+                sched_yield();
+            }
+            MEM_BARRIER();
```
ppenna, line 151: Shouldnt this `MEM_BARRIER` be after the loop, so as to foce a fetch from main memory of `initialized_libc`?

shim/src/interpose.c
```diff
@@ -94,50 +95,88 @@ static int (*libc_epoll_create1)(int) = NULL;
...
+            initialized = 1;
+        }
+        else
+        {
+            while (initialized == 0)
+            {
+                sched_yield();
+            }
+            MEM_BARRIER();
```
ppenna, line 178: Shouldn't the `MEM_BARRIER` be in the loop?

stolet: The variable `initialized` is declared as volatile and from my understanding this should prevent the compiler from optimizing things related to this variable and makes sure to fetch the updated value instead of an outdated cached version of it. The `MEM_BARRIER` at the end is to prevent instructions after `init()` from being reordered to somewhere during the initialization code inside `init()`.

shim/src/interpose.c
stolet: Same reasoning as the `initialized` value below.

ppenna: Ok.

shim/src/interpose.c
ppenna: Ok.

shim/src/interpose.c
ppenna: When we call `__init()`

shim/src/ctrl/setsockopt.c
```diff
@@ -58,5 +61,14 @@ int __setsockopt(int sockfd, int level, int optname, const void *optval, socklen
...
+    ret = __demi_setsockopt(sockfd, level, optname, optval, optlen);
+    // TODO: Add SO_REUSEADDR and uncomment this out. Without the
+    //       SO_REUSEADDR option implemented the Redis pipeline will fail.
+    // if (ret != 0)
+    // {
+    //     errno = ret;
+    //     return -1;
+    // }
+
+    return (ret);
```
ppenna, line 73: OK

ppenna (approved):
LGTM

PR #1295, opened 2024-05-25
[catnap] Enhancement: Yield to kernel 

iyzhang:
This PR adds a yield to Catnap into the kernel for a configurable amount of time, currently set to 1 milliseconds. This ensures that we do not continuously poll and run the CPU at 100%, while still occasionally returning to Demikernel, every second. This is safe in Catnap because there are no background coroutines, however, it does mean that every poll takes at least 1 milliseconds to complete.


iyzhang:
Closing this for now since we have a new PR #1485 that reorganizes task groups

PR #1296, opened 2024-05-30
[runtime] Enhancement: Move background coroutines to own group

iyzhang:
This PR separates coroutines that perform I/O from background coroutines in the runtime using the new scheduler task groups. Splitting these two lets us distinguish between coroutines where we are waiting on a result and coroutines that run forever. We also have the ability to poll a single coroutine, so we replace our poll on every system call with a poll to the specific coroutine.


iyzhang:
this PR is superseded by PR #1485

PR #1298, opened 2024-05-31
[libos] Bug Fix: Remove profiler scopes from coroutine polling functions

iyzhang:
This PR removes an error where we lose the current scope in the profiler. Because all coroutines belong to the root scope, when we enter a coroutine while in another scope, that scope is not saved. Normally, any scope entered within a previous scope is rooted at the previous scope. However, this would create a confusing hierarchy for the coroutines where they all belong to what ever scope they were first scheduled in. Thus, we consider scopes that enter coroutines to be invalid for our profiler.


anandbonde (approved):
Looks good.

PR #1299, opened 2024-05-31
[ci] Move from connection-string to managed identity approach

anandbonde:
This PR moves from the connection-string based approach to managed identity for security compliance.

`Note: The secrets file needs to change for this PR to work. The change is to remove the connection string and add the (storage account url) endpoint to it.`


iyzhang (approved):
LGTM, thanks for working on this

anandbonde:
Superseded by PR https://github.com/microsoft/demikernel/pull/1302.

PR #1305, opened 2024-06-07
[runtime] Adaptable timer resolution based on CPU cycles

### LINKED ISSUE #1226: [runtime] More accurate timeouts with rdtsc
## Context

Currently we use a configurable value for timer resolution which is based on the number of iterations that we run the scheduler. Since our tasks vary in runtime, this is an extremely imprecise way to decide when to check the system time for a timeout.

## Proposed Solution

Use rdtsc to estimate a cycle budget that matches how frequently we want to check the time. For example, 1 second = roughly 2.6 million cycles on a 2.6Ghz CPU. Once we run out the budget, we take the system time and figure out how long it actually took for each cycle. Then we adjust the budget for the next iteration based on that.

anandbonde:
This PR closes #1226.

### DIFF
--- a/src/rust/runtime/mod.rs
+++ b/src/rust/runtime/mod.rs
@@ -27,6 +27,7 @@ pub use queue::{
     QType,
 };
 pub use scheduler::TaskId;
+use x86::time::rdtscp;
 
 #[cfg(feature = "libdpdk")]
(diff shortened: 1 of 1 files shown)


anandbonde:
Timer resolution is no longer relevant. Closing this PR.

PR #1306, opened 2024-06-12
[ci] Archive perf data

anandbonde:
Archive perf data to GitHub for later retrieval/analysis.


anandbonde:
Will reopen when actually needed

PR #1310, opened 2024-06-17
[catpowder] XDP LibOS

### LINKED ISSUE #957: Add a LibOS using XDP
## Context

DPDK has scoped support for specific NICs. If a NIC is unsupported through this avenue, current acceleration is limited to io_uring or raw sockets. Windows clients have no alternatives. Adding a new LibOS using Express Data Path could provide a fast alternative which does not require NIC support.

## Proposed Solution

Add a new LibOS which uses XDP/BPF technologies to implement network I/O. OS-specific solutions are required: *nix platforms use BPF API and AF_XDP sockets (which may also benefit from io_uring). Windows platforms require [XDP for Windows](https://github.com/microsoft/xdp-for-windows). Layering needs further consideration w.r.t. io_uring and BPF.

## Alternative Solutions

RDMA/RoCE provide another avenue for acceleration, although the complexity is arguably much higher than XDP (not to mention NIC-specific support). Both solutions require OS-specific considerations: Windows supports a variant of libibverbs, but it's still divergent from the *nix support. Registered I/O may provide an acceleration avenue on Windows with similar characteristics to io_uring or raw sockets on *nix. "XDP for Windows" seems to be the spiritual successor to RIO. Viewing RIO as a companion to io_uring in providing acceleration alternatives to all OSes, XDP seems a more cogent/converged solution (even with different implementations for *nix/Windows).

ppenna:
## Description

This PR introduces a XDP LibOS for Demikernel. It closes #957


iyzhang:
This PR closes #957

src/rust/catpowder/win/runtime.rs
```diff
@@ -0,0 +1,173 @@
...
+    BorrowMut,
+};
+
+//======================================================================================================================
+// Structures
+//======================================================================================================================
+
+/// A LibOS built on top of Windows XDP.
+#[derive(Clone)]
+pub struct CatpowderRuntime(SharedObject<CatpowderRuntimeInner>);
```
iyzhang, line 46: Please label this SharedCatpowderRuntime, per our naming convention, which requires that we add Shared* in front of any SharedObject.

iyzhang (approved):

PR #1312, opened 2024-06-19
[testing] Enhancement: Turn on more inetstack tests

iyzhang:
This PR turns on more TCP echo tests. The goal of this PR is to align Catpowder and Catnip with the same set of tests that Catnap currently runs on our CI.


anandbonde (approved):
Looks good.

PR #1315, opened 2024-06-24
[ci] Flamegraph and stack perf data

anandbonde:
Creates flamegraph and stack perf data for PRs.

Flamegraphs (in SVG format) will be stored in archives. Stack perf data will be archived as well as posted as comments on PRs.

### DIFF
--- a/src/rust/catnap/linux/transport.rs
+++ b/src/rust/catnap/linux/transport.rs
@@ -39,6 +39,7 @@ use crate::{
         SharedDemiRuntime,
         SharedObject,
     },
+    timer,
 };
 use ::futures::FutureExt;
 use ::slab::Slab;
@@ -117,7 +118,7 @@ impl SharedCatnapTransport {
         }));
         let mut me2: Self = me.clone();
         runtime.insert_background_coroutine(
-            "catnap::transport::epoll",
+            "bgc::catnap::transport::epoll",
             Box::pin(async move { me2.poll().await }.fuse()),
         )?;
         Ok(me)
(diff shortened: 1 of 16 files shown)


src/rust/catnap/linux/transport.rs
```diff
@@ -38,7 +38,7 @@ use crate::{
         DemiRuntime,
         SharedDemiRuntime,
         SharedObject,
-    },
+    }, timer,
```
ppenna, line 41: Not sure if this was auto-formatted (if it was, just ignore).
To keep consistency, could you move this to the next line?

src/rust/catnap/win/overlapped.rs
ppenna, line 1: Dumb question: what does IOC stand for?

src/rust/catnap/linux/transport.rs
ppenna, line 1: Dumb question: what does bgc stand for?

ppenna (approved):
LGTM

src/rust/catnap/linux/transport.rs
anandbonde: Yes, I will move it. Thanks.

src/rust/catnap/win/overlapped.rs
anandbonde: ioc -> io coroutine

src/rust/catnap/linux/transport.rs
anandbonde: bgc -> background coroutine (I abbreviated it so that the collapsed stack is not a very long string).

PR #1317, opened 2024-06-25
[ci] Platform Switch Option

ppenna:
# Description

This PR closes https://github.com/microsoft/demikernel/issues/1316


tools/ci/config/test/linux/catnapw.yaml
anandbonde, line 1: Is this file needed anymore?

tools/ci/config/test/windows/catnap.yaml
anandbonde, line 1: This file is identical to the `linux/catnap.yaml` file. Do you foresee these 2 files diverging in future? Also, if we run the same commands for linux/windows, why do we need platform distinction at all? Under what circumstances do the CI runs for the 2 platforms differ? (I know that paths may be 1 difference, what are the others?)

tools/demikernel_ci.py
```diff
@@ -71,7 +72,7 @@ def run_pipeline(
     # STEP 4: Run integration tests.
     if test_integration:
         if status["checkout"] and status["compile"]:
-            if libos == "catnap" or libos == "catnapw" or libos == "catloop" or libos == "catnip" or libos == "catpowder":
+            if libos == "catnap" or libos == "catloop" or libos == "catnip" or libos == "catpowder":
```
anandbonde, line 75: How about using the `libos in [...]` syntax here?

anandbonde (approved):
Looks good.

tools/ci/config/test/linux/catnapw.yaml
ppenna: No it is not. Thanks for catching this.

tools/ci/config/test/windows/catnap.yaml
ppenna: Thanks fot cathing this. I fixed it so it is a symlink instead.

tools/demikernel_ci.py
ppenna: Oh yeah!

PR #1319, opened 2024-06-27
[inetstack] Re-org

iyzhang:
Small re-org of the inetstack in preparation for larger changes. This PR does the following:
1. Moves the hashttlcache to the crate::collections module
2. Removes the public access to the TCP and UDP modules from the top level mod file and instead calls through the IPv4 peer.


ppenna (approved):
LGTM

PR #1324, opened 2024-07-02
[inetstack] Better layering

iyzhang:
This PR moves the NetworkRuntime to the inetstack and renames it the PhysicalLayer. Then it enforces the direct OSI hierarchy through the inetstack where the Inetstack only has a reference to layer 4 and the next layer only has a reference to the next layer.


ppenna (approved):
LGTM

PR #1325, opened 2024-07-02
[shim] Optimizing epoll_wait

stolet:
Enhancement to the shim, where instead of looking at every element of the array for new events, we look only at added events.

### DIFF
--- a/shim/include/glue.h
+++ b/shim/include/glue.h
@@ -12,6 +12,7 @@
 #include <unistd.h>
 
 extern int is_reentrant_demi_call();
+extern void init_reent_guards();
 
 extern int __demi_init(const struct demi_args *args);
 extern int __demi_create_pipe(int *memqd_out, const char *name);
--- a/shim/include/glue.h
+++ b/shim/include/glue.h
@@ -70,8 +70,6 @@ extern int __epoll_create(int size);
 extern int __epoll_create1(int flags);
 extern int __epoll_ctl(int epfd, int op, int fd, struct epoll_event *event);
 extern int __epoll_wait(int epfd, struct epoll_event *events, int maxevents, int timeout);
-extern void print_table(void);
-
 
 //======================================================================================================================
 
--- a/shim/include/glue.h
+++ b/shim/include/glue.h
@@ -70,6 +70,8 @@ extern int __epoll_create(int size);
 extern int __epoll_create1(int flags);
 extern int __epoll_ctl(int epfd, int op, int fd, struct epoll_event *event);
 extern int __epoll_wait(int epfd, struct epoll_event *events, int maxevents, int timeout);
+extern void print_table(void);
+
 
 //======================================================================================================================
 
--- a/shim/src/epoll/epoll.c
+++ b/shim/src/epoll/epoll.c
@@ -2,10 +2,7 @@
 // Licensed under the MIT license.
 
 #include "../epoll.h"
-#include "../log.h"
 #include <assert.h>
-
-#include <stdio.h>
 
 struct epoll_table
 {
@@ -36,11 +33,18 @@ void epoll_table_init(void)
     for (int i = 0; i < EPOLL_MAX_FDS; i++)
     {
         epoll_table[i].used = 0;
-        epoll_table[i].head = INVALID_EV;
-        epoll_table[i].tail = INVALID_EV;
+        epoll_table[i].head = 0;
+        epoll_table[i].tail = 0;
         for (int j = 0; j < MAX_EVENTS; j++)
         {
-            epoll_table[i].events[j].used = 0;
+            /* We use an empty element in the start of the list, to
+               make the checks for insertion and deletion simpler, thus
+               mark the first element as being used. */
+            if (j == 0)
+                epoll_table[i].events[j].used = 1;
+            else
+                epoll_table[i].events[j].used = 0;
+
             epoll_table[i].events[j].id = j;
             epoll_table[i].events[j].next_ev = INVALID_EV;
             epoll_table[i].events[j].prev_ev = INVALID_EV;
@@ -90,24 +94,20 @@ void epoll_set_tail(int epfd, int i)
 
 struct demi_event *epoll_get_next(int epfd, struct demi_event *ev)
 {
-    int i;
-
     if (ev == NULL || ev->next_ev == INVALID_EV)
         return NULL;
 
-    i = ev->next_ev;
+    int i = ev->next_ev;
 
     return (epoll_get_event(epfd, i));
 }
 
 struct demi_event *epoll_get_prev(int epfd, struct demi_event *ev)
 {
-    int i;
-
     if (ev == NULL || ev->prev_ev == INVALID_EV)
         return NULL;
 
-    i = ev->prev_ev;
+    int i = ev->prev_ev;
 
     return (epoll_get_event(epfd, i));
 }
--- a/shim/src/epoll/epoll.c
+++ b/shim/src/epoll/epoll.c
@@ -5,9 +5,13 @@
 #include "../log.h"
 #include <assert.h>
 
+#include <stdio.h>
+
 struct epoll_table
 {
     int used;
+    int head;
+    int tail;
     struct demi_event events[MAX_EVENTS];
 };
 
@@ -32,8 +36,15 @@ void epoll_table_init(void)
     for (int i = 0; i < EPOLL_MAX_FDS; i++)
     {
         epoll_table[i].used = 0;
+        epoll_table[i].head = INVALID_EV;
+        epoll_table[i].tail = INVALID_EV;
         for (int j = 0; j < MAX_EVENTS; j++)
+        {
             epoll_table[i].events[j].used = 0;
+            epoll_table[i].events[j].id = j;
+            epoll_table[i].events[j].next_ev = INVALID_EV;
+            epoll_table[i].events[j].prev_ev = INVALID_EV;
+        }
     }
 }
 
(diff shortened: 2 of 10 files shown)


shim/include/glue.h
```diff
@@ -69,6 +70,8 @@ extern int __epoll_create(int size);
 extern int __epoll_create1(int flags);
 extern int __epoll_ctl(int epfd, int op, int fd, struct epoll_event *event);
 extern int __epoll_wait(int epfd, struct epoll_event *events, int maxevents, int timeout);
+extern void print_table(void);
```
ppenna, line 73: Where is this function defined?

shim/src/hashtable.c
```diff
@@ -120,7 +117,7 @@ void hashtable_remove(struct hashtable *h, int key)
             return;
         }
 
-        hash = (hash + 1) & mask;
+        hash = hash + 1;
```
ppenna, line 120: We should keep modulo operation here, otherwise we may end up writing outside the bounds of the array.

shim/src/hashtable.c
```diff
@@ -81,19 +80,18 @@ uint64_t hashtable_get(struct hashtable *h, int key)
...
-    hash = key & mask;
+    hash = key % length;
 
     do
     {
         if (h->keys[hash] == key)
             return (h->values[hash]);
 
-        hash = (hash + 1) & mask;
+        hash = hash + 1;
```
ppenna, line 94: We should keep modulo operation here, otherwise we may end up writing outside the bounds of the array.

shim/src/hashtable.c
```diff
@@ -51,13 +51,12 @@ int hashtable_insert(struct hashtable *h, int key, uint64_t val)
...
     int hash = 0;
     const int length = (1 << h->length_log2);
-    const int mask = (length - 1);
 
     assert(h != NULL);
     assert(key != HASHTABLE_NULL);
     assert(val != (uint64_t)HASHTABLE_NULL);
 
-    hash = key & mask;
+    hash = key % length;
```
ppenna, line 59: Since length is a power of two, why should we use expensive modulo operations?

shim/src/hashtable.c
```diff
@@ -81,19 +80,18 @@ uint64_t hashtable_get(struct hashtable *h, int key)
     int skip = 0;
     int hash = 0;
     const int length = (1 << h->length_log2);
-    const int mask = (length - 1);
 
     assert(h != NULL);
     assert(key != HASHTABLE_NULL);
 
-    hash = key & mask;
+    hash = key % length;
```
ppenna, line 87: Since length is a power of two, why should we use expensive modulo operations?

shim/src/hashtable.c
```diff
@@ -104,12 +102,11 @@ void hashtable_remove(struct hashtable *h, int key)
...
```
ppenna, line 109: Since length is a power of two, why should we use expensive modulo operations?

shim/src/hashset.c
ppenna, line 1: Please see my comments on the `hashtable.c` file.

shim/src/glue.c
```diff
@@ -1,23 +1,38 @@
...
 
-#define DEMI_CALL(type, fn, ...)    \
-    {                               \
-        __demi_reent_guard = 1;     \
-        type ret = fn(__VA_ARGS__); \
-        __demi_reent_guard = 0;     \
-        return (ret);               \
+#define DEMI_CALL(type, fn, ...)                        \
+    {                                                   \
+        pid_t tid = getpid();                           \
```
ppenna, line 20: Well, this actually gives you the process identifier, which should be the same for all threads... Did you mean `gettid()` instead?

shim/src/interpose.c
```diff
@@ -54,10 +54,10 @@
 
 #define INTERPOSE_CALL(type, fn_libc, fn_demi, ...)           \
     {                                                         \
-        bool reentrant = is_reentrant_demi_call();            \
-                                                              \
         init_libc();                                          \
```
ppenna, line 57: What is the rationale for that?

shim/src/epoll/epoll_ctl.c
```diff
@@ -16,6 +16,9 @@
 
 static int __do_demi_epoll_ctl_add(int epfd, int fd, struct epoll_event *event)
 {
+    int i;
```
ppenna, line 19: What is this variable used for?

shim/src/epoll/epoll_ctl.c
```diff
@@ -16,6 +16,9 @@
...
+    struct demi_event *ev, *tail, *head;
```
ppenna, line 20: Please initialize all with `NULL`.

shim/src/epoll/epoll_ctl.c
```diff
@@ -38,18 +41,35 @@ static int __do_demi_epoll_ctl_add(int epfd, int fd, struct epoll_event *event)
     for (int i = 0; i < MAX_EVENTS; i++)
     {
         // Found.
-        struct demi_event *ev = epoll_get_event(epfd, i);
+        ev = epoll_get_event(epfd, i);
```
ppenna, line 44: Why have you nived `ev` declaration to the outter scope? The lifetime of `ev` does not need to be that long.

shim/src/epoll/epoll.c
```diff
@@ -5,9 +5,13 @@
 #include "../log.h"
 #include <assert.h>
 
+#include <stdio.h>
```
ppenna, line 8: Are we using this?

shim/src/epoll/epoll.c
```diff
@@ -42,3 +53,61 @@ struct demi_event *epoll_get_event(int epfd, int i)
...
+struct demi_event *epoll_get_next(int epfd, struct demi_event *ev)
+{
+    int i;
+
+    if (ev == NULL || ev->next_ev == INVALID_EV)
+        return NULL;
+
+    i = ev->next_ev;
+
+    return (epoll_get_event(epfd, i));
```
ppenna, line 100: Inline this variable. This is not for performance reason, but to reduce surface of bugs :)

shim/src/epoll/epoll.c
```diff
@@ -42,3 +53,61 @@ struct demi_event *epoll_get_event(int epfd, int i)
...
+struct demi_event *epoll_get_prev(int epfd, struct demi_event *ev)
+{
+    int i;
+
+    if (ev == NULL || ev->prev_ev == INVALID_EV)
+        return NULL;
+
+    i = ev->prev_ev;
+
+    return (epoll_get_event(epfd, i));
```
ppenna, line 112: Inline this variable. This is not for performance reason, but to reduce surface of bugs :)

shim/src/epoll/epoll_wait.c
```diff
@@ -17,6 +17,7 @@
 int __epoll_wait(int epfd, struct epoll_event *events, int maxevents, int timeout)
 {
     int nevents = 0;
+    struct demi_event *ev;
```
ppenna, line 20: Move this to initialization.

shim/src/epoll/epoll_ctl.c
```diff
@@ -38,18 +41,35 @@ static int __do_demi_epoll_ctl_add(int epfd, int fd, struct epoll_event *event)
...
+                assert(tail == NULL);
+                epoll_set_head(epfd, ev->id);
+                epoll_set_tail(epfd, ev->id);
+            }
+            else
+            {
+                ev->prev_ev = tail->id;
+                tail->next_ev = ev->id;
+                epoll_set_tail(epfd, ev->id);
+            }
```
ppenna, line 68: If can be avoided if the head element of the linked list is made a guard.

shim/include/glue.h
stolet: Sorry, I forgot this behind. I deleted it now.

shim/src/hashtable.c
stolet: Ok, this makes sense. I didn't realize that the modulo operation was more expensive. I reverted it back to the mask.

shim/src/glue.c
stolet: I did, I changed it back to `gettid()` now.

shim/src/interpose.c
stolet: I need to initialize the reentrancy guards hashset before calling is_reentrant_demi_call. The initialization for that is in init_libc. I'll change the name of that function to be more descriptive, since now it is not initializing only libc.

shim/src/epoll/epoll_ctl.c
stolet: It was a habit of declaring variables in the beginning of the function. Some recent projects I was working on followed this code style. There are some other comments related to this. I will keep the declarations in the inner scope when appropriate (such as inside the for loop in this instance), to be consistent with the rest of the code in the demikernel.

shim/src/glue.c
```diff
@@ -1,23 +1,38 @@
...
 
 int is_reentrant_demi_call()
 {
-    return __demi_reent_guard;
+    pid_t tid = gettid();
+    return hashset_contains(__demi_reent_guards, tid);
+}
+
+void init_reent_guards() {
+    __demi_reent_guards = hashset_create(10);
```
ppenna, line 34: Could you create a constant for this magic value? Also, could you add some rationale behind this value? For instance, why have you cherry-picked `10`?

shim/src/glue.c
```diff
@@ -1,23 +1,38 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
-
+#define _GNU_SOURCE
 #include <demi/libos.h>
 #include <demi/sga.h>
 #include <demi/wait.h>
 
-static __thread int __demi_reent_guard = 0;
+#include <string.h>
```
ppenna, line 8: Why do we need `string.h`?

shim/src/interpose.c
```diff
@@ -95,19 +95,20 @@ static int (*libc_epoll_create1)(int) = NULL;
...
-static volatile uint8_t initialized_libc = 0;
-static volatile uint8_t in_init_libc = 0;
+static volatile uint8_t intialized_preinit = 0;
+static volatile uint8_t in_preinit = 0;
 
 static volatile uint8_t initialized = 0;
 static volatile uint8_t in_init = 0;
 
-static void init_libc(void)
+static inline void preinit(void)
```
ppenna, line 104: Why do we need to change the name of this function?

shim/src/interpose.c
```diff
@@ -153,7 +157,7 @@ static void init_libc(void)
     }
 }
 
-static void init(void)
+static inline void init(void)
```
ppenna, line 160: To foce `inline` you may want to consider overloading the keyword with GCC's attribute.You may consider putting this in `cc.h`:

```
#define inline __attribute__((always_inline))
```

shim/src/epoll/epoll_ctl.c
```diff
@@ -126,39 +132,38 @@ static int __do_demi_epoll_ctl_mod(int epfd, int fd, struct epoll_event *event)
 
 static int __do_demi_epoll_ctl_del(int epfd, int fd)
 {
+    struct demi_event *ev, *prev, *next, *head, *tail;
```
ppenna, line 135: Move whatever is possible for first definition.

shim/src/epoll/epoll.c
```diff
@@ -32,8 +33,22 @@ void epoll_table_init(void)
...
         for (int j = 0; j < MAX_EVENTS; j++)
-            epoll_table[i].events[j].used = 0;
+        {
+            /* We use an empty element in the start of the list, to
+               make the checks for insertion and deletion simpler, thus
+               mark the first element as being used. */
+            if (j == 0)
+                epoll_table[i].events[j].used = 1;
+            else
+                epoll_table[i].events[j].used = 0;
```
ppenna, line 46: Use ternary operator instead.

ppenna:
We are almost there :)

shim/src/glue.c
stolet: That was accidentally left behind when I was making some temporary modifications. It has been removed now.

shim/src/glue.c
stolet: The rationale was mostly that it was the same value used for the other hashsets and hash tables (1024 elements), but I believe we may be able to use a smaller number, since I imagine the number of threads will be smaller than the number of file descriptors.

ppenna (approved):
LGTM

PR #1329, opened 2024-07-11
[tcp-echo] Measure end-to-end CPU cycles

anandbonde:
Measure the end-to-end CPU cycles for the tcp-echo test.

### DIFF
--- a/examples/tcp-echo/main.rs
+++ b/examples/tcp-echo/main.rs
@@ -18,6 +18,7 @@ use clap::{
 };
 use client::TcpEchoClient;
 use demikernel::{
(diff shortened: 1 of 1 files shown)


examples/tcp-echo/main.rs
ppenna, line 1: why can't we leverage the profiler macro?

anandbonde: We can, just updated the PR. Earlier, I thought it would be easier to parse the other way. But this makes sense. Thanks for point out!

ppenna (approved):
LGTM

PR #1331, opened 2024-07-17
[examples] Measure end-to-end CPU cycles

anandbonde:
Adding timers for main() in all examples.

### DIFF
--- a/examples/tcp-close/main.rs
+++ b/examples/tcp-close/main.rs
@@ -20,12 +20,13 @@ use crate::{
     client::TcpClient,
     server::TcpServer,
 };
-use ::std::time::Duration;
-use anyhow::Result;
-use demikernel::{
+use ::anyhow::Result;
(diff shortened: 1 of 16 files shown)


examples/tcp-close/main.rs
```diff
@@ -20,12 +20,13 @@ use crate::{
...
-use ::std::time::Duration;
-use anyhow::Result;
-use demikernel::{
+use ::demikernel::{
+    timer,
     LibOS,
     LibOSName,
 };
+use ::std::time::Duration;
+use anyhow::Result;
```
ppenna, line 29: Since you touched this line could you fix it and prefix it with `::`?

ppenna (approved):
LGTM

examples/tcp-close/main.rs
anandbonde: Done.

PR #1333, opened 2024-07-17
[nettest] Initial Support for an Additional Protocol

ppenna:
## Description

This PR introduces required to support to include an additional protocol in the network test simulator.

## Summary of Changes

- Fixed various typos.
- Added some error log messages.
- Introduce protocol support in the simulator.

### DIFF
--- a/src/rust/inetstack/protocols/udp/mod.rs
+++ b/src/rust/inetstack/protocols/udp/mod.rs
@@ -19,6 +19,9 @@ mod tests;
 //==============================================================================
 
 pub use self::{
-    datagram::UdpHeader,
+    datagram::{
+        UdpDatagram,
+        UdpHeader,
+    },
     peer::SharedUdpPeer,
 };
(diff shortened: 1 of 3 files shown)


src/rust/inetstack/protocols/udp/mod.rs
```diff
@@ -19,6 +19,9 @@ mod tests;
 //==============================================================================
 
 pub use self::{
-    datagram::UdpHeader,
+    datagram::{
+        UdpDatagram,
```
anandbonde, line 23: Is this required?

anandbonde (approved):
Looks good.

src/rust/inetstack/protocols/udp/mod.rs
ppenna: Yes, it is. The UDP datagram should be exposed to the network simulator, as it should be able to create and parse UDP datagrams. Note that TCP segment is also exposed. But good idea, I can investigate if we can expose it only on test builds.

PR #1337, opened 2024-07-20
[ci] Automatically bootstrap VMs

anandbonde:
With this, you will never have to manually run steps on VMs after they reboot! They would possibly have to be rebooted less often as well...

### DIFF
--- a/.github/workflows/catnip.yml
+++ b/.github/workflows/catnip.yml
@@ -46,6 +46,28 @@ jobs:
         echo -e "\tPasswordAuthentication no" >> $HOME/.ssh/config
         echo -e "\tUser ${{ secrets.USERNAME }}" >> $HOME/.ssh/config
         echo -e "\tPort ${{ secrets.PORTNUM }}" >> $HOME/.ssh/config
+    - name: Bootstrap VMs
+      shell: bash
+      run: |
+        echo "Running azure.sh on SERVER"
+        ssh $SERVER "sudo bash ~/demikernel/scripts/setup/azure.sh"
+
+        echo "Running hugepages.sh on SERVER"
+        ssh $SERVER "sudo bash ~/demikernel/scripts/setup/hugepages.sh"
+
+        echo "Running ip addr flush on SERVER"
+        ssh $SERVER "sudo ip -br addr show dev eth1 | grep -q ' UP ' && \
+                     sudo ip addr flush dev eth1"
+
+        echo "Running azure.sh on CLIENT"
+        ssh $CLIENT "sudo bash ~/demikernel/scripts/setup/azure.sh"
+
+        echo "Running hugepages.sh on CLIENT"
+        ssh $CLIENT "sudo bash ~/demikernel/scripts/setup/hugepages.sh"
+
+        echo "Running ip addr flush on CLIENT"
+        ssh $CLIENT "sudo ip -br addr show dev eth1 | grep -q ' UP ' && \
+                     sudo ip addr flush dev eth1"
     - name: Run
       run: |
         if [[ "${{ github.event_name }}" == "pull_request" ]]; then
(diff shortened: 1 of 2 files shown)


.github/workflows/catnip.yml
```diff
@@ -46,6 +46,28 @@ jobs:
...
+
+        echo "Running azure.sh on CLIENT"
+        ssh $CLIENT "sudo bash ~/demikernel/scripts/setup/azure.sh"
+
+        echo "Running hugepages.sh on CLIENT"
+        ssh $CLIENT "sudo bash ~/demikernel/scripts/setup/hugepages.sh"
+
+        echo "Running ip addr flush on CLIENT"
+        ssh $CLIENT "sudo ip -br addr show dev eth1 | grep -q ' UP ' && \
+                     sudo ip addr flush dev eth1"
```
ppenna, line 70: We may want to add these steps in the release pipeline as well. A retrigger of the pipeline for any reason, would not re-run the bootstrap steps.

.github/workflows/catpowder.yml
```diff
@@ -46,6 +46,24 @@ jobs:
...
+        ssh $SERVER "sudo ip -br addr show dev eth1 | grep -q ' UP ' && \
+                     sudo ip addr flush dev eth1"
+
+        echo "Running iptables rule on CLIENT"
+        ssh $CLIENT "sudo iptables -C OUTPUT -p tcp -o eth1 --tcp-flags RST RST -j DROP || \
+                     sudo iptables -A OUTPUT -p tcp -o eth1 --tcp-flags RST RST -j DROP"
+
+        echo "Running ip addr flush on CLIENT"
+        ssh $CLIENT "sudo ip -br addr show dev eth1 | grep -q ' UP ' && \
+                     sudo ip addr flush dev eth1"
```
ppenna, line 66: We may want to add these steps in the release pipeline as well. A retrigger of the pipeline for any reason, would not re-run the bootstrap steps.

.github/workflows/catnip.yml
anandbonde: They are already added in the release-pipeline section as well. (If you scroll to Ln no. 121)

.github/workflows/catpowder.yml
anandbonde: They are already added in the release-pipeline section as well. (If you scroll to Ln no. 117)

.github/workflows/catpowder.yml
```diff
@@ -46,6 +46,24 @@ jobs:
         echo -e "\tPasswordAuthentication no" >> $HOME/.ssh/config
         echo -e "\tUser ${{ secrets.USERNAME }}" >> $HOME/.ssh/config
         echo -e "\tPort ${{ secrets.PORTNUM }}" >> $HOME/.ssh/config
+    - name: Bootstrap VMs
+      shell: bash
+      run: |
+        echo "Running iptables rule on SERVER"
+        ssh $SERVER "sudo iptables -C OUTPUT -p tcp -o eth1 --tcp-flags RST RST -j DROP || \
+                     sudo iptables -A OUTPUT -p tcp -o eth1 --tcp-flags RST RST -j DROP"
```
ppenna, line 54: It is not guaranteed (nor safe) to assume that  `eth1` is the network interface that we should target (this may change on reboot). A safer automated proceedure would query the target interface based on the IP address (which we can get from our secrets).

.github/workflows/catnip.yml
```diff
@@ -46,6 +46,28 @@ jobs:
...
+      run: |
+        echo "Running azure.sh on SERVER"
+        ssh $SERVER "sudo bash ~/demikernel/scripts/setup/azure.sh"
+
+        echo "Running hugepages.sh on SERVER"
+        ssh $SERVER "sudo bash ~/demikernel/scripts/setup/hugepages.sh"
+
+        echo "Running ip addr flush on SERVER"
+        ssh $SERVER "sudo ip -br addr show dev eth1 | grep -q ' UP ' && \
+                     sudo ip addr flush dev eth1"
```
ppenna, line 60: It is not guaranteed (nor safe) to assume that  `eth1` is the network interface that we should target (this may change on reboot). A safer automated proceedure would query the target interface based on the IP address (which we can get from our secrets).

anandbonde: Can you share how this is done in the startup scripts? I will reuse the same mechanism here as well.

.github/workflows/catnip.yml
ppenna: For some reason they did not render in VS code. Sorry for signaling it wrong.

ppenna (approved):
This solution will not work for all cases. For instance, `eth1` may not be the correct network interface to target.

Nevertheless, based on our current fleet this is unlikely to happen. Thus, I accept this solution as a good workaround for now.

.github/workflows/catnip.yml
anandbonde: We discussed offline, and will fix it if (and when) required.

.github/workflows/catpowder.yml
anandbonde: Refer to other duplicate comment for details.

PR #1338, opened 2024-07-22
[nettest] Adding UDP Test Cases

ppenna:
## Description

This PR introduces some test cases for UDP in our simulator.

## Summary of Changes
- Changed TCP tests to be annotated with "TCP" packets (these eases reading of traces)
- Added single test case for blocking push
- Added 2 test cases for blocking pop
- Added 2 test cases for blocking push-pop


anandbonde (approved):
Looks good.

PR #1339, opened 2024-07-23
[runtime] Bug Fix: Remove task id leak in scheduler

iyzhang:
This PR fixes a leak in our scheduler that was causing the id map to grow continuously.

### DIFF
--- a/src/rust/runtime/scheduler/scheduler.rs
+++ b/src/rust/runtime/scheduler/scheduler.rs
@@ -162,8 +162,17 @@ impl Scheduler {
         assert!(self.current_running_task.is_some());
         let result: Option<Box<dyn Task>> = group.poll_notified_task_and_remove_if_ready(self.current_task_id);
         assert!(self.current_running_task.is_some());
-        *self.current_running_task = None;
+        // Expect is safe here because we just looked up the external id.
+        let task_id: TaskId = self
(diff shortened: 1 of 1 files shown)


anandbonde (approved):
Looks good

src/rust/runtime/scheduler/scheduler.rs
```diff
@@ -162,8 +162,14 @@ impl Scheduler {
...
         let result: Option<Box<dyn Task>> = group.poll_notified_task_and_remove_if_ready(self.current_task_id);
         assert!(self.current_running_task.is_some());
-        *self.current_running_task = None;
+        let task_id: TaskId = self
+            .current_running_task
+            .take()
+            .expect("must have a current running task");
         assert!(self.current_running_task.is_none());
+        if result.is_some() {
+            self.ids.remove(&task_id);
```
ppenna, line 171: We are ignoring the return value of `result`. Could you please `assert!()` that the return value was `Some()`?

iyzhang: We are not ignoring it. It gets passed back as the result of the function.

iyzhang: Ah got it.

PR #1341, opened 2024-07-23
[catnip] Add timers to recieve and transmit paths

anandbonde:
Add timers to recieve and transmit paths for catnip (includes inetstack receive/transmit).


ppenna (approved):
LGTM

PR #1343, opened 2024-07-24
[shim] Missing Initialization on Many Calls

ppenna:
## Description

This PR fixes initialization issues of Demikernel in the POSIX SHIM layer.

Without this modification, initialization would be performed only when `epoll_create()` is the very first call performed by the application.

### DIFF
--- a/shim/src/interpose.c
+++ b/shim/src/interpose.c
@@ -509,9 +509,8 @@ int epoll_create(int size)
 
     init();
 
+    // Create epoll on kernel side.
     int ret = libc_epoll_create(size);
-
-    // First, create epoll on kernel side.
     if (ret == -1)
     {
         ERROR("epoll_create() failed - %s", strerror(errno));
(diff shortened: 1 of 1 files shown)


shim/src/interpose.c
```diff
@@ -58,7 +58,7 @@
                                                               \
         bool reentrant = is_reentrant_demi_call();            \
                                                               \
-        if ((!initialized) || (reentrant))                    \
+        if (reentrant)                                        \
```
stolet, line 61: Instead of just having(reentrant) you should check if the shim is in the initialization step. For example (in_init || reentrant). Right now, without that check any calls meant to libc made during the initialization will get stuck in the while loop in init() waiting for the initialization to finish.

stolet (changes requested):
Without the above change the shim will hang in a multi-threaded program like memcached, but it should run fine after being addressed.

shim/src/interpose.c
ppenna: Sure!

stolet (approved):
LGTM

PR #1345, opened 2024-07-26
[inetstack] Bug Fix: Wait for ack

### LINKED ISSUE #1340: [inetstack] Ack queue grows out of control and only counts bytes
## Description

We use an AsyncQueue to track acks received, however, we never drain the ack queue.

## How to Reproduce

1. Run the tcp-echo system test
2. Periodically print length of ack queue.
3. Observe in the code that EstablishedSocket async push function is never called.

## Expected Behavior

The push coroutine should wait for the async push function to return.

iyzhang:
This PR replaces a wait for the ack to the push function in our inetstack. This was accidentally removed by me at some point. It close #1340

### DIFF
--- a/examples/rust/tcp-ping-pong.rs
+++ b/examples/rust/tcp-ping-pong.rs
@@ -1,10 +1,12 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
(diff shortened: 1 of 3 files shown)


ppenna (approved):
LGTM Thanks for fixing this.

anandbonde (approved):

PR #1350, opened 2024-07-31
[shim] Fix evaluation of conditional expression in epoll_create

stolet:
Incorrect parenthesis were causing the evaluation of conditional expression to be incorrect.


ppenna (approved):
LGTM

PR #1351, opened 2024-07-31
Enhancement shim datapath

stolet:
Adding a batch of optimizations for the shim layer
- Use static arrays instead of custom hashtable and hashset
- Hint to the compiler likely and unlikely events in branch conditions
- Add missing O3 flag to the shim makefile
- Use demi_wait_any instead of regular demi_wait to reduce the number of demi calls and background processes run (greatly improves tail latency)

### DIFF
--- a/shim/src/glue.c
+++ b/shim/src/glue.c
@@ -12,21 +12,29 @@
 #include "utils.h"
 #include "log.h"
 
-static int __demi_reent_guard;
+static struct hashset *__demi_reent_guards;
 
 #define MAX_THREADS_LOG2 10
 
 #define DEMI_CALL(type, fn, ...)                        \
     {                                                   \
-        __demi_reent_guard = 1;                         \
+        pid_t tid = gettid();                           \
+        hashset_insert(__demi_reent_guards, tid);       \
         type ret = fn(__VA_ARGS__);                     \
-        __demi_reent_guard = 0;                         \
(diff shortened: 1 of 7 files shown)


ppenna:
> Use demi_wait_any instead of regular demi_wait to reduce the number of demi calls and background processes run (greatly improves tail latency)

@stolet I cannot find this enhancement in this PR.

stolet:
For some reason the demi_wait enhancement didn't show up. I will create another pull request in a different branch so that we can add this change independently. I noticed that the other changes in this PR were broken up and added in different merges.

shim/src/utils.h
ppenna, line 1: Changes on this file are not required.

shim/src/queue.c
ppenna, line 1: Changes on this file are not required.

shim/src/interpose.c
ppenna, line 1: Why are these changes required?

shim/src/glue.c
ppenna, line 1: Changes on this file are not required.

stolet: It seems like there are a few conflicting/duplicate changes after the merges of #1357, #1358 and #1359. I separated the remaining change in another PR #1373. Perhaps we can close this one and just merge #1373 to avoid confusion.

PR #1353, opened 2024-08-01
[shim] O3 For Compiling C Code

ppenna:
## Description

This PR fixes the build system to compile all C code with O3 optimization level on release build.


ppenna:
Bypassing review as this PR concerns the build system.

PR #1355, opened 2024-08-02
[inetstack] Enhancement: Remove most clones

iyzhang:
This PR removes unnecessary clones in the inetstack.

### DIFF
--- a/src/rust/inetstack/protocols/icmpv4/datagram/message.rs
+++ b/src/rust/inetstack/protocols/icmpv4/datagram/message.rs
@@ -26,7 +26,7 @@ pub struct Icmpv4Message {
     ethernet2_hdr: Ethernet2Header,
     ipv4_hdr: Ipv4Header,
     icmpv4_hdr: Icmpv4Header,
-    data: DemiBuffer,
+    data: Option<DemiBuffer>,
 }
 
 //======================================================================================================================
@@ -45,7 +45,7 @@ impl Icmpv4Message {
             ethernet2_hdr,
             ipv4_hdr,
             icmpv4_hdr,
-            data,
+            data: Some(data),
         }
     }
 }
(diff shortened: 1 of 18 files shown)


ppenna (approved):
Thanks for exposing hidden closing semantics around `take_body()`

src/rust/inetstack/protocols/icmpv4/datagram/message.rs
```diff
@@ -26,7 +26,7 @@ pub struct Icmpv4Message {
     ethernet2_hdr: Ethernet2Header,
     ipv4_hdr: Ipv4Header,
     icmpv4_hdr: Icmpv4Header,
-    data: DemiBuffer,
+    data: Option<DemiBuffer>,
```
anandbonde, line 29: If the code panics when `data` is `None`, then why make it an `Option`?

src/rust/inetstack/protocols/udp/datagram/mod.rs
```diff
@@ -38,7 +38,7 @@ pub struct UdpDatagram {
     /// UDP header.
     udp_hdr: UdpHeader,
     /// Payload
-    data: DemiBuffer,
+    data: Option<DemiBuffer>,
```
anandbonde, line 41: If the code panics when `data` is `None`, then why make it an `Option`?

src/rust/inetstack/protocols/icmpv4/datagram/message.rs
iyzhang: So I made it an option because we need to eventually transfer ownership out of the PacketBuf structure. The panic ensures that we don't mistake this PacketBuf for one without a payload at all. It should be unreachable, so maybe I should use that instead?

src/rust/inetstack/protocols/udp/datagram/mod.rs
anandbonde: We discussed offline. The PacketBuf may be valid when `new` is called, but will be `None` after `take_body` is called. So these checks are necessary.

anandbonde (approved):
Looks good.

PR #1356, opened 2024-08-02
[shim] Race Condition In Initialization

ppenna:
## Description

This PR fixes a race condition in the initialization of the SHIM library.

Before we were:
- Erroneous using atomic adds/subs for enclosing the critical session
- Not protecting initialization variables.
- Issuing memory barriers at wrong places.


stolet (approved):
Looks good!

PR #1361, opened 2024-08-06
[scripts] Typos in Config Script

ppenna:
## Description

This PR fixes typos in the config script.


ppenna:
Bypassing review as this PR concerns scripts.

PR #1363, opened 2024-08-06
[runtime] Add return values to transmit and receive in the network runtime

iyzhang:
This PR adds return values to the transmit and receive functions in the network runtime trait, so that we can remove panics and silent errors to return a value.

### DIFF
--- a/src/rust/catnip/runtime/mod.rs
+++ b/src/rust/catnip/runtime/mod.rs
@@ -400,7 +400,7 @@ impl NetworkRuntime for SharedDPDKRuntime {
         })))
     }
 
-    fn transmit<P: PacketBuf>(&mut self, mut pkt: P) {
+    fn transmit<P: PacketBuf>(&mut self, mut pkt: P) -> Result<(), Fail> {
         timer!("catnip::runtime::transmit");
 
         // TODO: Consider an important optimization here: If there is data in this packet (i.e. not just headers), and
@@ -421,10 +421,7 @@ impl NetworkRuntime for SharedDPDKRuntime {
         // Chain body buffer.
 
         // First, allocate a header mbuf and write the header into it.
-        let mut header_mbuf: DemiBuffer = match self.mm.alloc_header_mbuf() {
-            Ok(mbuf) => mbuf,
-            Err(e) => panic!("failed to allocate header mbuf: {:?}", e.cause),
-        };
+        let mut header_mbuf: DemiBuffer = self.mm.alloc_header_mbuf()?;
         let header_size = pkt.header_size();
         assert!(header_size <= header_mbuf.len());
         pkt.write_header(&mut header_mbuf[..header_size]);
@@ -438,7 +435,7 @@ impl NetworkRuntime for SharedDPDKRuntime {
                 assert!(header_size + body.len() >= MIN_PAYLOAD_SIZE);
 
                 // We're only using the header_mbuf for, well, the header.
-                header_mbuf.trim(header_mbuf.len() - header_size).unwrap();
+                header_mbuf.trim(header_mbuf.len() - header_size)?;
 
                 // Get the body mbuf.
                 let body_mbuf: *mut rte_mbuf = if body.is_dpdk_allocated() {
@@ -446,14 +443,14 @@ impl NetworkRuntime for SharedDPDKRuntime {
                     expect_some!(body.into_mbuf(), "'body' should be DPDK-allocated")
                 } else {
                     // The body is not dpdk-allocated, allocate a DPDKBuffer and copy the body into it.
-                    let mut mbuf: DemiBuffer = match self.mm.alloc_body_mbuf() {
-                        Ok(mbuf) => mbuf,
-                        Err(e) => panic!("failed to allocate body mbuf: {:?}", e.cause),
-                    };
+                    let mut mbuf: DemiBuffer = self.mm.alloc_body_mbuf()?;
                     assert!(mbuf.len() >= body.len());
                     mbuf[..body.len()].copy_from_slice(&body[..]);
-                    mbuf.trim(mbuf.len() - body.len()).unwrap();
-                    expect_some!(mbuf.into_mbuf(), "mbuf should not be empty")
+                    mbuf.trim(mbuf.len() - body.len())?;
+                    match mbuf.into_mbuf() {
+                        Some(body) => body,
+                        _ => return Err(Fail::new(libc::ENOMEM, "body mbuf could not be allocated")),
+                    }
                 };
 
                 let mut header_mbuf_ptr: *mut rte_mbuf =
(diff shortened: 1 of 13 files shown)


anandbonde (approved):
Looks good.

src/rust/catnip/runtime/mod.rs
```diff
@@ -438,22 +435,25 @@ impl NetworkRuntime for SharedDPDKRuntime {
...
-                    };
+                    let mut mbuf: DemiBuffer = self.mm.alloc_body_mbuf()?;
                     assert!(mbuf.len() >= body.len());
                     mbuf[..body.len()].copy_from_slice(&body[..]);
-                    mbuf.trim(mbuf.len() - body.len()).unwrap();
-                    expect_some!(mbuf.into_mbuf(), "mbuf should not be empty")
+                    mbuf.trim(mbuf.len() - body.len())?;
+                    match mbuf.into_mbuf() {
+                        Some(body) => body,
+                        _ => return Err(Fail::new(libc::EINVAL, "body mbuf could not be allocated")),
```
ppenna, line 455: This error happened not because an invalid argument was supplied to the function, but rather because we could not allocate a buffers to hold the body of the pakcet. My suggestion is that we should return either  `EAGAIN` or `ENOMEM` instead.

src/rust/catnip/runtime/mod.rs
```diff
@@ -438,22 +435,25 @@ impl NetworkRuntime for SharedDPDKRuntime {
...
+                header_mbuf.trim(header_mbuf.len() - header_size)?;
 
                 // Get the body mbuf.
                 let body_mbuf: *mut rte_mbuf = if body.is_dpdk_allocated() {
                     // The body is already stored in an MBuf, just extract it from the DemiBuffer.
-                    expect_some!(body.into_mbuf(), "'body' should be DPDK-allocated")
+                    match body.into_mbuf() {
+                        Some(body) => body,
+                        _ => return Err(Fail::new(libc::EINVAL, "'body' should be DPDK-allocated")),
+                    }
```
ppenna, line 446: In my opinion, this should really be an `expect_some!()`. If a wrongly allocated buffer made all its way down here, something really bad happend.

src/rust/catpowder/win/runtime.rs
```diff
@@ -81,21 +81,23 @@ impl NetworkRuntime for SharedCatpowderRuntime {
...
         }
 
         let mut idx: u32 = 0;
 
         if self.0.borrow_mut().tx.reserve_tx(Self::RING_LENGTH, &mut idx) != Self::RING_LENGTH {
-            warn!("failed to reserve producer space for packet");
-            return;
+            let cause = format!("failed to reserve producer space for packet");
+            warn!("{}", cause);
+            return Err(Fail::new(libc::EINVAL, &cause));
```
ppenna, line 100: We failed not becase an invalid argument was supplied, but rather we experienced a runtime error. In my opinion, `EAGAIN` would better reflect that condition.

src/rust/catpowder/win/runtime.rs
```diff
@@ -113,8 +115,9 @@ impl NetworkRuntime for SharedCatpowderRuntime {
             xdp_rs::_XSK_NOTIFY_FLAGS_XSK_NOTIFY_FLAG_POKE_TX | xdp_rs::_XSK_NOTIFY_FLAGS_XSK_NOTIFY_FLAG_WAIT_TX;
 
         if let Err(e) = self.0.borrow_mut().notify_socket(flags, u32::MAX, &mut outflags) {
-            warn!("failed to notify socket: {:?}", e);
-            return;
+            let cause = format!("failed to notify socket: {:?}", e);
+            warn!("{}", cause);
+            return Err(Fail::new(libc::EINVAL, &cause));
```
ppenna, line 120: We failed because due to a runtime error. In my opinion, `EAGAIN` would better reflect this condition.

src/rust/catpowder/win/runtime.rs
```diff
@@ -124,23 +127,25 @@ impl NetworkRuntime for SharedCatpowderRuntime {
             .reserve_tx_completion(Self::RING_LENGTH, &mut idx)
             != Self::RING_LENGTH
         {
-            warn!("failed to send packet");
-            return;
+            let cause = format!("failed to send packet");
+            warn!("{}", cause);
+            return Err(Fail::new(libc::EINVAL, &cause));
```
ppenna, line 132: We failed due to a runtime condition. In my opinion, `EGAIN` would better reflect this condition.

src/rust/inetstack/protocols/icmpv4/peer.rs
```diff
@@ -171,12 +171,14 @@ impl<N: NetworkRuntime> SharedIcmpv4Peer<N> {
...
             let local_ipv4_addr: Ipv4Addr = self.local_ipv4_addr;
-            self.transport.transmit(Icmpv4Message::new(
+            if let Err(e) = self.transport.transmit(Icmpv4Message::new(
                 Ethernet2Header::new(dst_link_addr, local_link_addr, EtherType2::Ipv4),
                 Ipv4Header::new(local_ipv4_addr, dst_ipv4_addr, IpProtocol::ICMPv4),
                 Icmpv4Header::new(Icmpv4Type2::EchoReply { id, seq_num }, 0),
                 data,
-            ));
+            )) {
+                warn!("COuld not receive send packet: {:?}", e);
```
ppenna, line 180: This may happen due to some unrecoverable error, such as data is incorrect, or because we had a runtime problem, such as we could not allocate enough resources to carry the operation. Normally, an OS would distinguish between the two and handle the latter by retrying a fixed amount of times before failing.

src/rust/inetstack/protocols/icmpv4/peer.rs
```diff
@@ -232,7 +234,7 @@ impl<N: NetworkRuntime> SharedIcmpv4Peer<N> {
             Icmpv4Header::new(echo_request, 0),
             data,
         );
-        self.transport.transmit(msg);
+        self.transport.transmit(msg)?;
```
ppenna, line 237: This may happen due to some unrecoverable error, such as data is incorrect, or because we had a runtime problem, such as we could not allocate enough resources to carry the operation. Normally, an OS would distinguish between the two and handle the latter by retrying a fixed amount of times before failing.

tests/rust/common/runtime.rs
```diff
@@ -88,15 +88,21 @@ impl NetworkRuntime for SharedDummyRuntime {
         if let Some(body) = pkt.take_body() {
             buf[header_size..].copy_from_slice(&body[..]);
         }
-        self.outgoing.try_send(buf).unwrap();
+        match self.outgoing.try_send(buf) {
```
ppenna, line 91: This failed due to a runtime error. In my opinion, we should fail with `EAGAIN`.

ppenna (approved):
Thanks for working on this. Overall looks good and we can merge it.

In summary, I left two suggestions:
- Change the error codes that we use to signal runtime errors.
- Retry sending on some places when we hit a runtime error.

src/rust/catnip/runtime/mod.rs
iyzhang: sounds good

src/rust/inetstack/protocols/arp/peer.rs
```diff
@@ -234,7 +234,9 @@ impl<N: NetworkRuntime> SharedArpPeer<N> {
                         ),
                     );
                     debug!("Responding {:?}", reply);
-                    self.network.transmit(reply);
+                    if let Err(e) = self.network.transmit(reply) {
+                        warn!("Could not transmit message: {:?}", e);
+                    }
```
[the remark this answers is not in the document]
iyzhang: it's probably ok to not retry here since the other end will naturally retry as well.

PR #1366, opened 2024-08-07
[catnip] Allow DPDK prepend function

iyzhang:
This PR adds support for the DPDK prepend function in anticipation of using prepend for packet headers instead of allocating a new mbuf.


ppenna (approved):
LGTM.

Thank you submitting this.

PR #1368, opened 2024-08-10
[runtime] Align DPDK and heap buffers for headroom

iyzhang:
This PR fixes a bug introduced in #1367. Now that we also allocate headroom for heap-allocated DemiBuffers, we can (and must) switch to using the same code for dereferencing both DPDK and heap-allocated buffers to account for the offset of the data in the buffer.

### DIFF
--- a/src/rust/runtime/memory/demibuffer.rs
+++ b/src/rust/runtime/memory/demibuffer.rs
@@ -741,20 +741,6 @@ impl DemiBuffer {
         Ok(cloned_buf)
     }
 
-    /// Provides a raw pointer to the buffer data.
-    ///
-    /// The reference count is not affected in any way and the DemiBuffer is not consumed.  The pointer is valid for as
-    /// long as there is a DemiBuffer in existance that is holding a reference on this data.
-    // This function is not marked unsafe, as the unsafe act is dereferencing the returned pointer, not providing it.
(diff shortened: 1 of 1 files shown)


src/rust/runtime/memory/demibuffer.rs
```diff
@@ -747,12 +747,7 @@ impl DemiBuffer {
...
     // This function is not marked unsafe, as the unsafe act is dereferencing the returned pointer, not providing it.
     pub fn as_ptr(&self) -> *const u8 {
-        // TODO: Review having this "match", since MetaData and MBuf are laid out the same, these are equivalent cases.
-        match self.get_tag() {
-            Tag::Heap => self.data_ptr(),
-            #[cfg(feature = "libdpdk")]
-            Tag::Dpdk => self.dpdk_data_ptr(),
-        }
+        self.data_ptr()
     }
```
ppenna, line 751: Since we no longer need the multiplexing logic, we should also move the implementation of `data_ptr()` to `as_ptr()` and get rid of `data_ptr()`.

ppenna (approved):
LGTM

src/rust/runtime/memory/demibuffer.rs
iyzhang: sounds good!

PR #1369, opened 2024-08-11
[inetstack] Support prepend

iyzhang:
This PR implements support for prepend that was implemented in previous PRs #1367 and #1366. It moves the prepending of the headers into the constructor of the PacketBuf and removes the separate header writing functionality in the PacketBuf ahead of removing the PacketBuf structure altogether. As a result, this PR mostly removes functionality. For example, the transmit function in the network runtime simply transmits the entire packet and does not do any unpacking or writing of the packet any more (except for in Catpowder where it needs to look at the IP addresses in the ethernet header).

I will submit another PR later to remove the header mbuf pool in Catnip (trying to keep this PR as small as possible).


src/rust/inetstack/protocols/arp/peer.rs
```diff
@@ -232,7 +232,9 @@ impl<N: NetworkRuntime> SharedArpPeer<N> {
                             header.get_sender_hardware_addr(),
                             header.get_sender_protocol_addr(),
                         ),
-                    );
+                    )
+                    .expect("Could not construct message");
```
ppenna: Instead of panicking, why dont we log an error message and `continue` the loop?

src/rust/catnip/runtime/mod.rs
```diff
@@ -402,103 +400,23 @@ impl NetworkRuntime for SharedDPDKRuntime {
 
     fn transmit<P: PacketBuf>(&mut self, mut pkt: P) -> Result<(), Fail> {
         timer!("catnip::runtime::transmit");
+        // Grab the packet and copy it if necessary. In general, this copy will happen for small packets without ////..
```
ppenna: Extra `////`?

src/rust/catnip/runtime/mod.rs
```diff
@@ -402,103 +400,23 @@ impl NetworkRuntime for SharedDPDKRuntime {
...
+        // payloads because we allocate actual data-carrying application buffers from the DPDK pool.
+        let outgoing_pkt: DemiBuffer = match pkt.take_body().unwrap() {
```
ppenna: This function returns a `Result`. Why should we `unwrap()` or `expect()`, if we can log an error message and fail instead?

src/rust/catpowder/win/runtime.rs
```diff
@@ -82,12 +88,11 @@ impl NetworkRuntime for SharedCatpowderRuntime {
 
     /// Transmits a packet.
     fn transmit<P: PacketBuf>(&mut self, mut pkt: P) -> Result<(), Fail> {
-        let header_size: usize = pkt.header_size();
-        let body_size: usize = pkt.body_size();
-        trace!("transmit(): header_size={:?}, body_size={:?}", header_size, body_size);
-
-        if header_size + body_size >= u16::MAX as usize {
-            let cause = format!("packet is too large: {:?}", header_size + body_size);
+        let outgoing_pkt: DemiBuffer = pkt.take_body().unwrap();
```
ppenna: This function returns a `Result`. Why should we `unwrap()` and `expect()` instead of failing?

src/rust/inetstack/protocols/udp/tests.rs
```diff
@@ -34,6 +41,8 @@ use ::std::{
     },
 };
 
+const HEADER_SIZE: usize = UDP_HEADER_SIZE + IPV4_HEADER_MAX_SIZE as usize + ETHERNET2_HEADER_SIZE;
```
ppenna: Should this be made a constant in mailine code?

src/rust/inetstack/test_helpers/runtime.rs
```diff
@@ -96,20 +101,18 @@ impl NetworkRuntime for SharedTestRuntime {
     }
 
     fn transmit<P: PacketBuf>(&mut self, mut pkt: P) -> Result<(), Fail> {
-        let header_size: usize = pkt.header_size();
-        let body_size: usize = pkt.body_size();
-        debug!("transmit frame: {:?} body: {:?}", self.outgoing.len(), body_size);
+        let outgoing_pkt: DemiBuffer = pkt.take_body().unwrap();
```
ppenna, line 104: We should consider failling here.

src/rust/inetstack/protocols/mod.rs
```diff
@@ -10,12 +14,32 @@ pub mod peer;
...
+// Imports
+//======================================================================================================================
+
 use ::std::slice::ChunksExact;
 
+//======================================================================================================================
+// Constants
+//======================================================================================================================
+
+pub const MAX_HEADER_SIZE: usize =
```
ppenna, line 28: I understand that this should be defined as the MAXIMUM header among all the protocols that are supported. Could you add a comment about that?

ppenna (approved):
LGTM

src/rust/inetstack/protocols/udp/tests.rs
iyzhang: I switched it to MAX_HEADER_SIZE

PR #1372, opened 2024-08-15
[catnip] Enhancement: Add DPDK errnos to error messages

iyzhang:
This PR adds support for grabbing DPDK errands and adding them to the error message that is printed and returned.

### DIFF
--- a/src/rust/catnip/runtime/memory/mempool.rs
+++ b/src/rust/catnip/runtime/memory/mempool.rs
@@ -50,7 +50,10 @@ impl MemoryPool {
 
         // Failed to create memory pool.
         if pool.is_null() {
-            return Err(Fail::new(libc::EAGAIN, "failed to create memory pool"));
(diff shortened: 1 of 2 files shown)


src/rust/catnip/runtime/memory/mempool.rs
```diff
@@ -50,7 +50,10 @@ impl MemoryPool {
 
         // Failed to create memory pool.
         if pool.is_null() {
-            return Err(Fail::new(libc::EAGAIN, "failed to create memory pool"));
+            let rte_errno: libc::c_int = unsafe { dpdk_rs::rte_errno() };
+            let cause: String = format!("failed to create memory pool: {:?}", rte_errno);
+            warn!("{}", cause);
```
ppenna: I would also include the name of the function in the warning messages, because if the same module has same error messages, we cannot distinguish between them.

ppenna (approved):
LGTM

PR #1373, opened 2024-08-19
[shim] Enhancement: Use demi_wait_any instead of demi_wait in shim ep…

stolet:
Reducing the number of demi calls and background tasks run by using demi_wait_any instead of demi_wait.


ppenna (approved):
LGTM

PR #1375, opened 2024-08-20
[testing] Enhancement: Free buffers in tests

### LINKED ISSUE #1340: [inetstack] Ack queue grows out of control and only counts bytes

iyzhang:
This PR adds a free for two places in the system tests where we were leaking memory. Closes #1340 because that was a side effect of this bug.


anandbonde (approved):
Looks good.

PR #1376, opened 2024-08-20
[catnip] Remove header pool

iyzhang:
Since we now prepend our headers, there is no need for a header pool separate from the pool for holding packet bodies. This PR removes them. This PR depends on PR #1375 because otherwise we fail the system tests when running out of memory.


anandbonde (approved):
Looks good.

PR #1377, opened 2024-08-22
[scheduler] Cleaned up unnecessary comments

anandbonde:
* Cleaned up unnecessary comments in the `scheduler` dir
* Uniform length for section headers at all places


iyzhang (approved):
LGTM

PR #1379, opened 2024-08-26
[inetstack] Move NetworkRuntime to Layer 1

iyzhang:
This PR moves the NetworkRuntime out of the runtime module and places it in the correct place as the layer1 API or physical layer abstraction. This is the beginning of a series of PRs to move the inetstack into proper OSI layers without cross layer dependencies.


src/rust/inetstack/protocols/layer1/mod.rs
```diff
@@ -0,0 +1,43 @@
...
+        },
+        network::consts::RECEIVE_BATCH_SIZE,
+    },
+};
+
+//======================================================================================================================
+// Traits
+//======================================================================================================================
+
+/// Network Runtime
```
anandbonde: This can either be removed or renamed.

anandbonde (approved):
Looks good.

ppenna (approved):

PR #1380, opened 2024-08-27
[examples] Cleaned up the rust TCP examples

anandbonde:
Cleaned up rust TCP examples.

### DIFF
--- a/examples/rust/tcp-dump.rs
+++ b/examples/rust/tcp-dump.rs
@@ -45,27 +45,16 @@ pub const AF_INET: i32 = libc::AF_INET;
 #[cfg(target_os = "linux")]
 pub const SOCK_STREAM: i32 = libc::SOCK_STREAM;
 
-//======================================================================================================================
-// Program Arguments
-//======================================================================================================================
-
-/// Program Arguments
 #[derive(Debug)]
 pub struct ProgramArguments {
-    /// Local socket IPv4 address.
-    local: SocketAddr,
+    local_socket_addr: SocketAddr,
 }
 
-/// Associate functions for Program Arguments
 impl ProgramArguments {
-    /// Default local socket IPv4 address.
-    const DEFAULT_LOCAL: &'static str = "127.0.0.1:12345";
+    const DEFAULT_LOCAL_IPV4_ADDR: &'static str = "127.0.0.1:12345";
 
-    /// Parses the program arguments from the command line interface.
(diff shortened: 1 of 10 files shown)


examples/rust/tcp-dump.rs
```diff
@@ -45,27 +45,16 @@ pub const AF_INET: i32 = libc::AF_INET;
...
-//======================================================================================================================
-// Program Arguments
-//======================================================================================================================
-
-/// Program Arguments
 #[derive(Debug)]
 pub struct ProgramArguments {
-    /// Local socket IPv4 address.
-    local: SocketAddr,
+    local_addr: SocketAddr,
```
iyzhang: do you want to label this local_ip_addr? Just to be clear because there are lots of different types of addresses and it's the one that we use in the inet stack.

examples/rust/tcp-push-pop.rs
```diff
@@ -44,25 +46,18 @@ pub const SOCK_STREAM: i32 = libc::SOCK_STREAM;
...
-const ITERATIONS: usize = 10;
+const BUFSIZE_BYTES: usize = 64;
 const FILL_CHAR: u8 = 0x65;
-const DEFAULT_TIMEOUT: Duration = Duration::from_secs(30);
-
-//======================================================================================================================
-// mksga()
-//======================================================================================================================
+const ITERATIONS: usize = 10;
+const TIMEOUT_SECONDS: Duration = Duration::from_secs(30);
```
iyzhang, line 52: This might be a good time to reconsider this time out (or configure our TCP retry timeouts on the CI VMs to be lower than this).

iyzhang (approved):

examples/rust/tcp-dump.rs
anandbonde: Yes, that sounds good. Will updated.

anandbonde: The SocketAddr enum eventually holds the IP Address, the Port and possibly few other things (for IPv6). I can rename all instances in the codebase of this type to `local_socket_addr` / `remote_socket_addr` to reduce ambiguity. What do you think?

examples/rust/tcp-push-pop.rs
anandbonde: We discussed offline, will experiement with varying timeouts at various stages: per test timeout, test invocation timeout on cli etc. Then we will decide the right values for this.

PR #1382, opened 2024-08-27
[config] Added interface_index to sample configs

anandbonde:
Added interface_index to sample configs.

### DIFF
--- a/scripts/config/azure.yaml
+++ b/scripts/config/azure.yaml
@@ -5,7 +5,8 @@ demikernel:
   local_ipv4_addr: XX.XX.XX.XX
   local_link_addr: "ff:ff:ff:ff:ff:ff"
 raw_socket:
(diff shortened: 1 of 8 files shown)


scripts/config/azure.yaml
```diff
@@ -6,6 +6,7 @@ demikernel:
   local_link_addr: "ff:ff:ff:ff:ff:ff"
 raw_socket:
   interface_name: "abcde"
+  interface_index: 0
```
iyzhang: should we annotate this with linux_interface_name and xdp_interface_index?

iyzhang (approved):

scripts/config/azure.yaml
anandbonde: Good idea! Will change this and the consuming code.

PR #1383, opened 2024-08-27
[test] increased timeouts to accommodate for OS timeouts

anandbonde:
Increased timeout with the anticipation that the tests may not fail as often as they do currently. We suspect that pkt loss causes timeouts to trigger on OS side so this may be a very small time to recover from that and hence the tests fail. We will revisit this if required or if tests still keep failing frequently.


iyzhang (changes requested):
If you change this, you also need to change the time out for the scripts that the CI runs from 120 to 240 at least or more

iyzhang (approved):

ppenna (approved):
LGTM

PR #1384, opened 2024-08-27
[examples] Cleaned up the tcp-close example

anandbonde:
Cleaned up the tcp-close example. No logic change, shouldn't break anything.

### DIFF
--- a/examples/tcp-close/client.rs
+++ b/examples/tcp-close/client.rs
@@ -1,13 +1,17 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+// This program is a client that tests the TCP close operation. It can run in two modes: 1) sequential, 2) concurrent.
+// Within each mode, it can run in two variations: 1) client closes the sockets right after connecting, 2) expecting the
+// server to close the sockets.
+
 //======================================================================================================================
 // Imports
 //======================================================================================================================
 
 use crate::{
     helper_functions,
-    DEFAULT_TIMEOUT,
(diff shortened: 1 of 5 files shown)


examples/tcp-close/client.rs
```diff
@@ -166,22 +146,19 @@ impl TcpClient {
...
 
-    /// Attempts to close several connections sequentially with the expectation
-    /// that the server will close sockets.
-    pub fn run_sequential_expecting_server_to_close_sockets(&mut self, nclients: usize) -> Result<()> {
-        for i in 0..nclients {
-            // Connect to the server and wait.
-            let qd: QDesc = self.issue_socket()?;
-            let qt: QToken = self.libos.connect(qd, self.remote)?;
-            let qr: demi_qresult_t = self.libos.wait(qt, Some(DEFAULT_TIMEOUT))?;
+    pub fn run_sequential_expecting_server_to_close_sockets(&mut self, num_clients: usize) -> Result<()> {
```
iyzhang, line 153: Can you leave a comment somewhere (either beginning of the file or here) about what this test is trying to achieve? Similar to what I put at the top of tcp-push-pop. Because it's very unclear from the documentation what sequential vs concurrent means and why we want to have the server close vs the client close (and not test something else instead).

iyzhang (approved):

examples/tcp-close/client.rs
anandbonde: Done, added comments at the top of both client and server programs.

PR #1321, opened 2024-07-01
[inetstack] Enhancement: Split into OSI layers

iyzhang:
This PR organizes the inetstack into the classic OSDI layers between 2 and 4.

PR #1385, opened 2024-08-28
[inetstack] Move ethernet code to layer 2 module

iyzhang:
This PR moves all ethernet implementations to a layer2 module. Currently, the module does not construct the ethernet headers because those are still in the PacketBuf data structure and will be removed in a future PR. It does parse ethernet headers but returns the entire header to layer 3. We may eventually want to just return the various parts of the ethernet header that are needed at the higher layers. 

The primary change in this PR is to remove all knowledge of the physical layer from layers 3 and 4. To remove those generics and encapsulate the physical layer within the layer 2 module, we do the following:

1. Move all functions with generics out of the PhysicalLayer trait. We now only pass DemiBuffers around.
2. Maintain a reference to the physical layer in the test engine so that we can still deliver packets.

This PR is largely similar to #1321 but does only the split for layer 2.


src/rust/demikernel/libos/mod.rs
```diff
@@ -142,28 +139,22 @@ impl LibOS {
             #[cfg(feature = "catpowder-libos")]
             LibOSName::Catpowder => {
                 // TODO: Remove some of these clones once we are done merging the libOSes.
```
anandbonde: Does this refer to the runtime.clone() below? Can that be avoided at all?

src/rust/inetstack/protocols/layer2/mod.rs
```diff
@@ -0,0 +1,149 @@
...
+
+impl DerefMut for SharedLayer2Endpoint {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        self.0.deref_mut()
+    }
+}
+
+/// Memory Runtime Trait Implementation for the network stack.
+impl MemoryRuntime for Layer2Endpoint {
+    /// Casts a [DPDKBuf] into an [demi_sgarray_t].
```
anandbonde: These (and such) comments can be skipped because they are not adding any more context than the code below (a few exceptions are: externally facing APIs, very complex pieces of code etc.). However, in this case, this function delegates the responsibility of casting further and doesn't do it here itself. These comments are susceptible to deviation from the code in future. I understand that we want to produce good rust docs, but we can be selective about that, so whenever we have any comments, they add a meaningful context to be consumed in conjunction with reading the code - otherwise, code is the best documentation. Also, I doubt that anybody would just read rust docs/comments without the need of reading the code in order to understand how the system works.

src/rust/inetstack/protocols/layer2/mod.rs
```diff
@@ -0,0 +1,149 @@
...
+    fn into_sgarray(&self, buf: DemiBuffer) -> Result<demi_sgarray_t, Fail> {
+        self.layer1_endpoint.into_sgarray(buf)
+    }
+
+    /// Allocates a [demi_sgarray_t].
+    fn sgaalloc(&self, size: usize) -> Result<demi_sgarray_t, Fail> {
```
anandbonde: It would be good to annotate the variable name with the unit (e.g., `size_bytes` etc.) to avoid looking up at the flow of this variable in the code.

src/rust/inetstack/protocols/mod.rs
```diff
@@ -27,7 +27,7 @@ use ::std::slice::ChunksExact;
 
 // Maximum header size of all possible headers.
 pub const MAX_HEADER_SIZE: usize =
-    tcp::MAX_TCP_HEADER_SIZE + ipv4::IPV4_HEADER_MAX_SIZE as usize + ethernet2::ETHERNET2_HEADER_SIZE;
+    tcp::MAX_TCP_HEADER_SIZE + ipv4::IPV4_HEADER_MAX_SIZE as usize + layer2::ETHERNET2_HEADER_SIZE;
```
anandbonde, line 30: Encoding the unit of size in these names would be super helpful to the reader.

src/rust/inetstack/protocols/peer.rs
```diff
@@ -132,7 +132,7 @@ impl<N: PhysicalLayer> Peer<N> {
     }
 
     /// Set an SO_* option on the socket.
-    pub fn set_socket_option(&mut self, sd: &mut Socket<N>, option: SocketOption) -> Result<(), Fail> {
+    pub fn set_socket_option(&mut self, sd: &mut Socket, option: SocketOption) -> Result<(), Fail> {
```
anandbonde, line 135: The naming in this file is a bit confusing. `sd` resonates with `qd` but I don't think there is any direct correlation. If not, then it would be good to call it `socket` or `s`.

src/rust/inetstack/protocols/udp/tests.rs
```diff
@@ -106,8 +106,8 @@ fn udp_push_pop() -> Result<()> {
     };
     now += Duration::from_micros(1);
 
-    // Receive data from Bob.
-    carrie.receive(bob.pop_frame()).unwrap();
+    // Receive data from Bob and insert into Carrie's incoming queue.
+    carrie.push_frame(bob.pop_frame());
```
anandbonde, line 110: Why is this comment different from comments at other instances of the same code? Is this a special case?

src/rust/inetstack/test_helpers/engine.rs
```diff
@@ -51,125 +52,127 @@ use ::std::{
...
+
+pub struct SharedEngine(SharedObject<Engine>);
 
 impl SharedEngine {
-    pub fn new(config_path: &str, test_rig: SharedTestRuntime, now: Instant) -> Result<Self, Fail> {
+    pub fn new(config_path: &str, layer1_endpoint: SharedTestPhysicalLayer, now: Instant) -> Result<Self, Fail> {
         let config: Config = Config::new(config_path.to_string())?;
         // Instantiate all of the layers.
-        // Shared Demikernel runtime.
+        // We create a test
```
anandbonde: This seems an incomplete comment.

anandbonde (approved):
Looks good!

src/rust/demikernel/libos/mod.rs
iyzhang: I think I have removed all possible clones. I'm not sure that it is possible to remove the final set of clones as both the inetstack (as the network transport) and the libOS can create coroutines, so need access to the scheduler.

src/rust/inetstack/protocols/layer2/mod.rs
iyzhang: I am only going to change it here. Since there is a trait, we would need to change this everywhere which would be a huge change.

src/rust/inetstack/protocols/mod.rs
iyzhang: I am going to make this change in another PR as well because it would change a bunch of things that are unrelated to this already large PR.

src/rust/inetstack/protocols/peer.rs
iyzhang: Ah, the socket descriptor is anything from a file descriptor to an actual data structure, in this case a Socket struct.

PR #1386, opened 2024-08-29
[inetstack] Move PacketBuf to Layer 2

iyzhang:
This PR moves the PacketBuf data structure to Layer 2 and removes many of the dependencies on abstractions from Layer2 (e.g., local MAC address) into the L2 module. The goal is to keep moving until we get rid of the PacketBuf data structure altogether.


ppenna (approved):
LGTM

src/rust/inetstack/protocols/tcp/established/ctrlblk.rs
```diff
@@ -858,7 +854,11 @@ impl SharedControlBlock {
         };
 
         // Call the runtime to send the segment.
```
anandbonde: This should be adjusted.

src/rust/inetstack/protocols/tcp/passive_open.rs
```diff
@@ -334,7 +330,11 @@ impl SharedPassiveSocket {
         };
 
         // Send it.
```
anandbonde: This can be removed.

src/rust/inetstack/protocols/tcp/tests/simulator.rs
```diff
@@ -850,36 +850,48 @@ impl Simulation {
...
             Some(Self::cook_buffer(tcp_packet.seqnum.win as usize, None))
         } else {
             None
         };
+        let mut segment: TcpSegment =
+            TcpSegment::new(ipv4_hdr, tcp_hdr, data, false).expect("Should be able to create TCP segment");
+        let mut pkt: DemiBuffer = segment.take_body().unwrap();
 
-        TcpSegment::new(ethernet2_hdr, ipv4_hdr, tcp_hdr, data, false).unwrap()
+        // Actually prepend L2 header.
```
anandbonde, line 866: This is identical for tcp and udp, we should extract it into a separate fn if possible.

PR #1388, opened 2024-08-29
[inetstack] Move files to layer 3

iyzhang:
This PR only moves files to a separate layer 3 module. I will submit a subsequent PR to actually remove dependencies and move the PacketBuf up a layer. This PR has no functional changes.


anandbonde (approved):
Looks good.

PR #1389, opened 2024-08-30
[inetstack] Enhancement: Move PacketBuf to L3

iyzhang:
This PR moves the PacketBuf data structure up by one OSI level so that it does not encompass the IP layer any more. Instead, IP and link layer functionality is handled in the appropriate layers, including removing the ARP cache from every other layer except where it belongs.

My general goals here were to: 
1. Minimize changes (there are definitely some things that I could have cleaned up but I wanted to minimize changes as much as possible)
2. Minimize functional changes (I think there are some things that I could have optimized but didn't)
3. Make some variable name changes if those variables are going to remain in the code after I am done


src/rust/inetstack/protocols/tcp/active_open.rs
```diff
@@ -271,32 +256,6 @@ impl SharedActiveOpenSocket {
         // Start connection handshake.
         let handshake_retries: usize = self.tcp_config.get_handshake_retries();
         let handshake_timeout = self.tcp_config.get_handshake_timeout();
-        // Look up remote MAC address.
```
iyzhang, line 274: This code has been moved to a centralized place. I believe it was originally suppose to replicate the retry mechanism within the ARP cache but is actually off by one retry compared to that code.

src/rust/inetstack/protocols/layer2/mod.rs
```diff
@@ -92,7 +91,15 @@ impl SharedLayer2Endpoint {
         Ok(batch)
     }
 
-    pub fn transmit(
+    pub fn transmit_arp_packet(&mut self, remote_link_addr: MacAddress, pkt: DemiBuffer) -> Result<(), Fail> {
```
anandbonde, line 94: Why are these functions needed? It looks like they are just switching the `EtherType2`. If the aim is to bring clarity, another way to think about this is to move the `EtherType2` parameter first in the `transmit()` fn, so it's evident to the reader what's happening. Then, only `transmit()` can be made public.

src/rust/inetstack/protocols/layer3/arp/packet/message.rs
```diff
@@ -33,6 +35,7 @@ pub struct ArpMessage {
 impl ArpMessage {
     /// Creates an ARP message.
     pub fn new(pdu: ArpHeader) -> Result<Self, Fail> {
+        trace!("Creating ARP packet");
```
anandbonde, line 38: Would printing some field of the header help in tracing?

src/rust/inetstack/protocols/layer3/mod.rs
```diff
@@ -10,9 +10,242 @@ pub mod icmpv4;
...
+                    let protocol: IpProtocol = header.get_protocol();
+                    match protocol {
+                        IpProtocol::ICMPv4 => {
+                            self.icmpv4.receive(header, payload);
+                            continue;
+                        },
+                        _ => batch.push((header.get_src_addr(), protocol, payload)),
+                    }
+                },
+                EtherType2::Ipv6 => continue, // Ignore for now.
```
anandbonde: We could possible `warn!("Ipv6 not supported yet");` here?.

src/rust/inetstack/protocols/tcp/peer.rs
```diff
@@ -270,25 +262,25 @@ impl SharedTcpPeer {
...
         debug!("TCP received {:?}", tcp_hdr);
-        let local: SocketAddrV4 = SocketAddrV4::new(ip_hdr.get_dest_addr(), tcp_hdr.dst_port);
-        let remote: SocketAddrV4 = SocketAddrV4::new(ip_hdr.get_src_addr(), tcp_hdr.src_port);
-
-        if remote.ip().is_broadcast() || remote.ip().is_multicast() || remote.ip().is_unspecified() {
-            let cause: String = format!("invalid remote address (remote={})", remote.ip());
-            error!("receive(): {}", &cause);
-            return;
-        }
+        let local: SocketAddrV4 = SocketAddrV4::new(self.local_ipv4_addr, tcp_hdr.dst_port);
```
anandbonde, line 282: Could rename these to `local_socket_addrv4` and `remote_socket_addrv4` becuase there are several addrs in this area of the codebase.

src/rust/inetstack/protocols/tcp/tests/simulator.rs
```diff
@@ -861,11 +861,17 @@ impl Simulation {
...
+            self.local_sockaddr.ip(),
+            tcp_hdr,
+            data,
+            false,
+        )
+        .expect("Should be able to create TCP segment");
         let mut pkt: DemiBuffer = segment.take_body().unwrap();
-
-        // Actually prepend L2 header.
+        let payload_bytes: usize = pkt.len();
```
anandbonde: This should be renamed to `num_payload_bytes` or something similar. For a sec I thought these are the actual bytes. :) There are a few other places where similar code needs to be scanned backwards once it's clear that its the len, and not the actual bytes.

src/rust/inetstack/protocols/tcp/tests/simulator.rs
```diff
@@ -876,21 +882,36 @@ impl Simulation {
...
+            udp_hdr,
+            data,
+            false,
+        )
+        .unwrap();
 
         let mut pkt: DemiBuffer = datagram.take_body().unwrap();
-
-        // Actually prepend L2 header.
+        let payload_bytes: usize = pkt.len();
```
anandbonde: This should be renamed to num_payload_bytes or something similar. For a sec I thought these are the actual bytes. :)

anandbonde (approved):
Looks good.

src/rust/inetstack/protocols/layer2/mod.rs
iyzhang: It is so that I can remove a dependency on EtherType2 from the higher layer. I thought this would be easier to not mess up, then to accidentally send a packet of the wrong type. Let me know what you think.

src/rust/inetstack/protocols/layer3/arp/packet/message.rs
iyzhang: This is going away soon, so I didn't bother.

src/rust/inetstack/protocols/tcp/tests/simulator.rs
iyzhang: ah, good point, will do

src/rust/inetstack/protocols/tcp/tests/simulator.rs
iyzhang: Ah yes, definitely!

src/rust/inetstack/protocols/tcp/peer.rs
iyzhang: I'm going to skip doing this because this code will be moved soon but will note it to fix later

src/rust/inetstack/protocols/layer2/mod.rs
anandbonde: Makes sense, we can keep it the way you have it now.

PR #1390, opened 2024-08-30
[inetstack] Create Layer 4 module

iyzhang:
This PR only creates a Layer 4 module and moves files to it. It has no functional changes, only renaming of modules and headers.


anandbonde (approved):
Looks good.

PR #1393, opened 2024-08-31
[inetstack] Finish layer 4 changes and remove PacketBuf

iyzhang:
This PR finishes the OSI changes with the following:

1. Each layer now only refers to the layer below it. Dependencies between layers are almost all removed
2. Each layer attaches the appropriate layer header on transmit and strips it on receive.
3. The receive path now pulls packets up through the layers, dispatching at each layer as it goes. Only packets that are meaningful to the higher layers are received, the rest are filtered out.
4. The send path no longer constructs a PacketBuf data structure, which required every layer to know about all of the layers below it. Now, headers are prepended as the packet moves through the layers.
5. Removed a copy from the UDP stack where we copied from a slice into a Demibuffer. Now we just strip the headers and directly pass the DemiBuffer back.

We still have some lingering weird dependencies because of tests but this PR largely finishes the reorganization of the inetstack into OSI layers. The transport trait and API terminates at around level 5; i.e., the inetstack implements and maintains ports and session state with an API that matches the one that the other libOSes implement on top of POSIX and Windows APIs. The Demikernel LibOS layer translates Demikernel queue tokens into meaningful session tokens (i.e., SocketDescriptors in the Transport trait). I would consider renaming the trait, except I think no one understands what the session layer means and it does essentially encapsulate the TCP/UDP protocols that are defined as being the basis of the transport layer.


src/rust/inetstack/protocols/layer2/ethernet2/frame.rs
```diff
@@ -34,30 +50,27 @@ impl Ethernet2Header {
         }
     }
 
-    pub fn compute_size(&self) -> usize {
```
iyzhang, line 37: We no longer need this function because we only use the ethernet header size internally now

src/rust/inetstack/protocols/layer3/arp/header.rs
```diff
@@ -121,9 +120,12 @@ impl ArpHeader {
         Ok(pdu)
     }
 
-    /// Serializes the target ARP PDU.
-    pub fn serialize(&self, buf: &mut [u8]) {
-        let buf: &mut [u8; ARP_MESSAGE_SIZE] = (&mut buf[..ARP_MESSAGE_SIZE]).try_into().unwrap();
+    /// Create an ARP message and return the packetized buffer.
+    pub fn serialize_and_attach(&self) -> DemiBuffer {
+        let mut pkt: DemiBuffer = DemiBuffer::new_with_headroom(0, (ARP_MESSAGE_SIZE + ETHERNET2_HEADER_SIZE) as u16);
```
iyzhang, line 127: Since we are creating a brand new packet here, we need to know how much room to allocate for all of its headers, requiring a downward dependency to the ethernet layer.

src/rust/inetstack/protocols/layer3/icmpv4/header.rs
```diff
@@ -70,25 +66,26 @@ impl Icmpv4Header {
...
         let (type_byte, rest_of_header) = self.protocol.serialize();
         buf[0] = type_byte;
         buf[1] = self.code;
         // Skip the checksum for now.
         buf[2] = 0;
         buf[3] = 0;
         buf[4..8].copy_from_slice(&rest_of_header[..]);
-        let checksum: u16 = Self::compute_checksum(buf, data);
+        let (hdr_buf, payload): (&[u8], &[u8]) = buf[..].split_at(ICMPV4_HEADER_SIZE as usize);
+        let checksum: u16 = Self::compute_checksum(hdr_buf.try_into().unwrap(), payload);
```
iyzhang: Not sure if this try here is required.

src/rust/inetstack/protocols/layer3/icmpv4/peer.rs
```diff
@@ -165,22 +174,12 @@ impl SharedIcmpv4Peer {
...
-            ) {
-                Ok(msg) => msg,
-                Err(e) => {
-                    warn!("Could not assemble packet: {:?}", e);
-                    continue;
-                },
-            };
+            let icmp_hdr: Icmpv4Header = Icmpv4Header::new(Icmpv4Type2::EchoReply { id, seq_num }, 0);
+            icmp_hdr.serialize_and_attach(&mut buf);
+            let ipv4_hdr: Ipv4Header = Ipv4Header::new(local_ipv4_addr, dst_ipv4_addr, IpProtocol::ICMPv4);
```
iyzhang, line 179: Not really sure how to get the ICMP packet through the layer 3 IP layer. The OSI model clearly says that it belongs at layer 3 because it doesn't understand ports, but we need to think about how to put it still above the IP layer.

src/rust/inetstack/protocols/layer4/udp/header.rs
```diff
@@ -102,32 +94,27 @@ impl UdpHeader {
...
-
-    /// Parses a buffer into a UDP header.
-    pub fn parse(
-        src_ipv4_addr: &Ipv4Addr,
-        dst_ipv4_addr: &Ipv4Addr,
-        buf: DemiBuffer,
-        checksum_offload: bool,
-    ) -> Result<(Self, DemiBuffer), Fail> {
-        match Self::parse_from_slice(src_ipv4_addr, dst_ipv4_addr, &buf[..], checksum_offload) {
-            Ok((udp_hdr, bytes)) => Ok((udp_hdr, DemiBuffer::from_slice_with_headroom(bytes, MAX_HEADER_SIZE)?)),
```
iyzhang, line 116: Removing a copy here.

src/rust/inetstack/protocols/layer3/arp/header.rs
anandbonde: Do you think adding this as a comment in the code would be better/clearer? And we can remove it once this is no longer true.

src/rust/inetstack/protocols/layer3/icmpv4/header.rs
anandbonde: What would be the alternative? Just use into()? I guess that would be a compile-time check...? That would be better IMO.

src/rust/inetstack/protocols/layer4/mod.rs
```diff
@@ -78,15 +98,41 @@ impl Peer {
...
+            layer3_endpoint,
+        })
     }
 
-    pub fn receive_tcp_packet(&mut self, src_ipv4_addr: Ipv4Addr, pkt: DemiBuffer) {
-        self.tcp.receive(src_ipv4_addr, pkt)
-    }
+    pub fn poll_once(&mut self) {
+        let batch: ArrayVec<(Ipv4Addr, IpProtocol, DemiBuffer), RECEIVE_BATCH_SIZE> = match {
+            timer!("inetstack::layer4_endpoint::poll_once");
```
anandbonde, line 110: Would this essentially only measure layer3's `receive()`? Should we add the timer there then?

src/rust/inetstack/protocols/layer4/mod.rs
```diff
@@ -78,15 +98,41 @@ impl Peer {
...
+            self.layer3_endpoint.receive()
+        } {
+            Ok(batch) => batch,
+            Err(_) => {
+                warn!("Could not receive from network interface, continuing ...");
+                return;
+            },
+        };
+
+        {
```
anandbonde: This place is possibly a good opportunity to create a separate function.

src/rust/inetstack/protocols/layer4/udp/header.rs
anandbonde: This is great!

anandbonde (approved):
Looks good.

src/rust/inetstack/protocols/layer3/icmpv4/header.rs
iyzhang: yes, I think we could use just an into here (which I think would happen automatically)

src/rust/inetstack/protocols/layer4/mod.rs
iyzhang: This measures how long it takes the poll the underlying layer/hardware for incoming packets.

anandbonde: I mean that the current timer will only measure the call to layer3.receive() and everything inside it. There's nothing else happening in the layer4 here. So it would be ok to put it in layer3. But this is ok for now.

PR #1397, opened 2024-09-11
[inetstack] add udp integration tests

anandbonde:
1. Moved some layer 4 udp tests into integration tests.
2. Cleaned up unnecessary comments and renamed for clarity.
3. Renamed tcp-test directory to make it plural.
4. Fixed bug: setsockopt should not happen for UDP sockets.


nettest/input/linux/udp/bind-addr-in-use.pkt
iyzhang, line 1: Can we move this test to the integration tests?

nettest/input/windows/udp/bind-addr-in-use.pkt
iyzhang, line 1: Can we move this test to the integration tests?

src/rust/inetstack/protocols/layer4/udp/tests.rs
```diff
@@ -451,87 +420,6 @@ fn udp_loop2_ping_pong() -> Result<()> {
     Ok(())
 }
 
-//======================================================================================================================
```
iyzhang, line 438: Let's move these three tests to the integration tests. We should create a tests/rust/udp-test directory and move these tests to a bind/mod.rs. Then we will not need a separate directory for windows and linux in nettest, which is not designed to handle error cases.

iyzhang (changes requested):

nettest/input/linux/udp/bind-addr-in-use.pkt
anandbonde: Sure, do you mean, keep them in code instead of .pkt files?

nettest/input/windows/udp/bind-addr-in-use.pkt
anandbonde: Sure, do you mean, keep them in code instead of .pkt files?

src/rust/inetstack/protocols/layer4/udp/tests.rs
anandbonde: Done.

nettest/input/windows/udp/bind-addr-in-use.pkt
anandbonde: Done.

nettest/input/linux/udp/bind-addr-in-use.pkt
anandbonde: Done.

linux.mk
```diff
@@ -214,21 +196,17 @@ export MTU ?= 1500
 export MSS ?= 1500
 export PEER ?= server
 export TEST ?= udp-push-pop
-export TEST_INTEGRATION ?= tcp-test
+export TEST_INTEGRATION ?= tcp-tests
```
iyzhang, line 199: Should this also include udp-tests?

iyzhang (approved):

linux.mk
anandbonde: This is the default value for the integration test. This line changed becuase I changed `tcp-test` to `tcp-tests` (plural). The CI pipeline does run `udp-tests`.

PR #1399, opened 2024-09-17
[demikernel] prepared for publishing to crates.io

anandbonde:
Prepares the dpdk-rs crate for stand-alone publishing to crates.io.


Cargo.toml
```diff
@@ -1,13 +1,9 @@
...
 # Licensed under the MIT license.
 
-#=======================================================================================================================
-# Package
-#=======================================================================================================================
-
 [package]
 name = "demikernel"
-version = "1.5.13"
+version = "0.0.1"
```
iyzhang: We should leave this number as is because it has meaning for our Github repo.

Cargo.toml
```diff
@@ -44,12 +35,15 @@ socket2 = "0.5.7"
...
-nettest = { path = "nettest" }
-dpdk-rs = { path = "dpdk-rs", optional = true }
+# Demikernel crates (published on crates.io).
+# The demikernel-dpdk-rs crate is optional because it is only used for DPDK support.
+demikernel-dpdk-rs = { version = "0.0.1", optional = true }
+demikernel-nettest = { version = "0.0.1" }
 
 # Windows-specific dependencies.
 [target.'cfg(windows)'.dependencies]
+# The demikernel-xdp-rs crate is optional because it is only used for XDP support on Linux.
```
iyzhang: Do you mean Windows?

Cargo.toml
```diff
@@ -44,12 +35,15 @@ socket2 = "0.5.7"
 yaml-rust = "0.4.5"
 x86 = "0.52.0"
 
-# Local crates.
-nettest = { path = "nettest" }
-dpdk-rs = { path = "dpdk-rs", optional = true }
+# Demikernel crates (published on crates.io).
+# The demikernel-dpdk-rs crate is optional because it is only used for DPDK support.
+demikernel-dpdk-rs = { version = "0.0.1", optional = true }
```
iyzhang: is there a reason to prefix with demikernel for crates.io? Is there another crate already named dpdk-rs? Also if we are going to rename the crate, the rs seems kind of redundant.

dpdk-rs/Cargo.toml
```diff
@@ -2,13 +2,14 @@
 # Licensed under the MIT license.
 
 [package]
-name = "dpdk-rs"
-version = "1.1.6"
+name = "demikernel-dpdk-rs"
```
iyzhang: Same here, the number likely has a meaning on Github

iyzhang (approved):

Cargo.toml
anandbonde: I added the `demikernel-` prefix so that all Demikernel crates are easily searchable on https://crates.io/. There are around 160k crates online... and `nettest` already exists. Eventually, I also kept the name in code to match the crate names. We can change the names to remove -rs suffix.

_**Note about crates:** Not a blocker, but something to keep in mind: once a version is published on crates.io, there is no way to go back! It can only be blocked for future uses, but existing uses will continue to work forever. So if we change the name, it will be treated as a completely new crate._

Cargo.toml
anandbonde: Yes. Will change it. 😃

Cargo.toml
anandbonde: Ok we can start from these numbers on crates.io. What are our GitHub version numbers used for?

Cargo.toml
iyzhang: Ok, sounds good!

Cargo.toml
iyzhang: We have tags on Github for different releases so that people can build against a specific release if they want to.

anandbonde: Done.

PR #1404, opened 2024-09-18
[shim] Enhancement: Add send qtoken to epoll

stolet:
Changes the way we interpose send in the shim. Before we did a blocking send, but this caused the latency to spike because the Demikernel only returns from demi_wait on a push operation when it receives back an ACK from the receiver. To improve latency this PR does an async send and adds the qtoken for the send operation to a special epoll table. This table gets checked any time epoll_wait is called.

### DIFF
--- a/shim/src/data/send.c
+++ b/shim/src/data/send.c
@@ -43,10 +43,31 @@ ssize_t __send(int sockfd, const void *buf, size_t len, int flags)
     len = MIN(len, sga.sga_segs[0].sgaseg_len);
     memcpy(sga.sga_segs[0].sgaseg_buf, buf, len);
     assert(__demi_push(&qt, sockfd, &sga) == 0);
-    assert(__demi_wait(&qr, qt, NULL) == 0);
-    assert(qr.qr_opcode == DEMI_OPC_PUSH);
-    __demi_sgafree(&sga);
 
+    // We need to add the send to a special epoll table that gets
+    // polled at epoll_wait. This is because demi_wait
+    // only returns from a push when the demikernel receives an ACK back.
+    // Doing the wait here messes up the behaviour of applications
+    // that expect the send to return immediately after the OS adds
+    // the buffer to the transmit queue.
+    for (int i = 0; i < MAX_EVENTS; i++)
+    {
+        struct demi_event *ev = epoll_get_event(SEND_EPFD, i);
+        if (ev->used == 0)
+        {
+            epoll_init_event(ev, qt, sockfd, &sga);
+
+            struct demi_event *tail = epoll_get_tail(SEND_EPFD);
+            ev->prev_ev = tail->id;
+            tail->next_ev = ev->id;
+            epoll_set_tail(SEND_EPFD, ev->id);
+
+            TRACE("adding send event id=%d qt=%d", ev->id, ev->qt);
+            return len;
+        }
+    }
+
+    UNIMPLEMETED("Deal with case where we run out of send events");
     return (len);
 }
 
--- a/shim/src/epoll/epoll.c
+++ b/shim/src/epoll/epoll.c
@@ -16,7 +16,11 @@ static struct epoll_table epoll_table[EPOLL_MAX_FDS];
 
 int epoll_table_alloc(void)
 {
-    for (int i = 0; i < EPOLL_MAX_FDS; i++)
+    // We skip the first epfd that is used to keep qtokens
+    // for send requests. This function gets called in epoll_create
+    // so we want to avoid this process overriding the values for
+    // the send epoll table.
+    for (int i = SEND_EPFD + 1; i < EPOLL_MAX_FDS; i++)
     {
         if (epoll_table[i].used == 0)
         {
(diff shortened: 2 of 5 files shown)


shim/src/data/send.c
```diff
@@ -43,10 +43,36 @@ ssize_t __send(int sockfd, const void *buf, size_t len, int flags)
...
+    {
+        struct demi_event *ev = epoll_get_event(SEND_EPFD, i);
+        if (ev->used == 0)
+        {
+            ev->used = 1;
+            ev->qt = qt;
+            ev->sockqd = sockfd;
+            ev->next_ev = INVALID_EV;
+            ev->prev_ev = INVALID_EV;
+            ev->sga = sga;
```
ppenna: Do we have this code snippet all around? I feel that we should have an initializer function for that if we are always initializing with some default values.

shim/src/epoll/epoll.c
```diff
@@ -13,10 +14,11 @@ struct epoll_table
 };
 
 static struct epoll_table epoll_table[EPOLL_MAX_FDS];
+static struct epoll_table epoll_table_send;
 
 int epoll_table_alloc(void)
 {
-    for (int i = 0; i < EPOLL_MAX_FDS; i++)
+    for (int i = SEND_EPFD + 1; i < EPOLL_MAX_FDS; i++)
```
ppenna, line 23: Could you add a comment here that explains the rationale for using a non-trivial start value for this for loop?

shim/src/epoll/epoll.c
```diff
@@ -48,12 +50,25 @@ void epoll_table_init(void)
...
+    epoll_table[SEND_EPFD].head = 0;
+    epoll_table[SEND_EPFD].tail = 0;
+    for (int i = 0; i < MAX_EVENTS; i++)
+    {
+        epoll_table[SEND_EPFD].events[i].used = i == 0? 1 : 0;
+        epoll_table[SEND_EPFD].events[i].id = i;
+        epoll_table[SEND_EPFD].events[i].qt = -1;
+        epoll_table[SEND_EPFD].events[i].next_ev = INVALID_EV;
+        epoll_table[SEND_EPFD].events[i].prev_ev = INVALID_EV;
+    }
```
ppenna: This logic is similar to the logic just above. Could we unify that in a single function?

shim/src/epoll/epoll_wait.c
```diff
@@ -103,8 +103,21 @@ int __epoll_wait(int epfd, struct epoll_event *events, int maxevents, int timeou
...
+                        epoll_set_tail(SEND_EPFD, prev->id);
+                    else
+                        next->prev_ev = evs[ready_offset]->prev_ev;
+
+                    evs[ready_offset]->used = 0;
+                    evs[ready_offset]->sockqd = -1;
+                    evs[ready_offset]->qt = (demi_qtoken_t)-1;
+                    evs[ready_offset]->next_ev = INVALID_EV;
+                    evs[ready_offset]->prev_ev = INVALID_EV;
+                    __demi_sgafree(&evs[ready_offset]->sga);
```
ppenna: could we have an assert here?

shim/src/epoll/epoll.c
```diff
@@ -3,6 +3,7 @@
 
 #include "../epoll.h"
 #include <assert.h>
+#include <stdio.h>
```
ppenna: As we don't use `<stdio.h>` in this file, could you remove this include?

shim/src/epoll/epoll.c
```diff
@@ -32,16 +37,19 @@ void epoll_table_init(void)
 {
     for (int i = 0; i < EPOLL_MAX_FDS; i++)
     {
-        epoll_table[i].used = 0;
         epoll_table[i].head = 0;
         epoll_table[i].tail = 0;
+        if (i == SEND_EPFD)
+            epoll_table[i].used = 1;
+        else
+            epoll_table[i].used = 0;
```
ppenna: Could you change for a ternary operator?

shim/src/epoll/epoll.c
```diff
@@ -67,9 +76,44 @@ int epoll_get_ready(int epfd, demi_qtoken_t *qts, struct demi_event **evs)
...
+        if ((send_ev->used) && (send_ev->qt != (demi_qtoken_t)-1))
+        {
+            qts[nevents] = send_ev->qt;
+            evs[nevents] = send_ev;
+            nevents++;
+        }
+
+        send_ev = epoll_get_next(SEND_EPFD, send_ev);
+    }
+
```
ppenna: Could you use a for loop to limit the scope of `send_ev` to the for loop itself?

stolet: Sounds good! I also refactored epoll_get_ready into a more general function, so there is no code replication for the procedure that get the ready events from the regular epoll table and the send events.

ppenna (approved):
LGTM

shim/src/data/send.c
```diff
@@ -43,10 +43,31 @@ ssize_t __send(int sockfd, const void *buf, size_t len, int flags)
     len = MIN(len, sga.sga_segs[0].sgaseg_len);
     memcpy(sga.sga_segs[0].sgaseg_buf, buf, len);
     assert(__demi_push(&qt, sockfd, &sga) == 0);
-    assert(__demi_wait(&qr, qt, NULL) == 0);
-    assert(qr.qr_opcode == DEMI_OPC_PUSH);
-    __demi_sgafree(&sga);
 
+    // We need to add the send to a special epoll table that gets
+    // polled at epoll_wait. This is because demi_wait
+    // only returns from a pop when the demikernel receives an ACK back.
```
anandbonde: Do you mean `push` here?

anandbonde (approved):

shim/src/data/send.c
stolet: You are correct, I will fix that.

PR #1406, opened 2024-09-24
[demikernel] Got rid of 2 unstable rustfmt directives and ran fmt on the codebase

anandbonde:
* Removed 2 unstable rustfmt directives
* Ran fmt on the whole codebase


iyzhang (approved):
Thanks for doing this!

PR #1407, opened 2024-09-25
[demikernel] added cargo fmt check pre-commit hook

anandbonde:
With this change, rustfmt will run automatically on each local commit. It will report all formatting errors and will not allow commits in case of any errors.


iyzhang (approved):
Thanks for doing this!

PR #1408, opened 2024-09-25
[demikernel] updated nightly build version and 3rd party crates

anandbonde:
Updated nightly build and 3rd party crates


iyzhang (approved):
LGTM

PR #1411, opened 2024-09-29
[catnip] Enhancement: removed unnecessary fields

anandbonde:
removed unnecessary fields


iyzhang (approved):
This looks fine but it's possible at some point we might want at least check the Mac address against the one in the config file.

anandbonde:
> This looks fine but it's possible at some point we might want at least check the Mac address against the one in the config file.

Filed issue: https://github.com/microsoft/demikernel/issues/1413

PR #1412, opened 2024-09-30
[inetstack] Enhancement: Rename established socket variables

iyzhang:
This is the first of a number of PRs to clean up the TCP stack, especially around established sockets and the TCP data path. This initial PR simply renames some variables for clarity and units. It makes no changes to the code itself.


anandbonde (commented):
Looks good.

anandbonde (approved):

PR #1414, opened 2024-09-30
[demikernel] Enhancement: remove unnecessary conditional compilation

anandbonde:
Now that we have removed catloop and catmem, the mainstream liboses don't need code to be conditionally compiled.


iyzhang (approved):
LGTM

PR #1415, opened 2024-09-30
[inetstack] Enhancement: Collapse some background coroutines

iyzhang:
This PR accomplishes two primary tasks. First, it removes `unsent_seq_no` and instead uses `unsent_queue` to track whether there is more unsent data. Next, it moves the background send and retransmit coroutines to the Sender data structure. While this is likely not where the code will land, it sets up the next PR which will deduplicate a lot of the send logic that tests for windows sizes and unsent and unacked segments. Finally, by moving the background send and retransmit coroutines to the Sender data structure, we can remove quite a few accessor functions from the ControlBlock. Eventually, I hope to keep all of the TCP state in the ControlBlock and the various queues where they are used.


src/rust/collections/async_queue.rs
```diff
@@ -101,6 +101,16 @@ impl<T> AsyncQueue<T> {
     pub fn get_mut_values(&mut self) -> IterMut<T> {
         self.queue.iter_mut()
     }
+
+    /// Get refernce to first item.
+    pub fn get_front(&self) -> Option<&T> {
```
anandbonde, line 106: Where are these fucntions used? I tried searching using intellisense, but coudln't find any usages. Or are they intended for future use?

anandbonde (approved):
Looks good.

src/rust/collections/async_queue.rs
iyzhang: yes, these are for future use. There are some number of functions for future use in that data structure but that one in particular, I will use in the next PR.

PR #1418, opened 2024-10-01
[network-simulator] Enhancement: cleanup the simulator

anandbonde:
Cleaned up the simulator. Eventually, we would like to move this component into the network-simulator crate because it belongs there.

### DIFF
--- a/src/rust/inetstack/protocols/layer4/tcp/tests/simulator.rs
+++ b/src/rust/inetstack/protocols/layer4/tcp/tests/simulator.rs
@@ -1,11 +1,14 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+// TODO: Move this component to the `network-simulator` crate.
+
 //======================================================================================================================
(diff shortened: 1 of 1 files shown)


src/rust/inetstack/protocols/layer4/tcp/tests/simulator.rs
```diff
@@ -1,11 +1,14 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+// Todo: Move this component to the `network-simulator` crate.
```
iyzhang: can you make this TODO: so it will be flagged by our vscode extension?

iyzhang (approved):
Thanks for doing this! I assume that there are no functional changes.

src/rust/inetstack/protocols/layer4/tcp/tests/simulator.rs
anandbonde: Done

PR #1419, opened 2024-10-01
[inetstack] Dedup some code on the send path

### LINKED ISSUE #198: [inetstack] Send Side Buffer Handling Issues

iyzhang:
This PR closes #198. At this point, we do not need to support attaching FIN to the last data packet.

### DIFF
--- a/src/rust/inetstack/protocols/layer4/mod.rs
+++ b/src/rust/inetstack/protocols/layer4/mod.rs
@@ -305,10 +305,6 @@ impl Peer {
 
 #[cfg(test)]
 impl Peer {
-    pub fn tcp_mss(&self, socket: &SharedTcpSocket) -> Result<usize, Fail> {
-        socket.remote_mss()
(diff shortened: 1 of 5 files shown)


anandbonde (approved):
Looks good.

PR #1420, opened 2024-10-02
[catpowder] Bug Fix: fix compiler warnings for unused code

anandbonde:
Cleaned up some unnecessary comments and removed unused code. Also turned on warnings as errors now that we have cleaned up all of our warnings.


iyzhang (approved):
Looks good to me. Can you leave a note that we've turned on warnings as errors in the PR?

PR #1421, opened 2024-10-03
[inetstack] Enhancement: Move rto and retransmit into Sender

iyzhang:
This PR moves the RTO calculator and retransmit timer out of the ctrl block and into the sender. The sender now updates these values as part of processing an incoming ack, instead of the ctrl block managing them.

### DIFF
--- a/src/rust/inetstack/protocols/layer4/tcp/established/ctrlblk.rs
+++ b/src/rust/inetstack/protocols/layer4/tcp/established/ctrlblk.rs
@@ -17,7 +17,6 @@ use crate::{
             constants::MSL,
             established::{
                 congestion_control::{self, CongestionControlConstructor},
-                rto::RtoCalculator,
                 sender::Sender,
             },
             header::TcpHeader,
@@ -173,15 +172,6 @@ pub struct ControlBlock {
     // Previous send variables and queues.
     // TODO: Consider incorporating this directly into ControlBlock.
     sender: Sender,
-
-    // Send timers
-    // Current retransmission timer expiration time.
-    // TODO: Consider storing this directly in the RtoCalculator.
-    send_retransmit_deadline_time_secs: SharedAsyncValue<Option<Instant>>,
-
-    // Retransmission Timeout (RTO) calculator.
-    send_rto_calculator: RtoCalculator,
-
     // Receive Sequence Variables from RFC 793.
 
     // RCV.NXT - receive next
@@ -290,8 +280,6 @@ impl SharedControlBlock {
                 sender_initial_seq_no,
                 congestion_control_options,
             ),
-            send_retransmit_deadline_time_secs: SharedAsyncValue::new(None),
-            send_rto_calculator: RtoCalculator::new(),
             recv_queue,
             receive_ack_queue_frame_bytes,
             parent_passive_socket_close_queue,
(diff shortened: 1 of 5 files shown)


src/rust/inetstack/protocols/layer4/tcp/established/ctrlblk.rs
```diff
@@ -661,47 +625,27 @@ impl SharedControlBlock {
...
+            self.receive_ack_queue_frame_bytes.push(nbytes)
         } else {
-            // Duplicate ACK (doesn't acknowledge anything new).  We can mostly ignore this, except for fast-retransmit.
-            // TODO: Implement fast-retransmit.  In which case, we'd increment our dup-ack counter here.
-            warn!("process_ack(): received duplicate ack ({:?})", header.ack_num);
+            // This segment acknowledges data we have yet to send!?  Send an ACK and drop the segment.
+            // TODO: See RFC 5961, this could be a Blind Data Injection Attack.
+            let cause: String = format!("Received segment acknowledging data we have yet to send!");
+            warn!("process_ack(): {}", cause);
+            self.send_ack();
```
anandbonde, line 645: Should this code send an ack if this potentially and attack?

src/rust/inetstack/protocols/layer4/tcp/established/sender.rs
```diff
@@ -393,68 +402,96 @@ impl Sender {
...
-            if let Some(mut segment) = self.unacked_queue.try_pop() {
-                // Add sample for RTO if we have an initial transmit time.
-                // Note that in the case of repacketization, an ack for the first byte is enough for the time sample.
-                // TODO: TCP timestamp support.
-                if let Some(initial_tx) = segment.initial_tx {
-                    cb.rto_add_sample(now - initial_tx);
-                }
+    // Process an ack.
+    pub fn process_ack(&mut self, header: &TcpHeader, now: Instant) {
+        // Process the ACK.
```
anandbonde: These comments: ("Process the ack") can be removed because they don't add any context.

src/rust/inetstack/protocols/layer4/tcp/established/sender.rs
```diff
@@ -393,68 +402,96 @@ impl Sender {
...
+        // Start by checking that the ACK acknowledges something new.
+        // TODO: Look into removing Watched types.
+        //
```
anandbonde: A comment with a blank line can be removed.

src/rust/inetstack/protocols/layer4/tcp/established/sender.rs
```diff
@@ -393,68 +402,96 @@ impl Sender {
...
-
-                    // Leave this segment on the unacknowledged queue.
-                    self.unacked_queue.push_front(segment);
-                    break;
-                }
+        if send_unacknowledged < header.ack_num {
+            // Remove the now acknowledged data from the unacknowledged queue, update the acked sequence number
+            // and update the sender window.
+
+            let bytes_acknowledged: u32 = (header.ack_num - self.send_unacked.get()).into();
```
anandbonde, line 417: `bytes_acknowledged` could directly be of the type `usize` instead, to avoid typecasting it on the next line.

src/rust/inetstack/protocols/layer4/tcp/established/sender.rs
```diff
@@ -393,68 +402,96 @@ impl Sender {
...
+            let mut bytes_remaining: usize = bytes_acknowledged as usize;
+            // Remove bytes from the unacked queue.
+            while bytes_remaining != 0 {
+                if let Some(mut segment) = self.unacked_queue.try_pop() {
+                    // Add sample for RTO if we have an initial transmit time.
```
anandbonde, line 423: We can probably enhance this comment to explain why this is done.

src/rust/inetstack/protocols/layer4/tcp/established/sender.rs
```diff
@@ -393,68 +402,96 @@ impl Sender {
...
-            } else {
-                debug_assert!(false); // Shouldn't have bytes_remaining with no segments remaining in unacked_queue.
+            // Update SND.UNA to SEG.ACK.
+            self.send_unacked.set(header.ack_num);
+
+            // Update send window.
+            // Check that the ACK we're using to update the window isn't older than the last one used to update it.
+            if self.send_window_last_update_seq < header.seq_num
+                || (self.send_window_last_update_seq == header.seq_num
+                    && self.send_window_last_update_ack <= header.ack_num)
```
anandbonde: It would be good to extract this condition into a separate function or variable with a meaningful name.

src/rust/inetstack/protocols/layer4/tcp/established/sender.rs
```diff
@@ -393,68 +402,96 @@ impl Sender {
...
+            {
+                // Update our send window.
```
anandbonde: This comment is redundant.

src/rust/inetstack/protocols/layer4/tcp/established/ctrlblk.rs
iyzhang: Yes, this is what the RFC says that we must do. I thought it was weird too so I actually looked up [RFC 5961](https://datatracker.ietf.org/doc/html/rfc5961)

src/rust/inetstack/protocols/layer4/tcp/established/sender.rs
iyzhang: This needs to be a u32 because we can only convert SeqNumber into u32

anandbonde (approved):

PR #1422, opened 2024-10-03
[pal] Enhancement: flattened unnecessary layering in PAL and added mem unit for clarity

anandbonde:
1. There were several layers/indirections in the Platform Abstraction Layer (PAL), which is now condensed or flattened into a single layer. This will make it easy to find, use and maintain the PAL going forward.

2. This will reduce ambiguity to the reader by making it evident in the constant name what the unit of memory is (bytes).


src/rust/catpowder/linux/rawsocket/rawsockaddr.rs
```diff
@@ -5,25 +5,14 @@
...
 #[derive(Clone, Copy)]
-
-/// Raw socket address.
 pub struct RawSocketAddr(libc::sockaddr_ll);
 
-//======================================================================================================================
-// Associate Functions
-//======================================================================================================================
-
-/// Associated functions for raw socket addresses.
```
iyzhang, line 24: Is there a reason to delete this header?

iyzhang (approved):

src/rust/catpowder/linux/rawsocket/rawsockaddr.rs
anandbonde: Reverted.

PR #1424, opened 2024-10-03
[doc] remove branching from CONTRIBUTING.md and add CI info

anandbonde:
We no longer run CI pipelines for all branches based on patterns. Now the CI runs only for PRs and pushes on 3 special branches (main, unstable, dev).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,19 +22,12 @@ From a practical point of view, you should use these branches as follows:
 
 - Use `dev` whenever you want to introduce changes to this project.
 
-Naming Branches
-----------------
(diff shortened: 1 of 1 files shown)


iyzhang (changes requested):
We might still want some organization for branches. I think this is good practice in case others want to contribute.

anandbonde:
> We might still want some organization for branches. I think this is good practice in case others want to contribute.

This was already done quite some time back, I just updated the readme now. Do you suggest forcing format constraints on branch names? The format specification earlier only triggered CI for branches with particular prefixes, but still allowed any branch name to be created otherwise.

iyzhang (approved):

PR #1426, opened 2024-10-04
ci: relax commit message format restrictions

anandbonde:
There were unnecessary restrictions on the commit message format which were never used.


anandbonde:
We discussed offline that it's better to leave these in place for now.

PR #1428, opened 2024-10-04
[inetstack] Bug Fix: Correct retransmission timer

### LINKED ISSUE #202: [inetstack] Retransmit Timer Issues

iyzhang:
This PR closes #202. I have checked on all other cases and they seem to be correct. With this fix, we should be transmitting at the right times for retransmit.

### DIFF
--- a/src/rust/inetstack/protocols/layer4/tcp/established/sender.rs
+++ b/src/rust/inetstack/protocols/layer4/tcp/established/sender.rs
@@ -407,7 +407,6 @@ impl Sender {
         // Start by checking that the ACK acknowledges something new.
         // TODO: Look into removing Watched types.
         let send_unacknowledged: SeqNumber = self.send_unacked.get();
-        let send_next: SeqNumber = self.send_next.get();
 
         if send_unacknowledged < header.ack_num {
             // Remove the now acknowledged data from the unacknowledged queue, update the acked sequence number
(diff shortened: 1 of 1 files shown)


src/rust/inetstack/protocols/layer4/tcp/established/sender.rs
```diff
@@ -459,18 +459,23 @@ impl Sender {
...
-            } else {
-                // Update the retransmit timer.  Some of our outstanding data is now acknowledged, but not all.
-                // TODO: This looks wrong.  We should reset the retransmit timer to match the deadline for the
-                // oldest still-outstanding data.  The below is overly generous (minor efficiency issue).
-                let deadline: Instant = now + self.rto_calculator.rto();
-                self.retransmit_deadline_time_secs.set(Some(deadline));
-            }
+            // Reset the retransmit timer if necessary. If there is more data that hasn't been acked, then set to the
+            // next segment deadline, otherwise, do not set.
+            let retransmit_deadline_time_secs: Option<Instant> = match self.unacked_queue.get_front() {
```
anandbonde: The logic looks good. This is a good candidate for extracting into a separate function as it represents a nice logically independent block.

anandbonde (approved):

PR #1430, opened 2024-10-07
[testing] Add support for multiple ops to simulator and add new test

### LINKED ISSUE #1427: [simulator] Support multiple inflight requests
## Context

Currently, the simulator does not support multiple in flight requests. This leaves us with some cases that we cannot support.

## Proposed Solution

Support queue tokens so that we can wait on separate I/O requests.

## Alternative Solutions

If we always waited on requests in FIFO order, we could use that assumption in the simulator.

iyzhang:
This PR adds:
1. Support for multiple ops in the simulator. It assumes that the waits are in the order that the ops are issued. Closes #1427 
2. A new test for PR #1428 and issue #202

### DIFF
--- a/network_simulator/input/tcp/push/push-retransmission-2.pkt
+++ b/network_simulator/input/tcp/push/push-retransmission-2.pkt
@@ -1,0 +1,48 @@
+// Test for blocking push with 2 lost packets and retransmission.
+
+// Accept a connection.
++.0 socket(..., SOCK_STREAM, IPPROTO_TCP) = 500
++.0 bind(500, ..., ...) = 0
++.0 listen(500, 1) = 0
++.2 accept(500, ..., ...) = 0
+
+// Receive SYN packet.
++.2 TCP < S seq 0(0) win 65535 <mss 1450,wscale 0>
+// Send SYN-ACK packet.
++.0 TCP > S. seq 0(0) ack 1 win 65535 <mss 1450,wscale 0>
+// Receive ACK on SYN-ACK packet.
++.2 TCP < . seq 1(0) ack 1 win 65535 <nop>
+
+// Succeed to accept connection.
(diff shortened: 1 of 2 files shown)


src/rust/inetstack/protocols/layer4/tcp/established/sender.rs
```diff
@@ -459,26 +458,36 @@ impl Sender {
...
-            } else {
-                // Update the retransmit timer.  Some of our outstanding data is now acknowledged, but not all.
-                // TODO: This looks wrong.  We should reset the retransmit timer to match the deadline for the
-                // oldest still-outstanding data.  The below is overly generous (minor efficiency issue).
-                let deadline: Instant = now + self.rto_calculator.rto();
-                self.retransmit_deadline_time_secs.set(Some(deadline));
+            // Reset the retransmit timer if necessary. If there is more data that hasn't been acked, then set to the
+            // next segment deadline, otherwise, do not set.
+            let retransmit_deadline_time_secs: Option<Instant> = self.update_retransmit_deadline(now);
+            #[cfg(debug_assertions)]
```
anandbonde: The macro already tests for this, so this can be removed. Unless we want to conditionally compile out any trace of this macro in the codebase when built in the `release` mode. If so, then we should do the same at all references of this macro. I believe the rust compiler is smart enough to remove unnecessary blocks anyways, so it's probably ok to omit this check from everywhere else too.

Here's the macro def.:

image

iyzhang: Ah ok!

anandbonde: We discussed offline, this will be fixed in subsequent PR.

anandbonde (approved):

PR #1432, opened 2024-10-08
[testing] Enhancement: Add retransmission on close test

iyzhang:
This PR enables writing to connected sockets on the simulator and then retransmission after close.

### DIFF
--- a/src/rust/inetstack/protocols/layer4/tcp/tests/simulator.rs
+++ b/src/rust/inetstack/protocols/layer4/tcp/tests/simulator.rs
@@ -502,15 +502,15 @@ impl Simulation {
             },
         };
 
-        let remote_qd: QDesc = match self.remote_qd {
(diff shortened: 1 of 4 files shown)


src/rust/inetstack/protocols/layer4/tcp/tests/simulator.rs
```diff
@@ -502,15 +502,15 @@ impl Simulation {
             },
         };
 
-        let remote_qd: QDesc = match self.remote_qd {
-            Some((_, qd)) => qd.unwrap(),
+        let qd: QDesc = match args.qd {
```
anandbonde: In the rest of the file, the variable that holds `args.qd` (`qd` in this case) is referred to either `local_qd` or `remote_qd`. It would be good to stick with that naming for clarity.

anandbonde (approved):

PR #1434, opened 2024-10-08
[catpowder] Enhancement: bind to all available receive queues in XDP

### LINKED ISSUE #1429: [Catpowder] Support multiple XDP RSS queues
## Context

Currently we only support a single RSS queue for the XDP physical layer on Windows. Since the OS will likely be configured for more than one RSS queue in most cases, we should support multiple queues.

## Proposed Solution

Implement a round-robin polling mechanism for multiple RSS queues. This will work if we run the inetstack on only a single core.

## Alternative Solutions

Instantiate one inetstack per RSS queue. If we need to expand to multiple cores, this would be the way to go to avoid lock contention.

kyleholohan:
Update XDP to bind to all receive queues. Receive queues are always a zero-based contiguous sequence of indices.

This approach is a little problematic because it could ignore other errors instantiating a socket on the non-zeroth queue, such as out of memory, etc. Alternatives get complex quickly. Looking at MsQuic, that platform uses a similar approach in a separate round of socket creation: it creates sockets for the express purposes of validating queues and observing CPU identity of each queue. Alternatively, the software could query the `MSFT_NetAdapterRssSettingData` WMI data source to determine this info (more info [here](https://learn.microsoft.com/en-us/windows/win32/fwp/wmi/netadaptercimprov/msft-netadapterrsssettingdata).

This also fixes a bug in UmemReg which dropped the allocated user memory after the constructor.

See also #1429

### DIFF
--- a/src/rust/catpowder/win/runtime.rs
+++ b/src/rust/catpowder/win/runtime.rs
@@ -9,24 +9,29 @@ use crate::{
     catpowder::win::{
         api::XdpApi,
         ring::{RxRing, TxRing, XdpBuffer},
-    },
-    demi_sgarray_t, demi_sgaseg_t,
-    demikernel::config::Config,
-    inetstack::protocols::{layer1::PhysicalLayer, MAX_HEADER_SIZE},
-    runtime::{
+    }, demi_sgarray_t, demi_sgaseg_t, demikernel::config::Config, inetstack::protocols::{layer1::PhysicalLayer, MAX_HEADER_SIZE}, runtime::{
         fail::Fail,
         libxdp,
         memory::{DemiBuffer, MemoryRuntime},
         network::consts::RECEIVE_BATCH_SIZE,
         Runtime, SharedObject,
-    },
+    }
 };
 use ::arrayvec::ArrayVec;
 use ::libc::c_void;
 use ::std::{
-    borrow::{Borrow, BorrowMut},
+    borrow::BorrowMut,
     mem,
 };
+use windows::Win32::{
+    Foundation::ERROR_INSUFFICIENT_BUFFER,
+     System::SystemInformation::{
+        GetLogicalProcessorInformationEx,
+        RelationProcessorCore,
+        SYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX
+    }
+};
+
 
 //======================================================================================================================
 // Structures
@@ -40,7 +45,7 @@ pub struct SharedCatpowderRuntime(SharedObject<CatpowderRuntimeInner>);
 struct CatpowderRuntimeInner {
     api: XdpApi,
     tx: TxRing,
-    rx: RxRing,
+    rx_rings: Vec<RxRing>,
 }
 //======================================================================================================================
 // Implementations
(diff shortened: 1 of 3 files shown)


src/rust/catpowder/win/runtime.rs
```diff
@@ -57,16 +62,26 @@ impl SharedCatpowderRuntime {
...
         let mut api: XdpApi = XdpApi::new()?;
 
         // Open TX and RX rings
-        let tx: TxRing = TxRing::new(&mut api, Self::RING_LENGTH, ifindex, QUEUEID)?;
-        let rx: RxRing = RxRing::new(&mut api, Self::RING_LENGTH, ifindex, QUEUEID)?;
+        let tx: TxRing = TxRing::new(&mut api, Self::RING_LENGTH, ifindex, 0)?;
+
+        let sys_proc_count: usize = count_processor_cores()?;
+        let mut rx_rings: Vec<RxRing> = Vec::with_capacity(sys_proc_count);
+        rx_rings.push(RxRing::new(&mut api, Self::RING_LENGTH, ifindex, 0)?);
```
anandbonde, line 74: Can this be moved inside the loop? Also, if this fails, the function would fail, but inside the loop, failures are ignored when creating RxRings for other queueids. Is it ok to let those errors be ignored? Maybe we should print a warn! msg...?

src/rust/catpowder/win/runtime.rs
```diff
@@ -126,19 +141,23 @@ impl PhysicalLayer for SharedCatpowderRuntime {
         let mut ret: ArrayVec<DemiBuffer, RECEIVE_BATCH_SIZE> = ArrayVec::new();
         let mut idx: u32 = 0;
 
-        if self.0.borrow_mut().rx.reserve_rx(Self::RING_LENGTH, &mut idx) == Self::RING_LENGTH {
-            let xdp_buffer: XdpBuffer = self.0.borrow().rx.get_buffer(idx);
-            let out: Vec<u8> = xdp_buffer.into();
-
-            let dbuf: DemiBuffer = DemiBuffer::from_slice(&out)?;
+        for rx in self.0.borrow_mut().rx.iter_mut() {
+            if rx.reserve_rx(Self::RING_LENGTH, &mut idx) == Self::RING_LENGTH {
```
anandbonde, line 148: We can possibly reducing nesting by inverting this condition.

src/rust/catpowder/win/runtime.rs
```diff
@@ -126,19 +141,23 @@ impl PhysicalLayer for SharedCatpowderRuntime {
         let mut ret: ArrayVec<DemiBuffer, RECEIVE_BATCH_SIZE> = ArrayVec::new();
         let mut idx: u32 = 0;
 
-        if self.0.borrow_mut().rx.reserve_rx(Self::RING_LENGTH, &mut idx) == Self::RING_LENGTH {
-            let xdp_buffer: XdpBuffer = self.0.borrow().rx.get_buffer(idx);
-            let out: Vec<u8> = xdp_buffer.into();
-
-            let dbuf: DemiBuffer = DemiBuffer::from_slice(&out)?;
+        for rx in self.0.borrow_mut().rx.iter_mut() {
```
anandbonde: Would it be possible to use `self.api` (for clarity)... instead of `self.0`?

anandbonde (approved):
Looks good.

src/rust/catpowder/win/runtime.rs
kyleholohan: This is the caveat mentioned in the PR description. The only trivial way we have to determine the number of RSS queues is to try to bind to each subsequent index until bind fails.
There are three ways we could have done this:
1. As is; treat all failures on queue binding >0 as an indication that we've reached the max RSS queue ID.
2. Same as 1, but refactor `RxQueue` and `Socket` to inform on error specifically from `XskBind`. We still have no way to determine failures from an invalid RSS queue ID versus other failures with `XskBind`, so this still can have mis-identified errors.
3. We can add infrastructure to query MSI to determine the RSS settings. This is probably a better solution, but is quite complex and will take some doing to get right.

I chose 1 for expediency, with the expectation of targeting a solution like 3 in the future.

kyleholohan: My preference would be to leave this as-is for now.

src/rust/catpowder/win/runtime.rs
kyleholohan: This happens per ring for all N rings, so the order cannot be inverted.

src/rust/catpowder/win/runtime.rs
anandbonde: This is not an issue. I misread. Closing the issue.

PR #1441, opened 2024-10-16
[inetstack] Add simulator test for out of order FIN

iyzhang:
This PR adds a new test for receiving an out of order FIN. It also fixes a few bugs around our close protocol:

1. Out of order FINs that were not on data carrying packets were not actually being stored for later.
2. When the out of order FIN is stored for later, we should not immediately process the FIN.
3. When the out of order FIN is stored for later, we should check on every arriving packet if the FIN is now ready because it might be by itself without any out of order data.

### DIFF
--- a/src/rust/inetstack/protocols/layer4/tcp/established/ctrlblk.rs
+++ b/src/rust/inetstack/protocols/layer4/tcp/established/ctrlblk.rs
@@ -417,7 +417,7 @@ impl SharedControlBlock {
             warn!("Got packet with URG bit set!");
         }
 
-        if data.len() > 0 {
+        if data.len() > 0 || header.fin {
             self.process_data(&mut header, data, seg_start, seg_end, seg_len)?;
         }
         self.process_remote_close(&header)?;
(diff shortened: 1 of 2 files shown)


src/rust/inetstack/protocols/layer4/tcp/established/ctrlblk.rs
```diff
@@ -670,6 +670,8 @@ impl SharedControlBlock {
                             seg_len -= 1;
                             self.store_out_of_order_fin(seg_end);
                             seg_end = seg_end - SeqNumber::from(1);
+                            // Clear header FIN and process later.
```
anandbonde, line 673: What does "process later" mean here? Is this already implicitly happening in the code, or something that will happen from now on?

anandbonde (approved):
Good to go.

src/rust/inetstack/protocols/layer4/tcp/established/ctrlblk.rs
iyzhang: Ah, I can be more clear but it means process it later when the rest of the packets before this sequence number arrive.

anandbonde: I see. If it's implicit and easy to understand from reading the code, then we should let the reader rely on the semantics based on code itself instead of explaining/mentioning it here. Else it will leave the reader guessing.

OR - we should explain it fully in this comment.

PR #1442, opened 2024-10-18
[inetstack] Enhancement: Move ephemeral port allocator

iyzhang:
This PR moves the ephemeral port allocator into the inetstack since it is not needed for the other libOSes.

### DIFF
--- a/src/rust/demikernel/libos/network/libos.rs
+++ b/src/rust/demikernel/libos/network/libos.rs
@@ -28,11 +28,10 @@ use ::futures::FutureExt;
 use ::socket2::{Domain, Protocol, Type};
 use ::std::{
     mem,
-    net::{SocketAddr, SocketAddrV4},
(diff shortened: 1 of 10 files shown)


src/rust/demikernel/libos/network/libos.rs
```diff
@@ -28,10 +28,11 @@ use ::futures::FutureExt;
 use ::socket2::{Domain, Protocol, Type};
 use ::std::{
     mem,
-    net::{Ipv4Addr, SocketAddr, SocketAddrV4},
+    net::{SocketAddr, SocketAddrV4},
     ops::{Deref, DerefMut},
     time::Duration,
 };
+use std::net::Ipv4Addr;
```
anandbonde: If you add `::std`, it will be merged with the above.

anandbonde (approved):
Looks good.

PR #1443, opened 2024-10-18
[demikernel] Enhancement: cleanup comments, rename

anandbonde:
Cleaned up unnecessary and redundant comments which just explain what the accompanying code does. Such comments only add to visual clutter and more lines to the source code files.

Renamed a few variables for clarity and better code flow while reading.

### DIFF
--- a/examples/tcp-echo/server.rs
+++ b/examples/tcp-echo/server.rs
@@ -34,18 +34,12 @@ pub const SOCK_STREAM: i32 = libc::SOCK_STREAM;
 // Structures
 //======================================================================================================================
 
-/// A TCP echo server.
 pub struct TcpEchoServer {
-    /// Underlying libOS.
     libos: LibOS,
-    /// Local socket descriptor.
-    sockqd: QDesc,
-    /// Set of connected clients.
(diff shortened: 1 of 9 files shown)


examples/tcp-echo/server.rs
```diff
@@ -34,62 +34,51 @@ pub const SOCK_STREAM: i32 = libc::SOCK_STREAM;
...
     libos: LibOS,
-    /// Local socket descriptor.
-    sockqd: QDesc,
-    /// Set of connected clients.
-    clients: HashSet<QDesc>,
-    /// List of pending operations.
-    qts: Vec<QToken>,
-    /// Reverse lookup table of pending operations.
-    qts_reverse: HashMap<QToken, QDesc>,
+    local_sockqd: QDesc,
```
iyzhang: Not super important but maybe this should be better named `listening_qd`

iyzhang (approved):

examples/tcp-echo/server.rs
anandbonde: Done

PR #1445, opened 2024-10-21
[testing] Enhancement: Two new tests for pop

iyzhang:
This PR adds two basic tests for pop:
1. A duplicate packet arrives before the ack goes out.
2. A packet arrives out of order before the previous packet or the ack (and causing an ack).


anandbonde (approved):
Looks good.

Issue #1435, opened 2024-10-08
[Catpowder] Windows: support co-hosting of Demikernel when XDP is in use


kyleholohan:
## Context

Currently, the XDP programs created by catpowder will capture all traffic on the interface. This is not necessarily desirable, as it will prevent regular traffic from flowing through the Windows kernel data path. Additionally, the port allocation schema will not respect ports reserved by the kernel.

## Proposed Solution

Restructure the Catpowder windows LibOS to rebind XDP rules when a socket is bound or connected. Scope the rules such that XDP only inspects traffic relevant to the protocol/port. Add a new port manager to the layer4 components to replace the ephemeral port allocation and port binding which will validate ports against kernel-allocated ports (e.g., but creating a WinSock socket and binding it) and reserve those ports for use by Demikernel (i.e., so that no other application can attempt to use them through the kernel networking stack)

## Alternative Solutions

I have no idea whether this is supported by XDP for Windows, but in theory unneeded traffic could be re-injected to the driver stack for processing by the kernel.

PR #1446, opened 2024-10-23
[catpowder] Feature: Add port-based filtering to catpowder backend

kyleholohan:
This PR switches the windows catpowder backend from filtering all traffic to only traffic for bound ports. See also #1435 . This is better, but still problematic. Currently, there is no way to notify back from the layer4 module when a port is closed (and no longer needed, per any linger requirements). There is also no coordination with the OS re: port assignment, so ephemeral port assignment can steal ports from other programs on the system. This leaves the burden of port assignment on the client, hence the use of reprogram on bind only.


kyleholohan:
Taking a different approach

PR #1447, opened 2024-10-28
[catpowder] Bug fix: RSS queue detection issue

kyleholohan:
Change RSS queue count detection from using `RxRing`s to `TxRing`s. `RxRing`s do late-binding validation after the queue is programmed; `TxRing`s do immediate validation. This is causing issues with subsequent changes to catpowderW

### DIFF
--- a/src/rust/catpowder/win/runtime.rs
+++ b/src/rust/catpowder/win/runtime.rs
@@ -67,21 +67,10 @@ impl SharedCatpowderRuntime {
         // Open TX and RX rings
         let tx: TxRing = TxRing::new(&mut api, Self::RING_LENGTH, ifindex, 0)?;
 
-        let sys_proc_count: usize = count_processor_cores()?;
-        let mut rx_rings: Vec<RxRing> = Vec::with_capacity(sys_proc_count);
-        rx_rings.push(RxRing::new(&mut api, Self::RING_LENGTH, ifindex, 0)?);
-        for queueid in 1..sys_proc_count {
-            match RxRing::new(&mut api, Self::RING_LENGTH, ifindex, queueid as u32) {
-                Ok(rx) => rx_rings.push(rx),
-                Err(e) => {
-                    warn!(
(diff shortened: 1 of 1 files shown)


src/rust/catpowder/win/runtime.rs
```diff
@@ -228,6 +217,28 @@ fn count_processor_cores() -> Result<usize, Fail> {
...
+    let sys_proc_count: u32 = count_processor_cores()? as u32;
+
+    for queueid in 1..sys_proc_count {
+        match TxRing::new(api, DUMMY_QUEUE_LENGTH, ifindex, queueid) {
+            Ok(_) => (),
+            Err(e) => {
+                warn!(
+                    "Failed to create TX ring on queue {}: {:?}. This is only an error if {} is a valid RSS queue \
+                     ID",
+                    queueid, queueid, e
```
anandbonde: Is this the correct ordering of the variables?

src/rust/catpowder/win/runtime.rs
```diff
@@ -228,6 +217,28 @@ fn count_processor_cores() -> Result<usize, Fail> {
     return Ok(core_count);
 }
 
+/// Deduces the RSS settings for the given interface. Returns the number of valid RSS queues for the interface.
+fn deduce_rss_settings(api: &mut XdpApi, ifindex: u32) -> Result<u32, Fail> {
+    const DUMMY_QUEUE_LENGTH: u32 = 1;
+    let sys_proc_count: u32 = count_processor_cores()? as u32;
+
+    for queueid in 1..sys_proc_count {
```
anandbonde, line 227: Why is this loop from 1 to sys_proc_count-1? It would be good to add a comment explaining why this is done.

anandbonde (approved):

src/rust/catpowder/win/runtime.rs
kyleholohan: Added a comment.

src/rust/catpowder/win/runtime.rs
kyleholohan: Good catch; thanks.

PR #1448, opened 2024-10-30
[runtime] Bug Fix: Ensure condition variables sleep

iyzhang:
This PR fixes an issue in our condition variables where the following may happen:

1. Condition variable wakes
2. Condition is not met
3. New condition variable is created and immediately becomes runnable.
4. Condition is still not met
5. Other condition variables are starved instead of being woken

### DIFF
--- a/src/rust/runtime/condition_variable.rs
+++ b/src/rust/runtime/condition_variable.rs
@@ -18,14 +18,14 @@ use ::std::{
 // Structures
 //======================================================================================================================
 
-#[derive(Eq, PartialEq)]
+#[derive(Eq, PartialEq, Clone, Copy)]
 /// The state of the coroutine using this condition variable.
 enum YieldState {
     Running,
     Yielded,
 }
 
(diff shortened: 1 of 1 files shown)


src/rust/runtime/condition_variable.rs
```diff
@@ -131,17 +131,22 @@ impl Future for YieldPoint {
...
-            Poll::Pending
-        }
+                self_.state = YieldState::Yielded;
+            },
+            YieldState::Yielded if num_ready > 0 => {
+                self_.cond_var.num_ready -= 1;
+                self_.state = YieldState::Running;
+                return Poll::Ready(());
+            },
+            _ => (),
```
anandbonde: This line can be changed to `_ => Poll::Pending` to make all return statements coherent within the match.

anandbonde (approved):
Looks good! Thanks for fixing this!

src/rust/runtime/condition_variable.rs
iyzhang: sure, makes sense

PR #1449, opened 2024-10-30
Bug Fix: Retransmit FIN on close

iyzhang:
Our previous implementation did not retransmit the FIN on close if it was not acked. This was due in part to our coroutines exiting before the close protocol had finished and not continuing to run for things like retransmission. But we should have also just checked that we actually received an ack to our FIN before exiting.

### DIFF
--- a/src/rust/inetstack/protocols/layer4/tcp/established/sender.rs
+++ b/src/rust/inetstack/protocols/layer4/tcp/established/sender.rs
@@ -3,7 +3,6 @@
 
 use crate::{
     collections::{async_queue::SharedAsyncQueue, async_value::SharedAsyncValue},
-    expect_ok,
     inetstack::protocols::layer4::tcp::{
         established::{rto::RtoCalculator, SharedControlBlock},
         header::TcpHeader,
@@ -25,7 +24,7 @@ use std::cmp;
 // buffer structure that held everything we need directly, thus avoiding this extra wrapper.
 //
 pub struct UnackedSegment {
-    pub bytes: DemiBuffer,
+    pub bytes: Option<DemiBuffer>,
     // Set to `None` on retransmission to implement Karn's algorithm.
     pub initial_tx: Option<Instant>,
 }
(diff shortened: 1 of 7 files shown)


src/rust/inetstack/protocols/layer4/tcp/established/sender.rs
```diff
@@ -121,35 +130,161 @@ impl Sender {
...
+                        }
+                        // Otherwise, wait until something limiting the window changes and then try again to finish sending
+                        // the segment.
+                        futures::select_biased! {
+                            _ = send_unacked_watched.wait_for_change(None).fuse() => (),
+                            _ = self.send_next_seq_no.wait_for_change(None).fuse() => (),
+                            _ = win_sz_watched.wait_for_change(None).fuse() => (),
+                            _ = cwnd_watched.wait_for_change(None).fuse() => (),
+                            _ = ltci_watched.wait_for_change(None).fuse() => (),
+                        };
```
anandbonde: Is this the priority order (left to right) in which you intend to process these futures? If yes, then this is good, else we could use `select!` instead.

src/rust/inetstack/protocols/layer4/tcp/established/sender.rs
```diff
@@ -121,35 +130,161 @@ impl Sender {
...
+
+    pub async fn background_sender(&mut self, mut cb: SharedControlBlock) -> Result<Never, Fail> {
+        let mut send_unacked_watched: SharedAsyncValue<SeqNumber> = self.send_unacked.clone();
+        let mut cwnd_watched: SharedAsyncValue<u32> = cb.congestion_control_get_cwnd();
+
+        // The limited transmit algorithm may increase the effective size of cwnd by up to 2 * mss.
+        let mut ltci_watched: SharedAsyncValue<u32> = cb.congestion_control_get_limited_transmit_cwnd_increase();
+        let mut win_sz_watched: SharedAsyncValue<u32> = self.send_window.clone();
+
+        loop {
```
anandbonde, line 234: This loop is very long - it's hard to keep track of the logic while reading top to bottom. Can it be broken into meaningful chunks? (e.g. the inner/nested loop and the if/else bodies)

src/rust/inetstack/protocols/layer4/tcp/established/sender.rs
```diff
@@ -207,113 +337,28 @@ impl Sender {
...
-                    futures::select_biased! {
-                        _ = send_unacked_watched.wait_for_change(None).fuse() => (),
-                        _ = self.send_next.wait_for_change(None).fuse() => (),
-                        _ = win_sz_watched.wait_for_change(None).fuse() => (),
-                        _ = cwnd_watched.wait_for_change(None).fuse() => (),
-                        _ = ltci_watched.wait_for_change(None).fuse() => (),
-                    };
-                }
-            }
+        if win_sz > 0 && win_sz >= sent_data && effective_cwnd >= sent_data {
```
anandbonde: We could extract this long condition into a variable with a name that denotes what we intend to check for.

anandbonde (changes requested):

src/rust/inetstack/protocols/layer4/tcp/established/sender.rs
iyzhang: yes, I think priority order is important.

anandbonde (approved):
Looks good.

PR #1450, opened 2024-10-30
[collections] Bug Fix: Increasing arbitrary timeout

iyzhang:
This PR increases an arbitrary timeout in our data structures to keep our slower libOSes from timing out.


anandbonde (approved):
Looks good.

PR #1451, opened 2024-10-31
[docs] Improve docs for Windows initial setup

liu-weipeng:


doc/building-windows.md
```diff
@@ -79,3 +79,5 @@ nmake all
 # Build Demikernel in debug mode with the default configuration.
 nmake DEBUG=yes all
 ```
+
+Alternatively, you can also run `cargo build` in the root directory. (after which you may run `cargo build --examples` to build examples)
```
anandbonde, line 83: Is there any particular reason to use `cargo` commands directly to build? It would be good to just limit the interface to build using the `nmake` command, so that we have the room to change the behavior at a later point. If there's a need to use this command particularly, we should document it here.

anandbonde:
Closing this PR now, let's reopen if needed.

PR #1452, opened 2024-11-01
[inetstack] Bug Fix: Remove ack queue

iyzhang:
The ack queue is no longer needed to block push operations until the other end acks. Instead, we now wait for the ack sequence number to rise above the sequence number assigned to the buffer range that we pushed.


anandbonde (approved):
Looks good.

PR #1453, opened 2024-11-06
[catpowder] Support NetVSC/VF dual NIC in hyper-V VMs

liu-weipeng:
### Motivation
On certain Hyper-V provisioned VMs, there're two NICs, one NetVSC, another VF. They can share same MAC address to act as single device from external point of view. Because VF is faster, it may be used to optimize data plane performance, while NetVSC is used for control plane. This is transparent to application layer if networking stack is done within the kernel. However, XDP bypasses the kernel, so we have to deal with it manually. This change would allow user to configure two interfaces for XDP to attach to and poll, as incoming packets could come from either the VF or the NetVSC.

### DIFF
--- a/src/rust/catpowder/win/runtime.rs
+++ b/src/rust/catpowder/win/runtime.rs
@@ -44,6 +44,7 @@ struct CatpowderRuntimeInner {
     api: XdpApi,
     tx: TxRing,
     rx_rings: Vec<RxRing>,
+    vf_rx_rings: Vec<RxRing>,
 }
 //======================================================================================================================
 // Implementations
@@ -60,6 +61,7 @@ impl SharedCatpowderRuntime {
     /// Instantiates a new XDP runtime.
     pub fn new(config: &Config) -> Result<Self, Fail> {
         let ifindex: u32 = config.local_interface_index()?;
+        let vf_if_index = config.local_vf_interface_index(); // N.B. this one is optional
 
         trace!("Creating XDP runtime.");
         let mut api: XdpApi = XdpApi::new()?;
@@ -72,9 +74,30 @@ impl SharedCatpowderRuntime {
         for queueid in 0..queue_count {
             rx_rings.push(RxRing::new(&mut api, Self::RING_LENGTH, ifindex, queueid as u32)?);
         }
-        trace!("Created {} RX rings.", rx_rings.len());
+        trace!("Created {} RX rings on interface {}", rx_rings.len(), ifindex);
 
-        Ok(Self(SharedObject::new(CatpowderRuntimeInner { api, tx, rx_rings })))
+        if let Ok(vf_if_index) = vf_if_index { // Optionally create VF RX rings
+            let vf_queue_count = deduce_rss_settings(&mut api, vf_if_index)?;
+            let mut vf_rx_rings = Vec::with_capacity(vf_queue_count as usize);
+            for queueid in 0..vf_queue_count {
+                vf_rx_rings.push(RxRing::new(&mut api, Self::RING_LENGTH, vf_if_index, queueid as u32)?);
+            }
+            trace!("Created {} RX rings on VF interface {}.", vf_rx_rings.len(), vf_if_index);
+
+            Ok(Self(SharedObject::new(CatpowderRuntimeInner {
+                api,
+                tx,
+                rx_rings,
+                vf_rx_rings,
+            })))
+        } else {
+            Ok(Self(SharedObject::new(CatpowderRuntimeInner {
+                api,
+                tx,
+                rx_rings,
+                vf_rx_rings: Vec::new(),
+            })))
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/rust/demikernel/config.rs
```diff
@@ -292,6 +299,16 @@ impl Config {
         }
     }
 
+    #[cfg(all(feature = "catpowder-libos", target_os = "windows"))]
+    pub fn local_vf_interface_index(&self) -> Result<u32, Fail> {
+        // Parse local MAC address.
```
anandbonde: This comment seems incorrect.

src/rust/demikernel/config.rs
```diff
@@ -60,8 +60,15 @@ mod raw_socket_config {
...
     pub const LOCAL_INTERFACE_NAME: &str = "linux_interface_name";
+
+    // the primary interface index. this should be the virtualized interface for VMs.
     #[cfg(target_os = "windows")]
     pub const LOCAL_INTERFACE_INDEX: &str = "xdp_interface_index";
+
+    // N.B. hyper-V VMs can have both NetVSC and VF interfaces working in tandem, in which case
+    // we need to listen to the corresponding VF interface as well.
+    #[cfg(target_os = "windows")]
+    pub const LOCAL_VF_INTERFACE_INDEX: &str = "xdp_vf_interface_index";
```
anandbonde, line 71: Should this also be guarded behind "catpowder-libos" feature as well?

src/rust/demikernel/config.rs
```diff
@@ -60,8 +60,15 @@ mod raw_socket_config {
     pub const SECTION_NAME: &str = "raw_socket";
     #[cfg(target_os = "linux")]
     pub const LOCAL_INTERFACE_NAME: &str = "linux_interface_name";
+
+    // the primary interface index. this should be the virtualized interface for VMs.
```
anandbonde: The casing is off in this comment.

src/rust/catpowder/win/runtime.rs
```diff
@@ -153,6 +162,25 @@ impl PhysicalLayer for SharedCatpowderRuntime {
             }
         }
 
+        for rx in self.0.borrow_mut().vf_rx_rings.iter_mut() {
+            if rx.reserve_rx(Self::RING_LENGTH, &mut idx) == Self::RING_LENGTH {
+                let xdp_buffer: XdpBuffer = rx.get_buffer(idx);
+                let dbuf: DemiBuffer = DemiBuffer::from_slice(&*xdp_buffer)?;
+                rx.release_rx(Self::RING_LENGTH);
```
anandbonde, line 183: We typically leave an empty line after variable declarations for clarity. Also this is followed in other places in this file. It would be good to maintain consistency in this aspect unless you prefer this style for grouping/clarity.

src/rust/catpowder/win/runtime.rs
```diff
@@ -72,9 +74,16 @@ impl SharedCatpowderRuntime {
...
 
-        Ok(Self(SharedObject::new(CatpowderRuntimeInner { api, tx, rx_rings })))
+        let vf_queue_count = deduce_rss_settings(&mut api, vf_if_index)?;
+        let mut vf_rx_rings= Vec::with_capacity(vf_queue_count as usize);
+        for queueid in 0..vf_queue_count {
+            vf_rx_rings.push(RxRing::new(&mut api, Self::RING_LENGTH, vf_if_index, queueid as u32)?);
+        }
+        trace!("Created {} RX rings on interface {}.", vf_rx_rings.len(), vf_if_index);
+
+        Ok(Self(SharedObject::new(CatpowderRuntimeInner { api, tx, rx_rings, vf_rx_rings })))
```
anandbonde: Could you please run the formatter on this code? If you are using vscode it can automatically do this for you.

anandbonde (dismissed):
* Could you add PR description?
* The commit message format is incorrect.

src/rust/demikernel/config.rs
liu-weipeng: I believe line #58 puts the whole struct under `catpowder-libos` already.

iyzhang (approved):
Looks fine to me. Is there a reason to keep the VF rx queues separate from the other rx queues? Could we just append them to the same vec?

liu-weipeng:
> Looks fine to me. Is there a reason to keep the VF rx queues separate from the other rx queues? Could we just append them to the same vec?

I think we can merge them into one, although keeping them separate might serve some diagnostic purpose now or in the future.

src/rust/catpowder/win/runtime.rs
liu-weipeng: this is good for now. Trying to conform to the style of the rest of the code.

src/rust/demikernel/config.rs
anandbonde: The #[cfg...] does not apply to multiple lines in Rust. It's only applicable to the immediate next line.

anandbonde (approved):
Looks good.

src/rust/catpowder/win/runtime.rs
anandbonde: Ok.

src/rust/demikernel/config.rs
anandbonde: Makes sense.

PR #1454, opened 2024-11-06
[testing] Bug Fix: Correctly exit

iyzhang:
This PR fixes an issue in the tcp-echo closed loop client where it never exists if the log interval is set


anandbonde (approved):
Looks good.

PR #1455, opened 2024-11-06
[inetstack] Enhancement: replace immediate send

iyzhang:
This PR places back the immediate send path that use to exist for inlining a send if the send window is open.


iyzhang:
This PR improves single client echo latency by 4us on prometheus8/9. packet size: 1024 bytes. Run for 10 seconds. Each run 3 times.

Average before PR: 20.7us
Average after PR: 16.3us

anandbonde (approved):
Looks good.

PR #1456, opened 2024-11-10
[runtime] Enhancement: Add some reasonable defaults

iyzhang:
This PR removes some defaults and adds others:
1. It removes the default timeout values for asynchronous queues and values. This reduces the number of timers that we need to set and then cancel. If None is passed in, now we wait forever instead of for 1000 seconds.
2. It adds a default size for the AsyncQueue and condition variable vecdeque because increasing the size of these for the first few elements is very expensive and hurts tail latency.

### DIFF
--- a/src/rust/runtime/network/socket/state.rs
+++ b/src/rust/runtime/network/socket/state.rs
@@ -10,14 +10,6 @@ use crate::{
     runtime::{fail::Fail, network::socket::operation::SocketOp},
 };
 use ::socket2::Type;
-use ::std::time::Duration;
-
-//======================================================================================================================
-// Structures
-//======================================================================================================================
-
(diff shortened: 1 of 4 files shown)


src/rust/runtime/network/socket/state.rs
```diff
@@ -97,7 +89,7 @@ impl SocketStateMachine {
             match self.may_connect() {
                 Ok(()) => {
                     // If either a time out or the current state changed, check it again.
-                    _ = self.current.clone().wait_for_change(Some(TIMEOUT)).await;
```
anandbonde, line 100: The above comment doesn't hold true now. Same for other places where this comment is present.

anandbonde (approved):
Looks good.

PR #1457, opened 2024-11-13
[testing] Bug Fix: Do not leave lingering qtokens

iyzhang:
Remove data structure from tcp-echo that impacts performance.


anandbonde:
This change doesn’t seem to have any effect on the performance so we will just leave it here.

iyzhang:
This PR fixes #1492

anandbonde (approved):
Looks good.

PR #1459, opened 2024-11-14
[logging] Enhancement: Turn on more log details

iyzhang:
This PR turns on more details in our logging framework including timestamps, line numbers and thread ID. To help with the thread ID usefulness, I've also updated our unit tests with thread names.


anandbonde (approved):
Looks good.

PR #1460, opened 2024-11-15
[inetstack] Enhancement: Add receiver to arch

iyzhang:
This PR adds a receiver module for the receive path on established sockets. It also removes two SharedAsyncQueues to reduce latency.


src/rust/inetstack/protocols/layer4/tcp/passive_open.rs
```diff
@@ -134,64 +121,47 @@ impl SharedPassiveSocket {
...
+
+        // See if this packet is for an already established but not accepted socket.
+        if let Some((_, socket)) = self.ready.get_values().find(|(addr, _)| *addr == remote) {
+            if let Ok(socket) = socket {
+                socket.get_cb().receive(tcp_hdr, buf);
             }
+            return;
         }
+
+        // Otherwise if not a SYN, then this packet is not for a new connection and we throw it away.
```
anandbonde, line 142: This looks like an independent chunk, can be a separate fn.

anandbonde (approved):
Looks good.

PR #1462, opened 2024-11-15
[catpowder] Enhancement: Format win runtime

iyzhang:
Just a quick cargo fmt on some Windows side code because we are not enforcing that yet.


anandbonde (approved):

anandbonde (approved):
Approved.

PR #1463, opened 2024-11-15
[inetstack] Remove some dead code

iyzhang:
We do not use the dead_socket transmission channel any more, so this PR removes it.


anandbonde (approved):
Looks good.

PR #1464, opened 2024-11-21
[inetstack] Enhancement: Remove circular dependencies

iyzhang:
This PR removes circular dependencies between the sender, receiver and control block in established TCP sockets. We move all state into the control block structure, per the TCP recommendation, and to make it easier to move the state later. However, we pass the control block to static functions in the receiver and sender for receive and send path processing.


anandbonde (approved):
Looks good.

PR #1465, opened 2024-11-23
[inetstack] Handle RST packet from peer endpoint

anandbonde:
Handle RST packet sent by the peer and close the connection. This is crucial because the profiler information is only emited when the process exits cleanly. If this is not done, we cannot print the profiler information.


iyzhang (approved):
Looks good to me.

PR #1469, opened 2024-11-26
[inetstack] Bug Fix: Check windows sizes before allocating socket

### LINKED ISSUE #199: [inetstack] Check Maximum Sizes TCP Can Handle
TCP sequence numbers are 32-bit values, which limits the "distance" between sequence numbers that can be reasonably compared to be "greater than" or "less than" the other.  Also, even with window scaling, a TCP's advertised receive window can't be greater than 1 GiB.  We should review the code to ensure that there aren't any places where we could overflow these values.

[This Issue is loosely linked to Issue demikernel/demikernel#214]

iyzhang:
This closes #199


anandbonde (approved):
Looks good.

PR #1470, opened 2024-11-26
[catpowder] Feature: Add cohosting mode to XDP backend

kyleholohan:
Add a cohost setting with redirected ports specified in the config/environment as an initial take on cohosting.

Ideally, the system would do this automatically based on used ports, but there's currently no easy way to hook this up to the inetstack system.


iyzhang (approved):
LGTM

anandbonde (approved):
Looks good.

PR #1471, opened 2024-11-26
[inetstack] Enhancement: Move consts, types and config to inetstack

iyzhang:
This PR reorganizes all consts, types and config options that are specific to the inetstack from the runtime to the inetstack.


anandbonde (approved):
Looks good.

PR #1476, opened 2024-12-11
[inetstack] Bug Fix: Check extra error cases

iyzhang:
This PR includes the following:
1. Additional checks for error paths on passive open.
2. Three additional test cases for those error paths. (I also renamed another error test).
3. A removal of some code that now no longer needs the AsyncQueue data structure.


src/rust/inetstack/protocols/layer4/tcp/established/mod.rs
```diff
@@ -164,9 +163,8 @@ impl SharedEstablishedSocket {
             layer3_endpoint,
         }));
 
-        trace!("inital receive_queue size {:?}", recv_queue.len());
         // Process all pending received packets while setting up the connection.
```
anandbonde: Does this comment need to be adjusted with the new logic in place?

log
anandbonde, line 1: Does this file need to be in git?

anandbonde (approved):
Looks good.

network_simulator/input/tcp/accept/accept-incorrect-ack.pkt
anandbonde, line 1: Later, we should think of renaming all the .pkt files using some explanatory naming scheme because currently it's difficult to understand what the test does without reading the full file. (e.g. `<what is being tested>_<what is the expectation>.pkt` etc.)

anandbonde (approved):
Looks good.

log
iyzhang: nope, removed! thanks!

PR #1479, opened 2024-12-13
[inetstack] Enhancement: Hardcode for Windows

### LINKED ISSUE #1391: [inetstack] Configurable dummy MAC address
## Context

On Azure, local MAC address to IP mappings have no meaning because all VMs run in VLANs. Thus, we can fill the destination MAC address in outgoing ethernet packets with anything that we want. However, occasionally, it needs to be a specific address for the packets to flow. For example, Windows expects 12-34-56-78-9a-bc

## Proposed Solution

We should make this constant configurable: [
DUMMY_MAC_ADDRESS](https://github.com/microsoft/demikernel/blob/25044f9605963a25965d6070ca8b33a4bcd9fcbb/src/rust/inetstack/protocols/layer3/arp/cache/mod.rs#L28)

iyzhang:
This closes #1391. Windows requires a specific dummy MAC on Azure, but Linux does not, so this should work for both.


anandbonde (approved):
Looks good.

PR #1480, opened 2024-12-16
[build] Enhancement: remove unnecessary echoes

anandbonde:
These unnecessary echo statements are causing duplicate output, which may cause confusion around how many times builds commands may be executed.


iyzhang (approved):
Thanks for fixing this

PR #1481, opened 2024-12-19
[profiler] Enhancement: write exclusive scope durations

anandbonde:
* Exclusive time is most important to understand the impact of each scope on the overall performance
* Compute percentages outside of Rust code for simplicity and correctness in case of multiple file outputs

### DIFF
--- a/src/rust/perftools/profiler/scope.rs
+++ b/src/rust/perftools/profiler/scope.rs
@@ -15,11 +15,11 @@ use ::std::{
     fmt::{self, Debug},
     future::Future,
     io,
+    ops::{Deref, DerefMut},
     pin::Pin,
     task::{Context, Poll},
     thread,
 };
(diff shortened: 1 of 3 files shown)


src/rust/perftools/profiler/scope.rs
```diff
@@ -36,6 +36,8 @@ pub struct Scope {
     pub num_calls: usize,
     /// In total, how much time has been spent in this scope?
     pub duration_sum: u64,
+    /// In total, how much time has been spent in this scope only (excluding chlid scopes)?
```
iyzhang: It seems a little convoluted to store the exclusive duration sum here just to print it. We should probably compute it and directly return it for printing without storing it in the scope.

iyzhang (approved):
Thanks for fixing!

PR #1482, opened 2024-12-26
[inetstack] Enhancement: correct timer statements

anandbonde:
The scope of timer statements governs what they measure. Also, nesting coroutine timers with regular timers causes scope reparenting issues based on the current design of the profiler, so correcting them.


iyzhang (approved):
Looks good to me

PR #1483, opened 2025-01-07
[profiler] Enhancment: panic on wrong timer usage

anandbonde:
Nesting async_timer!() within a timer!() scope causes reparenting issues in the profiler. To avoid such invalid uses in future, changing to panic if this happens. Also, removed all timers that violated this invariant.

### DIFF
--- a/src/rust/perftools/profiler/mod.rs
+++ b/src/rust/perftools/profiler/mod.rs
@@ -155,9 +155,7 @@ impl Profiler {
             current_scope.parent_scope.as_ref().cloned()
         } else {
             // This should not happen with proper usage.
-            log::error!("Called perftools::profiler::leave() while not in any scope");
-
-            None
+            unreachable!("Called perftools::profiler::leave() while not in any scope");
         };
     }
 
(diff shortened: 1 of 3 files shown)


src/rust/perftools/profiler/mod.rs
```diff
@@ -157,36 +155,33 @@ impl Profiler {
             current_scope.parent_scope.as_ref().cloned()
         } else {
             // This should not happen with proper usage.
-            log::error!("Called perftools::profiler::leave() while not in any scope");
-
-            None
+            panic!("Called perftools::profiler::leave() while not in any scope");
```
iyzhang: should this be unreachable?

iyzhang (approved):
looks good

src/rust/perftools/profiler/mod.rs
anandbonde: Yes, makes sense, will change it.

anandbonde: Done.

PR #1485, opened 2025-01-14
[scheduler] Enhancement: Add background coroutine group

iyzhang:
This PR splits our coroutines into foreground and background. We always prioritize foreground coroutines, running them until none are runnable before scheduling the background coroutines. The runtime controls this by polling different task groups in the scheduler rather than changing the scheduler.

### DIFF
--- a/network_simulator/input/tcp/close/close-local-retransmission.pkt
+++ b/network_simulator/input/tcp/close/close-local-retransmission.pkt
@@ -40,8 +40,9 @@
 
 // Receive FIN segment.
 +.1 TCP < F. seq 1(0) ack 1002 win 65535 <nop>
+
+// Succeed to close connection immediately because we set linger to 0.
++0 wait(500, ...) = 0
+
 // Send ACK on FIN segment.
 +.0 TCP > . seq 1002(0) ack 2 win 65534 <nop>
-
-// Succeed to close connection after 2 MLS.
-+240 wait(500, ...) = 0
--- a/src/rust/runtime/mod.rs
+++ b/src/rust/runtime/mod.rs
@@ -17,8 +17,8 @@ pub mod types;
 pub use condition_variable::SharedConditionVariable;
 mod poll;
 mod timer;
-pub use queue::{BackgroundTask, Operation, OperationResult, OperationTask, QDesc, QToken, QType};
-pub use scheduler::TaskId;
+pub use queue::{BackgroundTask, OperationResult, OperationTask, QDesc, QToken, QType};
+pub use scheduler::{Task, TaskId};
 
 #[cfg(feature = "libdpdk")]
 pub use demikernel_dpdk_bindings as libdpdk;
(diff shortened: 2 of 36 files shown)


network_simulator/input/tcp/close/close-local-retransmission.pkt
```diff
@@ -38,10 +38,11 @@
 // Send completes
 +.0 wait(500, ...) = 9
 
-// Receive FIN segment.
+ // Receive FIN segment.
```
anandbonde: extra space here

src/rust/runtime/mod.rs
```diff
@@ -454,6 +391,34 @@ impl SharedDemiRuntime {
         trace!("Check address in use: {:?}", socket_addrv4);
         self.socket_id_to_qdesc_map.is_in_use(socket_addrv4)
     }
+
+    pub fn poll_background_tasks(&mut self) {
+        let background_group_id: TaskId = self.background_group_id;
```
anandbonde, line 373: This can probably be inlined.

src/rust/runtime/scheduler/scheduler.rs
```diff
@@ -61,197 +42,78 @@ pub struct SharedScheduler(SharedObject<Scheduler>);
...
             }
-            // If we reach this point, then we have looped all the way around without finding any runnable tasks.
-            if self.current_group_id == starting_group_index {
-                return;
+            for id in ready_tasks {
+                // Now that we have a runnable task, actually poll it.
+                if let Some(task) = group.poll_notified_task_and_remove_if_ready(id) {
+                    completed_tasks.push(task);
+                }
+                iterations += 1;
```
anandbonde: Do we want to run until there are `max_iteration` number of ready tasks, or `max_iteration` number of times? Here, it seems to count the number of tasks that are ready.

src/rust/runtime/mod.rs
iyzhang: ah, this is just needed for ownership. Not sure if there is a more efficient way to write this.

anandbonde (approved):
Looks good.

src/rust/runtime/mod.rs
```diff
@@ -107,37 +109,35 @@ impl SharedDemiRuntime {
...
-    ) -> Result<QToken, Fail> {
-        self.insert_coroutine(task_name, coroutine)
+        self.insert_coroutine(task_name, Some(self.background_group_id), coroutine)
     }
 
     /// Inserts a coroutine of type T and task
     pub fn insert_coroutine<F: FusedFuture + 'static>(
         &mut self,
         task_name: &'static str,
+        group_id: Option<TaskId>,
```
anandbonde: The value passed into this parameter is `None` almost every time (except for io_polling coroutines). We should create a separate fn that does not accept this parameter at all to avoid ambigiuty. e.g., `insert_io_coroutine(self, task_name, coroutine)`.

iyzhang: done

PR #1486, opened 2025-01-14
[profiler] Enhancement: remove unnecessary timers

anandbonde:
remove unnecessary timers - these add no value to performance measurement/debugging

### DIFF
--- a/src/rust/demikernel/libos/mod.rs
+++ b/src/rust/demikernel/libos/mod.rs
@@ -57,8 +57,6 @@ pub enum LibOS {
 
 impl LibOS {
     pub fn new(libos_name: LibOSName, _perf_callback: Option<demi_callback_t>) -> Result<Self, Fail> {
-        timer!("demikernel::new");
-
         logging::initialize();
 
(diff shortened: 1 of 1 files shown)


src/rust/demikernel/libos/mod.rs
```diff
@@ -344,7 +338,6 @@ impl LibOS {
 
     pub fn sgaalloc(&mut self, size: usize) -> Result<demi_sgarray_t, Fail> {
         let result: Result<demi_sgarray_t, Fail> = {
-            timer!("demikernel::sgaalloc");
```
iyzhang: we might want to keep these just in case somehow this number goes up (ie., it becomes much more expensive to allocate memory through Demikernel)

iyzhang (approved):
lgtm, all the other functions are not data path calls, so I understand taking them out but we might want to keep the sga ones, especially as Kyle might be making some changes to how to allocate and free DemiBuffers.

src/rust/demikernel/libos/mod.rs
anandbonde: Done

PR #1489, opened 2025-01-24
[runtime] Enhancement: Make direct-mapping an env-var-driven feature

anandbonde:
Direct mapping can now be compiled in or out depending on an env var.


src/rust/collections/id_map.rs
```diff
@@ -5,8 +5,18 @@
 // Imports
 //======================================================================================================================
 
+//! The "direct-mapping" feature is used to expose internal ids directly to the user. This is useful for debugging and
+//! testing purposes. BEWARE: Turning on this feature will have a performance impact.
```
iyzhang: It should be that turning off direct mapping is useful for debugging and will impact performance.

iyzhang (approved):

PR #1490, opened 2025-01-26
[runtime] Enhancement: Move id_map into runtime

iyzhang:
This PR begins to move id_maps into a single unified place and use the data structure in a organized way to allocate ids.

### DIFF
--- a/src/rust/runtime/scheduler/scheduler.rs
+++ b/src/rust/runtime/scheduler/scheduler.rs
@@ -12,25 +12,29 @@
 //======================================================================================================================
 
 use crate::runtime::{
-    scheduler::{group::TaskGroup, Task, TaskId},
+    scheduler::{group::TaskGroup, Task},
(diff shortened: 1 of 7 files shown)


src/rust/runtime/scheduler/scheduler.rs
```diff
@@ -253,90 +235,64 @@ mod tests {
...
+            Ok(())
         } else {
             anyhow::bail!("task should have completed");
         }
-        Ok(())
     }
 
     #[test]
-    fn poll_next_with_one_long_task_completes_it() -> Result<()> {
+    fn poll_unrunnable_with_one_long_task_completes_it() -> Result<()> {
```
anandbonde: We can rename this to match the called fn (i.e., poll_until_unrunnable_...).

anandbonde (approved):
Looks good.

PR #1493, opened 2025-01-30
[testing] Enhancement: Removing duplicate code

iyzhang:
This PR removes some code that is shared between the sequential and concurrent clients for the tcp-echo test.


anandbonde (approved):
Looks good.

PR #1494, opened 2025-01-30
[dpdk-bindings] Enhancement: add indirect buffer methods

kyleholohan:
Add a few methods to DPDK bindings to support indirect buffers for upcoming memory pooling changes.


anandbonde (approved):
Looks good.

PR #1502, opened 2025-02-07
[runtime] Enhancement: return earliest task first and add some tests

iyzhang:
This PR reverses the order in which we iterate over the completed tasks, which ensures that we return the task earliest in the scheduler first. This does not guarantee that the earliest scheduled task is first, but it does ensure some amount of fairness. Since we do not have a specific policy yet for the order in which qtokens are returned, we use this one as a proxy for FIFO. This PR also adds some tests to ensure that the new behavior is correct.

If we want to actually ensure that the first scheduled qt is first, then we likely need to do some more sophisticated queuing. Also if we wanted to return the first qt in the list in wait_any further updates to the scheduler policy would be needed.

The PR looks large but is broken into a few parts (but a single commit because it won't compile):
1. Changes to the scheduler and ensuring the we run earlier tasks in the group pinslab first.
2. Changes to the runtime to match the completed returned tasks in the order in which they are returned.
3. Added tests to the runtime to test that we are respecting the scheduling policy.
4. Changes to the network simulator traces to account for this change to the scheduling policy.

### DIFF
--- a/src/rust/runtime/mod.rs
+++ b/src/rust/runtime/mod.rs
@@ -207,10 +207,8 @@ impl SharedDemiRuntime {
         timeout: Duration,
     ) -> Result<(usize, QToken, QDesc, OperationResult), Fail> {
         // If any are already complete, grab from the completion table, otherwise, make sure it is valid.
-        for (i, qt) in qts.iter().enumerate() {
-            if let Some((qd, result)) = self.completed_tasks.remove(qt) {
-                return Ok((i, *qt, qd, result));
-            } else if self.qtoken_to_scheduler_id.get(qt).is_none() {
+        for qt in qts.iter() {
+            if self.qtoken_to_scheduler_id.get(qt).is_none() {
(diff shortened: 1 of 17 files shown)


src/rust/runtime/mod.rs
```diff
@@ -208,77 +212,88 @@ impl SharedDemiRuntime {
...
+    /// completed task queue and then try again. Returns None if there are still no completed tasks after polling the
+    /// scheduler.
+    pub fn run_scheduler(&mut self) {
+        // 1. Run each of the background tasks once.
         self.poll_background_tasks();
-        self.poll_foreground_tasks()
+        // 2. Run all foreground tasks until none are ready.
+        let completed_tasks: Vec<OperationTask> = self.poll_foreground_tasks();
+
+        // 5. Update the list of previously completed tasks.
```
anandbonde: The numbering in the comments needs to be corrected.

anandbonde (approved):
Looks good.

PR #1504, opened 2025-02-11
[runtime] Enhancement: add a custom logging initializer

kyleholohan:
Add a custom logging initializer to `runtime.logging` mod. This allows clients to use non-stdio outputs. This also fixes a potential issue wherein the `LoggingHandle` is currently dropped. This is only a bug when a non-stdio logging output is used.


anandbonde (approved):
Looks good.

PR #1505, opened 2025-02-13
[demikernel] Enhancement: adjust crate versions to match crates.io

anandbonde:
We just updated versions for xdp and dpdk bindings on crates.io, so adjusting the Cargo.toml files to reflect the same for consistency.


kyleholohan (approved):
LGTM

PR #1510, opened 2025-03-12
[inetstack] Enhancement: adding some debug statements

iyzhang:
This PR cleans up log messages in the send and receive path of the UDP and TCP stacks.


anandbonde (approved):
Looks good!

PR #1512, opened 2025-03-13
[inetstack] Fix ack on ack and add log messages

iyzhang:
This PR added changes that keep the inetstack from sending an ack on a bare ack and some additional logging info

### DIFF
--- a/src/rust/inetstack/protocols/layer4/mod.rs
+++ b/src/rust/inetstack/protocols/layer4/mod.rs
@@ -93,7 +93,7 @@ impl Peer {
                     self.receive_batch(batch)
                 }
             },
-            Err(_) => warn!("Could not receive from network interface, continuing ..."),
+            Err(e) => warn!("Could not receive from network interface: {:?}", e),
         }
     }
 
--- a/src/rust/inetstack/protocols/layer4/tcp/established/sender.rs
+++ b/src/rust/inetstack/protocols/layer4/tcp/established/sender.rs
@@ -221,9 +221,11 @@ impl Sender {
         debug_assert!(cb.sender.fin_seq_no.is_none());
 
         // TODO: We need to fix this the correct way: limit our send buffer size to the amount we're willing to buffer.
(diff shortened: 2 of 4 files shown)


src/rust/inetstack/protocols/layer4/mod.rs
```diff
@@ -93,7 +93,9 @@ impl Peer {
                     self.receive_batch(batch)
                 }
             },
-            Err(_) => warn!("Could not receive from network interface, continuing ..."),
+            Err(e) => {
```
anandbonde: We could get rid of the curly braces around this statement.

src/rust/inetstack/protocols/layer4/tcp/established/sender.rs
```diff
@@ -221,16 +221,22 @@ impl Sender {
...
+        trace!("push(): unsent_queue.len() = {:?}", cb.sender.unsent_queue.len());
+
         // Place the buffer in the unsent queue.
         cb.sender.unsent_next_seq_no = cb.sender.unsent_next_seq_no + (buf.len() as u32).into();
         if cb.sender.send_window.get() > 0 {
             Self::send_segment(cb, layer3_endpoint, runtime.get_now(), &mut buf);
         }
         if buf.len() > 0 {
+            if cb.sender.unacked_queue.len() > 0 {
+                trace!("push(): unacked_queue.len() = {:?}", cb.sender.unacked_queue.len());
```
anandbonde: It would be good to differentiate between this and the identical trace msg above. They might optionally show up as consecutive traces in the logs.

anandbonde (approved):
Looks good.

src/rust/inetstack/protocols/layer4/tcp/established/sender.rs
iyzhang: I will change it. For future reference, our traces do print the file and line number if we need to differentiate. But I agree that it could be confusing

PR #1513, opened 2025-03-14
[catpowder] Multiple send and receive buffers

iyzhang:
This PR implements support for multiple send and receive buffers as well as cohosting with reserved ports.


Cargo.toml
```diff
@@ -53,7 +53,7 @@ windows = { version = "0.57.0", features = [
     "Win32_System_SystemInformation",
     "Win32_System_Threading",
 ] }
-demikernel-xdp-bindings = { version = "1.0.1", optional = true }
+demikernel-xdp-bindings = { version = "1.0.1", optional = true, path = "xdp_bindings" }
```
anandbonde: Do you intend to build xdp_bindings from local src code always? If this is just for testing local changes temporarily, then this should be removed so that the correct version of the crate will be used from `crates.io`.

iyzhang: @kyleholohan ? I can fix this either way.

kyleholohan: Yes, please remove. This was temporary. Thanks!

src/rust/demikernel/bindings.rs
```diff
@@ -511,6 +509,7 @@ pub extern "C" fn demi_wait_next_n(
     // Issue wait_any operation.
     let ret: Result<i32, Fail> = do_syscall(|libos| match libos.wait_next_n(wait_callback, duration) {
         Ok(()) => 0,
+        Err(e) if e.errno == libc::ETIMEDOUT => libc::ETIMEDOUT,
```
anandbonde: Is this done to avoid `ETIMEDOUT` to show up as failure in traces? It might be easier to grasp if we move this check inside the `Err(e)` case, but this works too...!

src/rust/demikernel/config.rs
```diff
@@ -82,6 +89,24 @@ mod raw_socket_config {
     // UDP ports for XDP to redirect when cohosting.
     #[cfg(target_os = "windows")]
     pub const XDP_UDP_PORTS: &str = "xdp_udp_ports";
+
+    // The number of ports to reserve in the Windows kernel for use by XDP.
+    #[cfg(target_os = "windows")]
+    pub const XDP_RESERVED_PORT_COUNT: &str = "xdp_reserved_port_count";
+
+    // The number of ports to reserve in the Windows kernel for use by XDP.
```
anandbonde: This comment needs to be corrected.

src/rust/inetstack/protocols/layer4/mod.rs
```diff
@@ -185,8 +186,8 @@ impl Peer {
             Socket::Udp(socket) => self.udp.bind(socket, socket_addr_v4),
         }?;
 
-        if EphemeralPorts::is_private(socket_addr_v4.port()) {
-            self.ephemeral_ports.reserve(socket_addr_v4.port())?;
+        if self.ephemeral_ports.is_private(socket_addr_v4.port()) {
+            let _ = self.ephemeral_ports.reserve(socket_addr_v4.port());
```
anandbonde: If `reserve` returns `None`, will this silently allow it? Is this desirable?

src/rust/inetstack/protocols/layer4/tcp/established/receiver.rs
```diff
@@ -35,7 +35,7 @@ use ::futures::never::Never;
 // TODO: Review this value (and its purpose).  It (16 segments) seems awfully small (would make fast retransmit less
 // useful), and this mechanism isn't the best way to protect ourselves against deliberate out-of-order segment attacks.
 // Ideally, we'd limit out-of-order data to that which (along with the unread data) will fit in the receive window.
-const MAX_OUT_OF_ORDER_SIZE_FRAMES: usize = 16;
+const MAX_OUT_OF_ORDER_SIZE_FRAMES: usize = 1024;
```
anandbonde, line 38: That's a big jump. I'm wondering if we should add a test somehow to exercise this case?

src/rust/inetstack/protocols/layer4/tcp/peer.rs
```diff
@@ -158,10 +158,7 @@ impl SharedTcpPeer {
         {
             // We should panic here because the ephemeral port allocator should not allocate the same port more than
             // once.
-            unreachable!(
-                "There is already a socket listening on this address: {:?} {:?}",
-                local, remote
-            );
+            return Err(Fail::new(libc::EADDRINUSE, "address already in use"));
```
anandbonde, line 160: The above comment needs to be corrected to match the new code (which doesn't panic anymore).

src/rust/runtime/memory/demibuffer.rs
```diff
@@ -508,6 +523,63 @@ impl DemiBuffer {
...
+    }
+
+    /// Returns a flag indicating whether this DemiBuffer is direct, i.e., that this instance is
+    /// not a clone of some other DemiBuffer.
+    pub fn is_direct(&self) -> bool {
+        self.as_metadata().ol_flags & METADATA_F_INDIRECT == 0
+    }
+
+    /// Get the direct DemiBuffer from an indirect DemiBuffer.
+    pub fn into_direct(self) -> DemiBuffer {
```
anandbonde, line 575: Can this be refactored into smaller fn()s?

src/rust/runtime/memory/memory_pool.rs
```diff
@@ -65,10 +65,13 @@ struct PackingIterator<'a> {
 impl MemoryPool {
     /// Create a new empty pool of buffers of the specified size.
     pub fn new(size: NonZeroUsize, align: NonZeroUsize) -> Result<Rc<Self>, LayoutError> {
-        Ok(Rc::new(Self {
+        let me = Rc::new(Self {
```
anandbonde, line 68: At some other places in code, we call a copy/ref to `self` as `this`. It would be good to stick to one naming throughout. Either one is fine.

anandbonde (approved):
Approved.

src/rust/demikernel/bindings.rs
kyleholohan: Yes, this was to avoid the trace.

iyzhang: Is this a situation where the trace was too verbose?

src/rust/inetstack/protocols/layer4/mod.rs
iyzhang: I assume that this is correct.

src/rust/inetstack/protocols/layer4/tcp/established/receiver.rs
iyzhang: We can add a trace but it'll be very long because we need to send 1024 frames out of order. I don't think we need a test because this number is an upper bound and will only be allocated if we have a long sequence of out of order frames.

src/rust/runtime/memory/memory_pool.rs
iyzhang: I've never seen that. Can you let me know where there is an example?

src/rust/runtime/memory/demibuffer.rs
iyzhang: done

PR #1516, opened 2025-03-17
[build script] Feature: add Demikernel build script on Windows and Linux

david-xie-david:
Add Demikernel build script on Windows and Linux

Windows build script: `scripts\windows\build.ps1`

Linux build script: `scripts/linux/build.bash`

README that runs scripts is located at `scripts\windows\README.md` and `scripts/linux/README.md`


tools/scripts/config_template/config.yaml.template
anandbonde, line 1: There are a couple of config.yaml files checked in as reference for Azure and BareMetal. We should try and converge all of these if possible, else we will end up with multiple copies that may go out of sync pretty soon.

anandbonde:
> Add Demikernel build script on Windows and Linux
> 
> Windows build script: `tools\scripts\windows\build.ps1`
> 
> Linux build script: `tools/scripts/linux/build.bash`
> 
> README that runs scripts is located at `tools/scripts/README.md`

Could you write a small explanation as to how these scripts will be used? They seem to be wrappers around the Makefiles and example programs. Are these scripts going to used as a standard in a recurring benchmark/test that we can automate?

tools/scripts/windows/run-client.ps1
```diff
@@ -0,0 +1,30 @@
...
+    Write-Host "Please input the server address."
+    Write-Host "Usage: run-cli.ps1 {server IP address}"
+    exit
+}
+
+$SERVER_HOST = $args[0]
+
+Write-Host " *******        Client Mode         ********************* "
+Write-Host " LIBOS:           $env:LIBOS"
+Write-Host " CONFIG:          $env:CONFIG_PATH"
```
anandbonde: Since this requires setting the `$env:LIBOS` before running this script, it may lead to unknowns to the user. It would be good to add complete, opinionated examples using these scripts which will lead to total clarity as to how to use them.

Otherwise, it's equivalent to using the Make system and running programs with appropriate arguments.

anandbonde (changes requested):

tools/scripts/config_template/config.yaml.template
david-xie-david: Reuse existing config.yaml file as starting point in next commit. Resolving this thread.

tools/scripts/windows/run-client.ps1
david-xie-david: I have added README.md in `scripts/linux` and `scripts/windows` folder that prompts users to adjust env.sh or env.ps1 to adjust LibOS they intend to use. Please let me know if you still have questions. Thanks.

david-xie-david:
> > Add Demikernel build script on Windows and Linux
> > Windows build script: `tools\scripts\windows\build.ps1`
> > Linux build script: `tools/scripts/linux/build.bash`
> > README that runs scripts is located at `tools/scripts/README.md`
> 
> Could you write a small explanation as to how these scripts will be used? They seem to be wrappers around the Makefiles and example programs. Are these scripts going to used as a standard in a recurring benchmark/test that we can automate?

I have added README.md in scripts/linux and scripts/windows folder regarding how to use new build scripts. Please let me know if you still have any concern. Thanks.

tools/scripts/windows/run-client.ps1
anandbonde: We already have steps explaining how to setup and build checked into the `doc` directory. Are those not enough to run these tests? If they are ambitious or incomplete, then we should modify those files instead of creating new documents because they will go out of sync soon. Look at the `.md` files in the `doc` dir. (e.g. https://github.com/microsoft/demikernel/blob/dev/doc/setup.md)

PR #1517, opened 2025-04-07
[inetstack] Bug Fix: Correctly initialize window size seq no

### LINKED ISSUE #1518: [inetstack] Correctly initialize send window and update
## Description

We use the wrong sequence number to initialize the send window. Instead of using the remote sequence number, we use the local sequence number. However, when we later update the send window, we compare this to the remote sequence number in the TCP packet, which does not make sense and is incorrect.

## How to Reproduce

Use the attached trace file.

## Expected Behavior

We should update the send window every time we get a window update in a packet with a higher sequence number.

iyzhang:
Currently, we are using the local sequence number to initialize the window size sequence number in the `Sender` struct. However, we then later compare it to the remote sequence number that comes in the TCP header to make a decisions about whether to update the sender window size. Closes #1518

### DIFF
--- a/network_simulator/input/tcp/close/close-duplicate-fin.pkt
+++ b/network_simulator/input/tcp/close/close-duplicate-fin.pkt
@@ -1,0 +1,46 @@
+// Test for duplicate FIN.
+
+// Establish a connection.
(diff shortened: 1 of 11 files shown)


anandbonde (approved):
Looks good to me.

PR #1521, opened 2025-04-09
[network_simulator] Feature: Start sequence numbers at non-zero

### LINKED ISSUE #1520: [testing] Support different starting sequence numbers
## Context

Currently, we always use 0 as a starting sequence number for both local and remote endpoints in the simulator. This does not check cases where the remote sequence is larger or smaller than the local sequence number.

## Proposed Solution

We start the local end point at a non-zero sequence number based on the local port and we should choose different numbers for the remote sequence number that are bigger or smaller.

iyzhang:
This closes #1520.

It also fixes two bugs that were not caught by our previous tests.
1. Recently we switched to two layers of storing operations in the runtime, one in the wait_next_n call which stores completed tasks and another in wait_any, which stores the operation results. Because we were using the wait_next_n call to run the scheduler in pop_frame, we were not storing any completed operation results. By switching to the wait_any call, we can ensure that the results are stored for later retrieval.
2. We were not correctly detecting duplicate SYN packets because both the local and remote sequence numbers were identical and started at 0. We were using the local sequence number to check instead of the remote one.

### DIFF
--- a/network_simulator/input/tcp/accept/accept-blocking-1.pkt
+++ b/network_simulator/input/tcp/accept/accept-blocking-1.pkt
@@ -7,11 +7,11 @@
 +.2 accept(500, ..., ...) = 0
 
 // Receive SYN packet.
-+.2 TCP < S seq 0(0) win 65535 <mss 1450,wscale 0>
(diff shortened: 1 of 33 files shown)


anandbonde (approved):
Looks good.

PR #1522, opened 2025-04-10
[catpowder] Enhancement: XDP refactor and bug fixes

kyleholohan:
This PR fixes several bugs in the XDP backend related to BufferPool population, specifically w.r.t. XDP-for-Windows' rules for truncation of buffers at the end of the pool. It also adds the ability to send out backend on the VF by default, rather than the default interface. Finally, it refactors and componentizes parts of the backend. The last point should be especially useful for the cohosting component, as it can be reused for the catnip backend.


scripts/config/azure.yaml
```diff
@@ -9,6 +9,10 @@ raw_socket:
   xdp_interface_index: 0
   xdp_cohost_mode: false
   xdp_always_poke_tx: false
+  # If true, will send all packets out on the interface identified by xdp_vf_interface_index;
+  # otherwise, or if xdp_vf_interface_index is not set, will send packets out on the interface
+  # identified by linux_interface_name.
```
anandbonde: We can think of potentially changing the name from `linux_interface_name` to just `interface_name` to avoid naming related confusion on Windows.

src/rust/catpowder/win/observability.rs
```diff
@@ -0,0 +1,234 @@
...
+use crate::{
+    catpowder::win::{api::XdpApi, interface::Interface, socket::XdpSocket},
+    runtime::{fail::Fail, timer::global_get_time},
+};
+
+//=======================================================================================================================
+// Constants
+//======================================================================================================================
+/// The minimum latency between polls before we start worrying about it.
+const MIN_LATENCY_IOTA: u32 = 1000;
```
anandbonde: It might be useful to add the unit of time in this (and related) variable names for clarity.

scripts/config/azure.yaml
kyleholohan: Ah, this comment is wrong -- I pasted the wrong config value

src/rust/catpowder/win/ring/umemreg.rs
```diff
@@ -49,21 +49,59 @@ impl UmemReg {
...
+        let page_size: NonZeroUsize = get_page_size();
+        let real_chunk_size: usize = pool.pool().layout().size();
+
+        // NB when the page size is not evenly divisible by the number of bytes in the last page
+        // of a buffer, BufferPool will pad the buffer to prevent excess page spanning; however,
+        // XDP requires that buffers are contiguous without padding. So here we grow the buffer to
+        // ensure we are not padding.
+        let real_last_page_bytes: usize = real_chunk_size % page_size.get();
+        let (pool, chunk_size): (BufferPool, u16) =
+            if real_chunk_size % pool.pool().layout().align() != 0 || page_size.get() % real_last_page_bytes != 0 {
```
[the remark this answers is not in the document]
kyleholohan: Good catch, Copilot!

scripts/config/azure.yaml
kyleholohan: Updated to fix comment.

src/rust/catpowder/win/observability.rs
kyleholohan: Added units on all bare time-based u32's

iyzhang (approved):
looks good to me. I mostly looked at the changes outside of the catpowder/win directory.

PR #1524, opened 2025-04-10
[runtime] Bug Fix: Correctly check for invalid qtoken

### LINKED ISSUE #1523: [runtime] Correctly check for invalid qtoken
## Description

Currently, we do not have an efficient way to check for invalid qtokens, so we time out on them instead.

## How to Reproduce

1. Run integration tests
2. See that the test for invalid qtokens fails.

## Expected Behavior

We should be able to detect invalid qtokens.

iyzhang:
This closes #1523


anandbonde (approved):
Looks good.

PR #1525, opened 2025-04-10
[testing] Bug Fix: Add synchronization points for determinism

iyzhang:
This PR adds:
1. Additional synchronization points to ensure that tests run in a specific intended order every time. There is still some non-determinism but this should ensure that the tests pass more reliably.
2. A function name to the print of the thread to help with debugging.


anandbonde (approved):
Looks good.

PR #1528, opened 2025-04-15
[build] Enhancement: upgrade nightly build version

anandbonde:
Upgraded the nightly build version and found that some explicit attributes are stable now, so they are no longer needed in the code. So I removed them. There should be no impact on the functionality and performance of the code.

This is the latest version which works fine with Linux Catpowder; further versions don't pass tests. Will investigate and address this if needed in a separate PR.


iyzhang (approved):
Thanks for doing this

PR #1530, opened 2025-04-16
[libos] Fix ordering of request separate from scheduling

### LINKED ISSUE #1526: [inetstack] TCP packets sent out of order
## Description

Currently if we push multiple packets and wait on them, they do not appear in order on the other end in the order in which they were pushed.

## How to Reproduce

1. Push multiple SGAs in a row
2. Wait on the pushes to complete
3. See that the SGAs do not appear on the other end in sequence.

## Expected Behavior

SGAs should be delivered reliably and in the order in which they were pushed on the other end of the TCP connection.

iyzhang:
This PR closes #1526. It includes the fix in one commit and a test to check for correctness in another commit. In order to reproduce the issue, run the test without the fix on dev.

### DIFF
--- a/examples/rust/tcp-push-pop.rs
+++ b/examples/rust/tcp-push-pop.rs
@@ -3,7 +3,8 @@
 
 /// This test exercises the following behavior: A server and client pair where the client is only sending data and the
 /// server is only receiving. We test this behavior because we want to make sure that the server correctly acknowledges
-/// the sent data, even though there is no data flowing the other direction.
+/// the sent data, even though there is no data flowing the other direction. We also fill the buffer with a sequence
+/// number on each iteration to be sure that packets are flowing
 //======================================================================================================================
 // Imports
 //======================================================================================================================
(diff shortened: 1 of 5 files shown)


examples/rust/tcp-push-pop.rs
```diff
@@ -144,24 +120,50 @@ impl TcpServer {
...
+        self.accepted_sockqd = match self.libos.wait(qt, Some(TIMEOUT_SECONDS)) {
+            Ok(qr) if qr.qr_opcode == demi_opcode_t::DEMI_OPC_ACCEPT => unsafe { Some(qr.qr_value.ares.qd.into()) },
+            Ok(qr) if qr.qr_opcode == demi_opcode_t::DEMI_OPC_FAILED => anyhow::bail!("accept failed: {}", qr.qr_ret),
+            Ok(qr) => anyhow::bail!("unexpected opcode: {:?}", qr.qr_opcode),
+            Err(e) => anyhow::bail!("operation failed: {:?}", e.cause),
+        };
+
+        // Wait for blocking pops.
+        self.pop_rounds()?;
+        // Wait for non-blocking pops.
```
anandbonde, line 163: Is this ordering implicit - how is it ensured? i.e., the first time `pop_rounds` is called, it will wait for all blocking pops and next time for all the non-blocking pops?

examples/rust/tcp-push-pop.rs
```diff
@@ -28,34 +29,36 @@ pub const SOCK_STREAM: i32 = libc::SOCK_STREAM;
...
-fn mksga(libos: &mut LibOS, size: usize, value: u8) -> Result<demi_sgarray_t> {
-    let sga: demi_sgarray_t = match libos.sgaalloc(size) {
+fn mksga(libos: &mut LibOS, value: u8) -> Result<demi_sgarray_t> {
+    let sga: demi_sgarray_t = match libos.sgaalloc(BUF_SIZE_BYTES) {
         Ok(sga) => sga,
         Err(e) => anyhow::bail!("failed to allocate scatter-gather array: {:?}", e),
     };
 
+    // Create pointer for filling the array.
+
```
anandbonde: This is probably a left over blank line, it can be deleted.

examples/rust/tcp-push-pop.rs
```diff
@@ -28,34 +29,36 @@ pub const SOCK_STREAM: i32 = libc::SOCK_STREAM;
...
+    let ptr: *mut u8 = sga.sga_segs[0].sgaseg_buf as *mut u8;
     // Ensure that allocated array has the requested size.
-    if sga.sga_segs[0].sgaseg_len as usize != size {
+    if sga.sga_segs[0].sgaseg_len as usize != BUF_SIZE_BYTES || ptr.is_null() || !ptr.is_aligned() {
```
anandbonde: If any of these conditions are met, it possibly represents sga corruption. Would this signify another serious underlying problem? Or could it just be a faulty send from the other side?

anandbonde (approved):
Looks good, once the comments are addressed.

examples/rust/tcp-push-pop.rs
iyzhang: I don't think this is a faulty send since this is a new allocated sga array. I took out the pointer alignment check since it must be aligned to bytes.

examples/rust/tcp-push-pop.rs
iyzhang: Thanks!

examples/rust/tcp-push-pop.rs
iyzhang: The order is imposed by the TCP connection because we're pushing all of the blocking requests and then the non-blocking ones.

PR #1531, opened 2025-04-18
[testing] Bug Fix: Make UDP tests more reliable

iyzhang:
This adds some more retries to our UDP tests so that we can lose more packets and still finish.

### DIFF
--- a/examples/rust/udp-ping-pong.rs
+++ b/examples/rust/udp-ping-pong.rs
@@ -6,11 +6,7 @@
 //======================================================================================================================
 
 use ::anyhow::Result;
-use ::demikernel::{
-    demi_sgarray_t,
-    runtime::types::{demi_opcode_t, demi_qresult_t},
-    LibOS, LibOSName, QDesc, QToken,
-};
+use ::demikernel::{demi_sgarray_t, runtime::types::demi_opcode_t, LibOS, LibOSName, QDesc, QToken};
 use ::std::{env, net::SocketAddr, slice, str::FromStr, time::Duration};
 
 #[cfg(target_os = "windows")]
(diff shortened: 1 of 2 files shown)


examples/rust/udp-ping-pong.rs
```diff
@@ -185,96 +166,55 @@ impl UdpClient {
...
+                let qt: QToken = match self.libos.pop(self.sockqd, None) {
+                    Ok(qt) => qt,
+                    Err(e) => anyhow::bail!("pop failed: {:?}", e),
+                };
+
+                match self.libos.wait(qt, Some(RETRY_TIMEOUT_SECONDS)) {
+                    Ok(qr) if qr.qr_opcode == demi_opcode_t::DEMI_OPC_POP => break unsafe { qr.qr_value.sga },
+                    Ok(_) => anyhow::bail!("unexpected result"),
+                    // Retry if we didn't receive a response in a second.
+                    Err(e) if e.errno == libc::ETIMEDOUT => continue,
```
anandbonde: Would you like to track this case by tracing? It might be useful to figure out in debug mode whether this is happening at all and how many times.

examples/rust/udp-push-pop.rs
```diff
@@ -91,7 +91,8 @@ impl UdpServer {
     }
 
     fn run(&mut self, local: SocketAddr, fill_char: u8, num_sends: usize) -> Result<()> {
-        let num_receives: usize = (8 * num_sends) / 128;
+        // Assume that if we send twice as many packets, most of them will get through.
```
anandbonde: Do you mean to say that if we send out `half` the number of packets, we can assume all will get through?

anandbonde (approved):
Looks good.

examples/rust/udp-ping-pong.rs
iyzhang: Sure, will add

examples/rust/udp-push-pop.rs
iyzhang: Sorry, we send out num_send and then expect half of of them to be received. I'll clarify.

PR #1532, opened 2025-04-25
[script] Bug Fix: remove unnecessary pip install

anandbonde:
The current CI workflows don't write to Azure Tables, so the Python package that supports this integration is no longer needed.

Moreover, this causes warnings on manual triggers of this script on machines which don't use Python Virtual Environments.


iyzhang (approved):
thanks!

PR #1533, opened 2025-04-25
[bindings] Bug Fix: tweak and fix minor behavior in headers/bindings

kyleholohan:
This PR fixes and tweaks a few things:
- Fix attributes and SAL annotations and getpeername and getsockopt
- Add a safety `#pragma pack` to types.h so we don't accidentally forget to restore packing for C clients
- Fix error message wording to indicate the correct method
- Stop tracing when a wait times out; this is not really an error condition
- Be more permissive in allowing the client to pass overlarge buffers to getpeername

### DIFF
--- a/include/demi/cc.h
+++ b/include/demi/cc.h
@@ -12,6 +12,7 @@
 #define _In_opt_
 #define _In_reads_(s)
 #define _In_reads_bytes_(b)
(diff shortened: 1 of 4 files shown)


iyzhang (approved):
looks good to me

PR #1534, opened 2025-04-25
[libos] Bug Fix: fix bug in send queue for pushto operations

kyleholohan:
This PR fixes a bug in `pushto` operations: the destination address and packet are no longer correlated after the addition of the send_queue field to SharedNetworkQueue. This change puts the destination address in the send_queue along with the DemiBuffer so that they flow together.


iyzhang (approved):
lgtm

PR #1536, opened 2025-04-28
[testing] Enhancement: UDP barriers for more consistent testing

iyzhang:
This PR adds a few more barrier points to the UDP tests. These are less necessary for the UDP tests as there is no connection set up and tear down but I added these just for symmetry with the TCP tests.


anandbonde (approved):
Looks good.

Issue #1537, opened 2025-05-01
[inetstack] Simultaneous close does not continue sending FIN


iyzhang:
## Description

The inetstack implementation and the simulator trace test for simultaneous close were both incorrect. In particular, they resent the FIN packet without the FIN bit set while in the Closing state, causing the other end to hang in a simultaneous close scenario.

## How to Reproduce

After adding synchronization points for our TCP unit tests, the tcp_bad_* tests were hitting the simultaneous close scenario frequently causing them to fail.

## Expected Behavior

If both end of a TCP connection independently call close, the connection should be shut down gracefully in a symmetric manner.

## Related Issues

#1529

PR #1538, opened 2025-05-01
[inetstack] Bug Fix: Simultaneous close misses FIN resend

### LINKED ISSUE #1529: Test tcp-bad-close is non-deterministic on catnap/windows
## Description

The tcp-bad-close test is non-deterministic on catnap/windows.

## How to Reproduce

1. Submit PR test run
2. Repeat # 1 until failure is encountered.

## Screenshots

See [this test run](https://github.com/microsoft/demikernel/actions/runs/14391164945/attempts/1) for an example.

## Expected Behavior

Determinism

iyzhang:
This PR closes #1529 and #1537

### DIFF
--- a/src/rust/inetstack/protocols/layer4/tcp/established/receiver.rs
+++ b/src/rust/inetstack/protocols/layer4/tcp/established/receiver.rs
@@ -111,94 +111,6 @@ impl Receiver {
         }
     }
 
-    // This function causes a EOF to be returned to the user. We also know that there will be no more incoming
-    // data after this sequence number.
-    fn push_fin(&mut self) {
-        self.pop_queue.push(DemiBuffer::new(0));
-        debug_assert_eq!(self.receive_next_seq_no, self.fin_seq_no.get().unwrap());
-        // Reset it to wake up any close coroutines waiting for FIN to arrive.
-        self.fin_seq_no.set(Some(self.receive_next_seq_no));
-        // Move RECV_NXT over the FIN.
-        self.receive_next_seq_no = self.receive_next_seq_no + 1.into();
-    }
-
-    pub fn get_receive_window_size(&self) -> u32 {
-        let bytes_unread: u32 = (self.receive_next_seq_no - self.reader_next_seq_no).into();
-        // The window should be less than 1GB or 64KB without scaling.
-        debug_assert!(
-            (self.window_scale_shift_bits == 0 && bytes_unread <= MAX_WINDOW_SIZE_WITHOUT_SCALING)
-                || bytes_unread <= MAX_WINDOW_SIZE_WITH_SCALING
-        );
-        debug!(
-            "Receive window size: bytes_unread={:?} buffer_size_bytes={:?} ",
-            bytes_unread, self.buffer_size_bytes
-        );
-        self.buffer_size_bytes - bytes_unread
-    }
-
-    pub fn hdr_window_size(&self) -> u16 {
-        let window_size: u32 = self.get_receive_window_size();
-        let hdr_window_size: u16 = expect_ok!(
-            (window_size >> self.window_scale_shift_bits).try_into(),
-            "Window size overflow"
-        );
-        debug!(
-            "Window size -> {} (hdr {}, scale {})",
-            (hdr_window_size as u32) << self.window_scale_shift_bits,
-            hdr_window_size,
-            self.window_scale_shift_bits,
-        );
-        hdr_window_size
-    }
-
-    // This routine takes an incoming in-order TCP segment and adds the data to the user's receive queue.  If the new
-    // segment fills a "hole" in the receive sequence number space allowing previously stored out-of-order data to now
-    // be received, it receives that too.
-    //
-    // This routine also updates receive_next to reflect any data now considered "received".
-    fn receive_data(&mut self, seg_start: SeqNumber, buf: DemiBuffer) {
-        // This routine should only be called with in-order segment data.
-        debug_assert_eq!(seg_start, self.receive_next_seq_no);
-
-        // Push the new segment data onto the end of the receive queue.
-        self.receive_next_seq_no = self.receive_next_seq_no + SeqNumber::from(buf.len() as u32);
-        // This inserts the segment and wakes a waiting pop coroutine.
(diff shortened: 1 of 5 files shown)


src/rust/inetstack/protocols/layer4/tcp/established/receiver.rs
```diff
@@ -505,40 +534,42 @@ impl Receiver {
...
-            match cb.state {
-                State::Established | State::FinWait1 | State::FinWait2 => {
-                    debug_assert_eq!(seg_len, data.len() as u32);
-                    cb.receiver.store_out_of_order_segment(seg_start, seg_end, data);
-                    // Sending an ACK here is only a "MAY" according to the RFCs, but helpful for fast retransmit.
-                    trace!("process_data(): send ack on out-of-order segment");
-                    Sender::send_ack(cb, layer3_endpoint);
-                },
-                state => warn!("Ignoring data received after FIN (in state {:?}).", state),
+        if data.len() > 0 {
```
anandbonde: We should return early by flipping the condition. This will reduce the level of nesting in this function as well.

src/rust/inetstack/protocols/layer4/tcp/established/receiver.rs
```diff
@@ -274,60 +186,166 @@ impl Receiver {
...
 
-        let has_data: bool = if data.len() > 0 {
-            Self::process_data(cb, layer3_endpoint, data, seg_start, seg_end, seg_len)?;
-            true
-        } else {
-            false
-        };
+        // Store whether the packet has data here because processing it will consume the DemiBuffer.
+        let has_data: bool = data.len() > 0;
+        Self::process_data_if_any(cb, layer3_endpoint, data, seg_start, seg_end, seg_len)?;
```
anandbonde: Can this call be avoided by checking `has_data` above?

src/rust/inetstack/protocols/layer4/tcp/established/receiver.rs
```diff
@@ -274,60 +186,166 @@ impl Receiver {
...
+    // This function causes a EOF to be returned to the user. We also know that there will be no more incoming
+    // data after this sequence number.
+    fn check_and_process_fin(
+        cb: &mut ControlBlock,
+        header: &TcpHeader,
+        seg_end: SeqNumber,
+        layer3_endpoint: &mut SharedLayer3Endpoint,
+        now: Instant,
+    ) -> Result<(), Fail> {
+        // Is this a FIN packet?
```
anandbonde: This comment can be removed as the next line is unambiguous.

src/rust/inetstack/protocols/layer4/tcp/established/receiver.rs
```diff
@@ -274,60 +186,166 @@ impl Receiver {
...
-            // We already owe our peer an ACK (the timer was already running), so cancel the timer and ACK now.
-            cb.receiver.ack_deadline_time_secs.set(None);
-            trace!("process_packet(): sending ack before deadline because another packet arrived");
-            Sender::send_ack(cb, layer3_endpoint);
+        Ok(())
+    }
+
+    pub fn get_receive_window_size(&self) -> u32 {
+        let bytes_unread: u32 = (self.receive_next_seq_no - self.reader_next_seq_no).into();
+        // The window should be less than 1GB or 64KB without scaling.
```
anandbonde, line 270: It would be good to not assume these absolute sizes in the comments. These are susceptible to going stale in case the values of constants change in future.

src/rust/inetstack/protocols/layer4/tcp/established/receiver.rs
```diff
@@ -274,60 +186,166 @@ impl Receiver {
...
+                if stored_entry.0 == self.receive_next_seq_no {
+                    // Move this entry's buffer from the out-of-order store to the receive queue.
+                    // This data is now considered to be "received" by TCP, and included in our RCV.NXT calculation.
+                    debug!("Recovering out-of-order packet at {}", self.receive_next_seq_no);
+                    if let Some(temp) = self.out_of_order_frames.pop_front() {
+                        self.receive_next_seq_no = self.receive_next_seq_no + SeqNumber::from(temp.1.len() as u32);
+                        // This inserts the segment and wakes a waiting pop coroutine.
+                        self.pop_queue.push(temp.1);
+                    }
+                } else {
```
anandbonde: This conditional can be inverted to reduce nesting. The indentation level is pretty high in this function. Along with that, it can be broken down into smaller functions if the indentation/nesting still remains high.

src/rust/inetstack/protocols/layer4/tcp/established/receiver.rs
iyzhang: Oh, these are TCP constants, so won't change unless something big changes in the RFC.

src/rust/inetstack/protocols/layer4/tcp/established/receiver.rs
iyzhang: Done! So much better!

src/rust/inetstack/protocols/layer4/tcp/established/receiver.rs
iyzhang: No, the data gets adjusted by the check_segment_in_window function. Given that each processing function might modify the buf, I thought doing a late binding was the best idea.

anandbonde: The whole body of `process_data_if_any` is wrapped in the `data.len() > 0` check. So it will be enough to just check for `has_data` before calling it instead. Also, this function is called just from here as of now.

anandbonde (approved):

src/rust/inetstack/protocols/layer4/tcp/established/receiver.rs
iyzhang: oh, I'm storing the data for later in the function below process_data which takes ownership of data, so I can't check it later.

anandbonde: Thanks for changing. The code is more readable now.

PR #1540, opened 2025-05-07
[testing] Bug Fix: Stability fixes to tcp tests

iyzhang:
This PR adds two changes for more reliable tcp testing. 
1. Increase the time out from 100ms to 5 seconds. Occasionally cargo seems to take a long time to start up one of the threads and there is no reason to time out early, so now when we get time outs they are definitely failures
2. Increase sync points. Previously tests sometimes failed because the client tried to connect before the listening socket was created or listening. Normally, this would not be a problem because the connecting socket would retry but we don't want to wait for the time out in this case (we have other tests that test this).


anandbonde (approved):
Looks good.

PR #1542, opened 2025-05-08
[runtime] Feature: remove memory runtime

iyzhang:
This PR removes most of the memory runtime abstraction as it is redundant. Instead it replaces it with a single function that allocates a DemiBuffer in whatever way is required by the underlying runtime (e.g., DPDK, raw sockets, XDP, etc.). This removes most of the redundant code across our memory runtimes, making it easier to make changes.


src/rust/catnip/runtime/mod.rs
```diff
@@ -330,7 +337,12 @@ impl PhysicalLayer for SharedDPDKRuntime {
         let outgoing_pkt: DemiBuffer = match pkt {
             buf if buf.is_dpdk_allocated() => buf,
             buf => {
-                let mut mbuf: DemiBuffer = self.mm.alloc_body_mbuf().expect("should be able to allocate mbuf");
+                let mut mbuf: DemiBuffer = {
+                    // Allocate a DPDK-managed buffer.
+                    let mbuf_ptr: *mut rte_mbuf = self.mem_pool.alloc_mbuf(None)?;
+                    // Safety: `mbuf_ptr` is a valid pointer to a properly initialized `rte_mbuf` struct.
+                    unsafe { DemiBuffer::from_mbuf(mbuf_ptr) }
+                };
```
anandbonde: We should create a fn for allocating a DPDK-managed buffer to avoid repetition.

src/rust/catpowder/linux/mod.rs
```diff
@@ -69,40 +67,12 @@ impl LinuxRuntime {
...
-
-        // Create and return a new scatter-gather array (which inherits the DemiBuffer's reference).
-        Ok(demi_sgarray_t {
-            sga_buf: buf.into_raw().as_ptr() as *mut c_void,
-            sga_numsegs: 1,
-            sga_segs: [sga_seg],
-            sga_addr: unsafe { mem::zeroed() },
-        })
+impl DemiMemoryAllocator for LinuxRuntime {
+    fn allocate_demi_buffer(&self, size: usize) -> Result<DemiBuffer, Fail> {
```
anandbonde, line 74: Do we need a size check here like SharedDPDKRuntime? Also at other points where a fn with this name exists?

src/rust/catpowder/win/runtime.rs
```diff
@@ -197,24 +195,10 @@ impl PhysicalLayer for SharedCatpowderRuntime {
 }
 
 /// Memory runtime trait implementation for XDP Runtime.
-impl MemoryRuntime for SharedCatpowderRuntime {
+impl DemiMemoryAllocator for SharedCatpowderRuntime {
     /// Allocates a scatter-gather array.
-    fn sgaalloc(&self, size: usize) -> Result<demi_sgarray_t, Fail> {
+    fn allocate_demi_buffer(&self, size: usize) -> Result<DemiBuffer, Fail> {
```
anandbonde, line 200: Do we need a size check here?

src/rust/runtime/memory/mod.rs
```diff
@@ -29,143 +29,130 @@ pub use self::{buffer_pool::*, demibuffer::*};
...
-
-            // An adjustment above would have reduced clone.len() by the adjustment amount.
-            clone_len -= adjustment_amount;
-            debug_assert_eq!(clone_len, clone.len());
-
-            // Trim the clone down to size.
-            let trim_amount: usize = clone_len - sga_len;
-            clone.trim(trim_amount)?;
+/// Clones a scatter-gather array.
+pub fn clone_sgarray(sga: &demi_sgarray_t) -> Result<DemiBuffer, Fail> {
```
anandbonde, line 103: This fn is long, can it be broken into smaller chunks? The mechanisms to clone and the checks can be separated into distinct fns.

anandbonde (changes requested):

anandbonde (approved):
Looks good.

PR #1543, opened 2025-05-16
[catpowder] Enhancement: remove queue count computation and traces

anandbonde:
These traces were possibly just remnants of earlier debugging. Also, the queue count is just maintained for the traces, so removed both.


kyleholohan (approved):
Agree this seems unnecessary. Thanks for removing.

PR #1544, opened 2025-05-17
[catnip] Enhancement: remove unnecessary assert

anandbonde:
This assert ensures that the DPDK api is doing what is promises. This is not something that Demikernel should be confirming, it's the responsibility of the DPDK tests.


anandbonde:
Irene will address this in a separate PR, so closing this PR.

PR #1548, opened 2025-05-20
[ci] Bug Fix: correct checkout sequence to get latest code

anandbonde:
The checkout sequence was incorrect because the pull was only applied to the current branch before switching to the target. This problem was clearly visible in case of reused branch names.

Also, the Windows cleanup did not correctly checkout the target branch name, so correct that.


iyzhang (approved):
Thanks for finding these!

PR #1550, opened 2025-05-28
[runtime] Feature: memory segments and batching

### LINKED ISSUE #88: [pdpix] Support Scatter-Gather Arrays with Multiple Segments
Description
========

Demikernel currently has the limitation of single-segment scatter-gather arrays:

```
#define DMTR_SGARRAY_MAXSIZE 1

...

typedef struct dmtr_sgarray {
    void *sga_buf;
    uint32_t sga_numsegs;
    dmtr_sgaseg_t sga_segs[DMTR_SGARRAY_MAXSIZE];
    // todo: to be removed when LWIP libOS is retired in favor of
    // `dpdk-catnip`.
    struct sockaddr_in sga_addr;
} dmtr_sgarray_t;
```

We should support the creation of scatter-gatther arrays with multiple segments.

iyzhang:
This PR closes #88

### DIFF
--- a/include/demi/types.h
+++ b/include/demi/types.h
@@ -29,7 +29,7 @@ extern "C"
 /**
  * @brief Maximum number of segments in a scatter-gather array.
  */
-#define DEMI_SGARRAY_MAXSIZE 1
+#define DEMI_SGARRAY_MAXSIZE 20
 
     /**
      * @brief An I/O queue token.
@@ -47,8 +47,9 @@ extern "C"
         typedef struct __attribute__((__packed__)) demi_sgaseg
 #endif
     {
-        void *sgaseg_buf;    /**< Underlying data.       */
-        uint32_t sgaseg_len; /**< Size in bytes of data. */
+        void *sgaseg_md;         /**< Reserved for Demikernel metadata.       */
+        void *data_buf_ptr;      /**< Underlying data.       */
+        uint32_t data_len_bytes; /**< Size in bytes of data. */
     } demi_sgaseg_t;
 #ifdef _WIN32
 #pragma pack(pop)
@@ -65,9 +66,8 @@ extern "C"
         typedef struct __attribute__((__packed__)) demi_sgarray
 #endif
     {
-        void *sga_buf;                                /**< Reserved.                                       */
         uint32_t sga_numsegs;                         /**< Number of segments in the scatter-gather array. */
-        demi_sgaseg_t sga_segs[DEMI_SGARRAY_MAXSIZE]; /**< Scatter-gather array segments.                  */
+        demi_sgaseg_t segments[DEMI_SGARRAY_MAXSIZE]; /**< Scatter-gather array segments.                  */
         struct sockaddr_in sga_addr;                  /**< Source address of scatter-gather array.         */
     } demi_sgarray_t;
 #ifdef _WIN32
@@ -148,8 +148,9 @@ extern "C"
         DemiLogLevel_Trace = 5,
     } demi_log_level_t;
 
-    // Logging callback. Arguments are: level, module name, module length, file name, file name length, line number, message, message length, 
-    typedef void (*demi_log_callback_t)(demi_log_level_t, const char*, uint32_t, const char*, uint32_t, uint32_t, const char*, uint32_t);
+    // Logging callback. Demikernel will call an external C function on each logged stat for inclusion into a
+    // performance tracking system.
+    typedef void (*demi_log_callback_t)(demi_log_level_t log_level, const char *module_name, uint32_t module_name_len_bytes, const char *file_name, uint32_t file_name_len_bytes, uint32_t line_number, const char *message, uint32_t message_len_bytes);
 
 /**
  * @brief Arguments for Demikernel.
--- a/src/catnap/linux/transport.rs
+++ b/src/catnap/linux/transport.rs
@@ -16,7 +16,7 @@ mod socket;
 use crate::{
     catnap::transport::socket::{SharedSocketData, SocketData},
     demikernel::config::Config,
-    expect_ok, expect_some,
+    expect_some,
     runtime::{
         fail::Fail,
         memory::{DemiBuffer, DemiMemoryAllocator},
@@ -24,9 +24,12 @@ use crate::{
             socket::option::{SocketOption, TcpSocketOptions},
             transport::NetworkTransport,
         },
-        poll_yield, DemiRuntime, SharedDemiRuntime, SharedObject,
+        poll_yield,
+        types::DEMI_SGARRAY_MAXLEN,
+        DemiRuntime, SharedDemiRuntime, SharedObject,
     },
 };
+use ::arrayvec::ArrayVec;
 use ::futures::FutureExt;
 use ::slab::Slab;
 use ::socket2::{Domain, Protocol, Socket, Type};
@@ -35,6 +38,7 @@ use ::std::{
     net::{Shutdown, SocketAddr, SocketAddrV4},
     ops::{Deref, DerefMut},
     os::fd::{AsRawFd, RawFd},
+    time::Duration,
 };
 
 //======================================================================================================================
@@ -530,7 +534,7 @@ impl NetworkTransport for SharedCatnapTransport {
                         libc::ENOTCONN => break,
                         errno if DemiRuntime::should_retry(errno) => {
                             // Wait for a new incoming event.
-                            data.pop(0).await?;
+                            data.pop(0, None).await?;
                             continue;
                         },
                         errno => return Err(Fail::new(errno, "operation failed")),
@@ -553,12 +557,12 @@ impl NetworkTransport for SharedCatnapTransport {
     async fn push(
         &mut self,
         sd: &mut Self::SocketDescriptor,
-        buffer: &mut DemiBuffer,
+        bufs: ArrayVec<DemiBuffer, DEMI_SGARRAY_MAXLEN>,
         addr: Option<SocketAddr>,
     ) -> Result<(), Fail> {
-        self.data_from_sd(sd).push(addr, buffer.clone()).await?;
-        // Clear out the original buffer.
-        expect_ok!(buffer.trim(buffer.len()), "Should be able to empty the buffer");
+        for buf in bufs {
+            self.data_from_sd(sd).push(addr, buf).await?;
+        }
         Ok(())
     }
 
@@ -569,8 +573,32 @@ impl NetworkTransport for SharedCatnapTransport {
         &mut self,
         sd: &mut Self::SocketDescriptor,
         size: usize,
-    ) -> Result<(Option<SocketAddr>, DemiBuffer), Fail> {
-        self.data_from_sd(sd).pop(size).await
+    ) -> Result<(Option<SocketAddr>, ArrayVec<DemiBuffer, DEMI_SGARRAY_MAXLEN>), Fail> {
+        let mut total_size = 0;
+        let mut bufs: ArrayVec<DemiBuffer, DEMI_SGARRAY_MAXLEN> = ArrayVec::new();
+        let mut addr: Option<SocketAddr> = None;
+        while total_size < size && bufs.len() < DEMI_SGARRAY_MAXLEN - 1 {
+            let (src_addr, buf) = if !bufs.is_empty() {
+                match self.data_from_sd(sd).pop(size - total_size, Some(Duration::ZERO)).await {
+                    Ok(result) => result,
+                    Err(e) if e.errno == libc::ETIMEDOUT => break,
+                    Err(e) => return Err(e),
+                }
+            } else {
+                self.data_from_sd(sd).pop(size - total_size, None).await?
+            };
+            match addr {
+                None => addr = src_addr,
+                addr if src_addr != addr => {
+                    self.data_from_sd(sd).push_front(buf, src_addr);
+                    return Ok((addr, bufs));
+                },
+                _ => (),
+            }
+            total_size += buf.len();
+            bufs.push(buf)
+        }
+        Ok((addr, bufs))
     }
 
     /// Close the socket on the underlying transport. Also unregisters the socket with epoll.
--- a/src/demikernel/libos/network/libos.rs
+++ b/src/demikernel/libos/network/libos.rs
@@ -19,11 +19,14 @@ use crate::{
             unwrap_socketaddr,
         },
         queue::{downcast_queue, IoQueue, OperationResult},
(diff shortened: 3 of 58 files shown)


src/runtime/memory/mod.rs
```diff
@@ -27,111 +28,128 @@ pub use self::{buffer_pool::*, demibuffer::*};
...
-    let data: *const u8 = buf.as_ptr();
-    let sga_seg: demi_sgaseg_t = demi_sgaseg_t {
-        sgaseg_buf: data as *mut c_void,
-        sgaseg_len: buf.len() as u32,
-    };
+/// Converts a list of DemiBuffers into a scatter-gather array.
+pub fn into_sgarray(bufs: ArrayVec<DemiBuffer, DEMI_SGARRAY_MAXLEN>) -> Result<demi_sgarray_t, Fail> {
+    // Check the sizes before allocating anything.
+    if bufs.is_empty() {
+        let cause: String = format!("cannot allocate a zero element scatter-gather array");
```
anandbonde, line 44: A string literal can be used here instead of `format!` to make it more efficient and readable, also there will be no need to take the reference to `cause` with that change. The block can be written like this:

```suggestion
        let cause: &'static str = "cannot allocate a zero element scatter-gather array";
        error!("into_sgarray(): {}", cause);
        return Err(Fail::new(libc::EINVAL, cause));
```

src/inetstack/protocols/layer2/mod.rs
```diff
@@ -77,23 +77,29 @@ impl SharedLayer2Endpoint {
...
         eth2_type: EtherType2,
-        mut pkt: DemiBuffer,
+        mut pkts: ArrayVec<DemiBuffer, MAX_BATCH_SIZE_NUM_PACKETS>,
     ) -> Result<(), Fail> {
         let eth2_header: Ethernet2Header = Ethernet2Header::new(remote_link_addr, self.local_link_addr, eth2_type);
         debug!("L2 OUTGOING {:?}", eth2_header);
-        eth2_header.serialize_and_attach(&mut pkt);
-        self.layer1_endpoint.transmit(pkt)
+        for mut pkt in pkts.iter_mut() {
+            eth2_header.serialize_and_attach(&mut pkt);
```
anandbonde, line 100: `pkt` is already a muatble reference, so there is no need to take a reference again. You can just use the code as below.

```suggestion
            eth2_header.serialize_and_attach(pkt);
```

src/inetstack/protocols/layer2/mod.rs
```diff
@@ -77,23 +77,29 @@ impl SharedLayer2Endpoint {
...
+        }
```
anandbonde: `pkt` is already a mutable reference, so there is no need to take a reference again. You can just use the code as below.

```suggestion
        for pkt in pkts.iter_mut() {
            eth2_header.serialize_and_attach(pkt);
        }
```

src/runtime/memory/mod.rs
```diff
@@ -27,111 +28,128 @@ pub use self::{buffer_pool::*, demibuffer::*};
...
+        error!("into_sgarray(): {}", cause);
+        return Err(Fail::new(libc::EINVAL, &cause));
```
anandbonde: A string literal can be used here instead of `format!` to make it more efficient and readable, also there will be no need to take the reference to `cause` with that change. The block can be written like this:

```suggestion
        let cause: &'static str = format!("cannot allocate a zero element scatter-gather array");
        error!("into_sgarray(): {}", cause);
        return Err(Fail::new(libc::EINVAL, cause));
```

_Note: this applies to any other places as well in this change which use string literals in format._

src/inetstack/protocols/layer4/tcp/established/receiver.rs
```diff
@@ -112,31 +116,43 @@ impl Receiver {
...
+    ) -> Result<ArrayVec<DemiBuffer, MAX_BATCH_SIZE_NUM_PACKETS>, Fail> {
+        let mut bufs: ArrayVec<DemiBuffer, MAX_BATCH_SIZE_NUM_PACKETS> = ArrayVec::new();
+        let mut buf: DemiBuffer = self.pop_queue.pop(None).await?;
+        loop {
+            if let Some(size) = size.as_mut() {
+                if buf.len() > *size {
+                    let remaining_buf: DemiBuffer = buf.split_front(*size)?;
+                    self.pop_queue.push_front(remaining_buf);
+                }
+                *size = *size - buf.len();
```
anandbonde: This can be shortened to
```suggestion
                *size -= buf.len();
```

src/inetstack/protocols/layer4/tcp/established/receiver.rs
```diff
@@ -112,31 +116,43 @@ impl Receiver {
...
-            _ => {
-                debug!("found FIN");
-                self.reader_next_seq_no = self.reader_next_seq_no + 1.into();
-            },
+                    break;
+                },
+            }
+            bufs.push(buf);
+            match size {
+                Some(size) if size == 0 => break,
```
anandbonde: You can simply use this representation instead to avoid specifying the condition
```suggestion
                Some(0) => break,
```

src/inetstack/mod.rs
```diff
@@ -268,10 +270,12 @@ impl NetworkTransport for SharedInetStack {
     async fn push(
         &mut self,
         sd: &mut Self::SocketDescriptor,
-        buf: &mut DemiBuffer,
+        bufs: ArrayVec<DemiBuffer, DEMI_SGARRAY_MAXLEN>,
         addr: Option<SocketAddr>,
     ) -> Result<(), Fail> {
-        self.layer4_endpoint.push(sd, buf, addr).await
+        debug_assert!(DEMI_SGARRAY_MAXLEN >= MAX_BATCH_SIZE_NUM_PACKETS);
```
anandbonde: Can this be done outside the critical path (even though it's only compiled in for debug mode)? Both operands are consts, so this should be checked just once, if possible.

src/inetstack/mod.rs
```diff
@@ -268,10 +270,12 @@ impl NetworkTransport for SharedInetStack {
...
+        let push_bufs: ArrayVec<DemiBuffer, MAX_BATCH_SIZE_NUM_PACKETS> = bufs.into();
```
anandbonde: This `.into()` is not needed because the types to and from are the same. Then, the question arises, do we need `push_bufs` at all? Or can we just pass in `bufs` into `layer4_endpoint.push()`?

src/catnap/linux/transport.rs
```diff
@@ -574,8 +577,23 @@ impl NetworkTransport for SharedCatnapTransport {
         &mut self,
         sd: &mut Self::SocketDescriptor,
         size: usize,
-    ) -> Result<(Option<SocketAddr>, DemiBuffer), Fail> {
-        self.data_from_sd(sd).pop(size).await
+    ) -> Result<(Option<SocketAddr>, ArrayVec<DemiBuffer, DEMI_SGARRAY_MAXLEN>), Fail> {
+        let total_size: usize = 0;
```
anandbonde: What's the role of `total_size` in the while loop? It's value never changes.

src/demikernel/libos/network/libos.rs
```diff
@@ -300,16 +303,23 @@ impl<T: NetworkTransport> SharedNetworkLibOS<T> {
     /// coroutine that asynchronously runs the push and any synchronous multi-queue functionality before the push
     /// begins.
     pub fn push(&mut self, qd: QDesc, sga: &demi_sgarray_t) -> Result<QToken, Fail> {
-        let buf: DemiBuffer = clone_sgarray(sga)?;
-        if buf.is_empty() {
-            let cause: String = String::from("zero-length buffer");
+        let bufs: ArrayVec<DemiBuffer, DEMI_SGARRAY_MAXLEN> = clone_sgarray(sga)?;
+        if bufs.is_empty() {
+            let cause: String = String::from("zero-length list of buffers");
```
anandbonde: `String::from` can be replaced with string literals and references to cause can be removed while passing into other blocks.

_Note: This comment applies to all other places in this change where `String::from` is used._

src/inetstack/consts.rs
```diff
@@ -37,10 +37,8 @@ pub const TCP_HANDSHAKE_TIMEOUT: Duration = Duration::from_secs(3);
 /// TODO: Auto-Discovery MTU Size
 pub const DEFAULT_MSS: usize = 1450;
 
-/// Length of a [crate::memory::DemiBuffer] batch.
-///
-/// TODO: This Should be Generic
-pub const RECEIVE_BATCH_SIZE: usize = 4;
+/// Max batch size of packets for both transmit and receive up and down the stack.
+pub const MAX_BATCH_SIZE_NUM_PACKETS: usize = 20;
```
anandbonde, line 42: Why is this number bumped up 5x?

src/inetstack/protocols/layer4/mod.rs
```diff
@@ -338,19 +338,39 @@ impl Peer {
...
+        bufs: ArrayVec<DemiBuffer, MAX_BATCH_SIZE_NUM_PACKETS>,
+        addr: Option<SocketAddr>,
+    ) -> Result<(), Fail> {
         match sd {
-            Socket::Tcp(socket) => self.tcp.push(socket, buf).await,
-            Socket::Udp(socket) => self.udp.push(socket, buf, addr).await,
+            Socket::Tcp(socket) => self.tcp.push(socket, bufs).await,
+            Socket::Udp(socket) => {
+                for buf in bufs {
+                    trace!("pushing");
```
anandbonde: If we need a trace here, then we should add slightly more information in this log statement.

src/runtime/types/memory.rs
```diff
@@ -16,6 +18,8 @@ pub const DEMI_SGARRAY_MAXLEN: usize = 1;
 #[repr(C, packed)]
 #[derive(Copy, Clone)]
 pub struct demi_sgaseg_t {
+    /// Reserved for DemiBuffer metadata.
+    pub sgaseg_md: *mut libc::c_void,
```
anandbonde: We should evaluate renaming these fields to avoid redundant prefix `sgaseg`. That will also gives us more space to write meaningful names.

src/runtime/types/ops.rs
```diff
@@ -97,7 +97,7 @@ mod test {
         // Size of a demi_qresult_t structure.
         crate::ensure_eq!(
             mem::size_of::<demi_qresult_t>(),
-            QR_OPCODE_SIZE + QR_QD_SIZE + QR_QT_SIZE + QR_RET_SIZE + QR_VALUE_SIZE
+            QR_OPCODE_SIZE + QR_QD_SIZE + QR_QT_SIZE + QR_RET_SIZE + QR_VALUE_SIZE + 4,
```
anandbonde: It would be good to give a name to this magic const `4`.

anandbonde (changes requested):
The overall approach looks good! Left a few comments.

src/inetstack/mod.rs
iyzhang: yes, let me check if there is a compile assert that I can use

src/inetstack/consts.rs
iyzhang: I bumped it to match the new sgaarray max so that we could potentially get up to one sgarray worth of buffers in a single poll. Otherwise, we will have to poll 4 times to fill a single batch that we return to the application. The max sgarray size was chosen because Kyle said that their RPCs grow up to 20 MTUs in size.

anandbonde: That makes sense! It would be good to log this thought process in form of comment near the const.

src/catnap/linux/transport.rs
iyzhang: Ooops, I should be updating it.

src/inetstack/mod.rs
iyzhang: So it looks like there is a static assert crate but we might not want to add that to the repo, so then the next best option is this: https://internals.rust-lang.org/t/nicer-static-assertions/15986/2

src/runtime/types/memory.rs
iyzhang: yes, I assumed that this was a left over convention from the C code

include/demi/types.h
```diff
@@ -148,8 +148,8 @@ extern "C"
         DemiLogLevel_Trace = 5,
     } demi_log_level_t;
 
-    // Logging callback. Arguments are: level, module name, module length, file name, file name length, line number, message, message length, 
-    typedef void (*demi_log_callback_t)(demi_log_level_t, const char*, uint32_t, const char*, uint32_t, uint32_t, const char*, uint32_t);
+    // Logging callback. Arguments are: level, module name, module length, file name, file name length, line number, message, message length,
+    typedef void (*demi_log_callback_t)(demi_log_level_t, const char *, uint32_t, const char *, uint32_t, uint32_t, const char *, uint32_t);
```
anandbonde: typedefs for function pointers can have parameter names. That would be much better than using comments to log this info becuase then there is no need to change at 2 places, plus the intent will stand out very clearly at a glance.

src/runtime/types/ops.rs
```diff
@@ -27,24 +27,24 @@ pub enum demi_opcode_t {
...
-    pub addr: libc::sockaddr,
+    pub qd: i32,              // 4 bytes.
+    pub addr: libc::sockaddr, // 16 bytes.
 }
 
 #[repr(C)]
 pub union demi_qr_value_t {
-    pub sga: demi_sgarray_t,
-    pub ares: demi_accept_result_t,
+    pub sga: demi_sgarray_t,        // 400 bytes.
```
anandbonde: Hardcoding these sizes in comments is risky. They are prone to going stale and will be misleading in future if that happens. It would be better have these explained in the form of tests, which I believe we already have somewhere in this file already. So we can just delete these size related comments.

src/runtime/types/memory.rs
anandbonde: Possibly... if we remove this prefix, we will have more space to write better/meaningful names for the fields (e.g., `reserved_metadata` would much better than `sgaseg_md`).

include/demi/types.h
iyzhang: possibly not for extern typedefs?

anandbonde: `extern typedef` is not a standard, so I'm not sure what would happen if we decided to make it `extern` (compiler-dependent). But, I believe it should work regardless, because those parameters wouldn't change the type in any way and would just serve the purpose of crystal clear and maintainable documentation.

anandbonde (approved):
Looks good.

PR #1551, opened 2025-05-28
demikernel: cleanup: several source code improvements

anandbonde:
This is a series of patches that improve the overall src code quality by cleaning up unnecessary code and visual clutter. Each commit has the change and reasoning behind it.

### DIFF
--- a/src/inetstack/protocols/layer3/arp/peer.rs
+++ b/src/inetstack/protocols/layer3/arp/peer.rs
@@ -108,18 +108,18 @@ impl SharedArpPeer {
 
     async fn do_wait_link_addr(&mut self, ipv4_addr: Ipv4Addr) -> MacAddress {
         let (tx, rx): (Sender<MacAddress>, Receiver<MacAddress>) = channel();
+
         if let Some(&link_addr) = self.cache.get(ipv4_addr) {
             let _ = tx.send(link_addr);
+        } else if let Some(wait_queue) = self.waiters.get_mut(&ipv4_addr) {
+            warn!("Duplicate waiter for IP address: {}", ipv4_addr);
+            wait_queue.push_back(tx);
         } else {
-            if let Some(wait_queue) = self.waiters.get_mut(&ipv4_addr) {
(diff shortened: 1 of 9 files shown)


src/inetstack/protocols/layer3/arp/peer.rs
```diff
@@ -108,18 +108,18 @@ impl SharedArpPeer {
 
     async fn do_wait_link_addr(&mut self, ipv4_addr: Ipv4Addr) -> MacAddress {
         let (tx, rx): (Sender<MacAddress>, Receiver<MacAddress>) = channel();
+
         if let Some(&link_addr) = self.cache.get(ipv4_addr) {
             let _ = tx.send(link_addr);
+        } else if let Some(wait_queue) = self.waiters.get_mut(&ipv4_addr) {
```
iyzhang, line 114: Is there some match condition that might make this easier to read?

iyzhang (approved):

src/inetstack/protocols/layer3/arp/peer.rs
anandbonde: Hmm, this is one way to do it, but seems very similar to the previous nested-if-based version...

```
match self.cache.get(ipv4_addr) {
    Some(&link_addr) => {
        let _ = tx.send(link_addr);
    }
    None => match self.waiters.get_mut(ipv4_addr) {
        Some(wait_queue) => {
            warn!("Duplicate waiter for IP address: {}", ipv4_addr);
            wait_queue.push_back(tx);
        }
        None => {
            let mut wait_queue: LinkedList<Sender<MacAddress>> = LinkedList::new();
            wait_queue.push_back(tx);
            self.waiters.insert(ipv4_addr, wait_queue);
        }
    },
}
```
Do you have any preference? Please feel free to suggest any alternatives, I'll be happy to update to the most simple version!

iyzhang: yeah your original version seems better. Thanks!

PR #1552, opened 2025-05-28
githook: relax commit msg restrictions

anandbonde:
While writing many commit messages, I realized that writing a (Workaround|Enhancement|Feature|Bug Fix) tag is unnecessary. We don't do any reporting on this, plus it's cumbersome to write. Better to add more context in the top commit msg instead. We should get rid of this restriction.


iyzhang (approved):
Makes sense!

PR #1555, opened 2025-05-29
inetstack: layer4: tcp: sender: remove dup check

anandbonde:
This check is duplicate, and seems like an artifact of prior merge issue.

### DIFF
--- a/src/inetstack/protocols/layer4/tcp/established/sender.rs
+++ b/src/inetstack/protocols/layer4/tcp/established/sender.rs
@@ -458,12 +458,6 @@ impl Sender {
             );
         }
 
-        if cb.sender.unacked_queue.len() > 0 {
-            trace!(
-                "send_segment(): unacked_queue.len() = {:?}",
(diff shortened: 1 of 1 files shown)


src/inetstack/protocols/layer4/tcp/established/sender.rs
```diff
@@ -458,12 +458,6 @@ impl Sender {
             );
         }
 
-        if cb.sender.unacked_queue.len() > 0 {
```
iyzhang, line 461: I think @kyleholohan was printing this for some debugging but it might not be necessary any more.

iyzhang (changes requested):

src/inetstack/protocols/layer4/tcp/established/sender.rs
anandbonde: This block of code is identical to the one immediately before it, so it’s safe to delete it.

iyzhang (approved):

PR #1556, opened 2025-05-30
demikernel: use is_empty() instead of len() comparisons to 0

anandbonde:
Some structures can answer .is_empty() much faster than calculating their length. So it is good to get into the habit of using .is_empty(), and having it is cheap. Besides, it makes the intent clearer than a manual comparison in some contexts.


iyzhang (approved):
Agreed! Thanks for doing this!

PR #1557, opened 2025-05-30
Inetstack: Remove duplicate ICMP code

iyzhang:
Just removes some duplicate code in the ICMPv4 peer implementation.

### DIFF
--- a/src/inetstack/protocols/layer3/icmpv4/peer.rs
+++ b/src/inetstack/protocols/layer3/icmpv4/peer.rs
@@ -110,6 +110,7 @@ impl SharedIcmpv4Peer {
     /// Background task for replying to ICMP messages.
     async fn poll(mut self) {
         loop {
+            // Look at incoming request.
             let (ipv4_hdr, mut buf): (Ipv4Header, DemiBuffer) = match self.recv_queue.pop(Some(PING_TIMEOUT)).await {
                 Ok(result) => result,
                 Err(_) => break,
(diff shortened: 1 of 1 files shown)


src/inetstack/protocols/layer3/icmpv4/peer.rs
```diff
@@ -252,6 +231,26 @@ impl SharedIcmpv4Peer {
...
+        request: Icmpv4Type2,
+        dst_ipv4_addr: Ipv4Addr,
+        mut buf: DemiBuffer,
+    ) -> Result<(), Fail> {
+        debug!("initiating ARP query");
+        let dst_link_addr: MacAddress = self.arp.query(dst_ipv4_addr).await?;
+        debug!("ARP query complete ({} -> {})", dst_ipv4_addr, dst_link_addr);
+        debug!("ping ({}, {:?})", dst_ipv4_addr, request);
+        // Send reply message.
+        let local_ipv4_addr: Ipv4Addr = self.local_ipv4_addr;
```
anandbonde: Is this always a reply?

anandbonde: We can get rid of this temporary variable now that it's in a nice self-contained block.

anandbonde (approved):
Looks good.

src/inetstack/protocols/layer3/icmpv4/peer.rs
iyzhang: I just took out the comment because it wasn't really adding anything

PR #1563, opened 2025-05-31
demikernel: use shorter assign op pattern

anandbonde:
Use `a += b` instead of `a = a + b`. This makes lines shorter and faster to read. So why not use it while the language supports this feature.


iyzhang (approved):
lol, I thought Rust didn't support this for a while.

PR #1575, opened 2025-06-05
catnapw: cleanup: patch series for code improvements

anandbonde:
This is a series of patches with serveral code improvements targeting CatnapW. It will be easier to walk through the list of commits because each commit is self-contained, groups similar changes together and has reasoning behind the changes.


anandbonde:
Will send out individual PRs because that's easier to review.

PR #1585, opened 2025-06-10
demikernel: v0: series of patches with code improvements

anandbonde:
This is a series of patches that focus on code clarity and simplification. Refer to individual commits for reasoning.

### DIFF
--- a/src/inetstack/protocols/layer4/tcp/peer.rs
+++ b/src/inetstack/protocols/layer4/tcp/peer.rs
@@ -109,7 +109,7 @@ impl SharedTcpPeer {
         // All other checks should have been done already.
         debug_assert!(!Ipv4Addr::is_unspecified(local.ip()));
         debug_assert!(local.port() != 0);
-        debug_assert!(self.addresses.get(&SocketId::Passive(local)).is_none());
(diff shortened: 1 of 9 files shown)


src/inetstack/protocols/layer4/tcp/peer.rs
```diff
@@ -109,7 +109,7 @@ impl SharedTcpPeer {
         // All other checks should have been done already.
         debug_assert!(!Ipv4Addr::is_unspecified(local.ip()));
         debug_assert!(local.port() != 0);
-        debug_assert!(self.addresses.get(&SocketId::Passive(local)).is_none());
```
iyzhang, line 112: We have a few of these in the runtime/mod.rs file as well.

iyzhang (approved):

src/inetstack/protocols/layer4/tcp/peer.rs
anandbonde: There is 1 instance in the runtime/mod.rs, but that is because the IDMap does not have a `contains_key` method. We can change this in a separate PR if required.

PR #1587, opened 2025-06-11
Scheduler: Run task on insert

iyzhang:
This PR changes our scheduler to run a task until it yields immediately on insert. If the task completes, it is still assigned a queue token and the result is stored for later retrieval.


src/runtime/mod.rs
```diff
@@ -178,6 +178,21 @@ impl SharedDemiRuntime {
                 // Stash the newly allocated qtoken in the task.
                 Ok(qt)
             },
+            Some(InsertResult::Completed(mut boxed_task)) => {
+                trace!("Completed while inserting coroutine: {:?}", boxed_task.get_name());
+                let qt: QToken = self
+                    .qtoken_to_scheduler_id
+                    .insert_with_new_id((0 as u64).into())
```
anandbonde: It would be better to use `0_u64` (literal constant) instead of `0 as u64` (runtime typecast). That would make the intent clearer and no perf cost at runtime (even if the compiler maybe smart to optimize it).

src/demikernel/libos/network/queue.rs
```diff
@@ -247,67 +214,56 @@ impl<T: NetworkTransport> SharedNetworkQueue<T> {
...
     pub async fn close_coroutine(&mut self) -> Result<(), Fail> {
-        match self.transport.clone().close(&mut self.socket).await {
-            Ok(()) => {
-                self.state_machine.prepare(SocketOp::Closed)?;
-                self.state_machine.commit();
-                Ok(())
-            },
-            Err(e) => Err(e),
+        self.state_machine.closing();
+        {
```
anandbonde: Is this artificial scoping intended to express clarity/separation of responsibilities, or for scoping the variables inside? If it's intended for clarity, the inner block can be a separate well-named fn instead. That ways, the reader can easily grasp what's happening.

src/runtime/scheduler/scheduler.rs
```diff
@@ -312,12 +269,13 @@ mod tests {
         const NUM_TASKS: usize = 1024;
         let mut task_ids: Vec<SchedulerId> = Vec::<SchedulerId>::with_capacity(NUM_TASKS);
 
-        for val in 0..NUM_TASKS {
+        for val in 1..NUM_TASKS {
```
anandbonde, line 272: Why was the starting point of the `for` changed from 0 to 1? Same for the other instance below?

tests/rust/udp.rs
anandbonde, line 1: Are the changes in this file required to achieve this functionality? If not, could they be a part of a separate commit?

anandbonde:
It would be very useful for reading if these changes can be broken down into a sequence of smaller commits. Are there logical groups of changes that can be compiled/tested separately?

src/runtime/mod.rs
iyzhang: Good idea!

src/demikernel/libos/network/queue.rs
iyzhang: It is for scoping ownership for the socket. I wanted to avoid cloning if not necessary, but I will use a separate function.

src/runtime/scheduler/scheduler.rs
iyzhang: Because if the val is 0, then it will be completed on insert and never polled.

tests/rust/udp.rs
iyzhang: Yes because we now send the packet on inserting the coroutine, so without the additional coordination, the packet was being sent and dropped before the receiving packet was created.

src/demikernel/libos/network/libos.rs
```diff
@@ -267,11 +267,12 @@ impl<T: NetworkTransport> SharedNetworkLibOS<T> {
             Ok(queue) => queue,
             Err(e) => return (qd, OperationResult::Failed(e)),
         };
+        let local: Option<SocketAddr> = queue.local();
```
anandbonde, line 271: It would be good to move the declaration of `local` near it's usage. Unless you want to cache the value across the `await` below.

anandbonde (approved):
Looks good.

src/demikernel/libos/network/libos.rs
iyzhang: yes, I do want to cache the value because once it closes, the value disappears.

PR #1591, opened 2025-06-19
demikernel: v3: series of patches with code improvements

anandbonde:
This is v3 of the series of patches that focus on code clarity and simplification. Refer to individual commits for reasoning.

### DIFF
--- a/src/demikernel/bindings.rs
+++ b/src/demikernel/bindings.rs
@@ -48,14 +48,14 @@ pub extern "C" fn demi_init(args: *const demi_args_t) -> c_int {
         args.log_callback
     };
 
-    if log_callback.is_none() {
-        logging::initialize();
-    } else {
+    if let Some(log_callback) = log_callback {
         logging::custom_initialize(move || {
             Logger::try_with_env()
                 .unwrap()
-                .log_to_writer(Box::new(CallbackLogWriter::new(log_callback.unwrap())))
+                .log_to_writer(Box::new(CallbackLogWriter::new(log_callback)))
         });
+    } else {
+        logging::initialize();
     }
 
     trace!("demi_init()");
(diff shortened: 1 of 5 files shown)


src/demikernel/bindings.rs
```diff
@@ -48,14 +48,14 @@ pub extern "C" fn demi_init(args: *const demi_args_t) -> c_int {
         args.log_callback
     };
 
-    if log_callback.is_none() {
-        logging::initialize();
-    } else {
+    if let Some(log_callback) = log_callback {
```
iyzhang, line 51: do we prefer a match statement there or the if let?

iyzhang (approved):

src/demikernel/bindings.rs
anandbonde: The Rust ebook says that `if let ... else` is a more concise way to represent this logic (especially when we care about just 2 patterns which is exhaustive in this case).

https://doc.rust-lang.org/stable/book/ch06-03-if-let.html

Match would result in a slightly more verbose representation and is generally good for exhaustive pattern match for more than 2 obvious patterns, but has the same possible patterns for a boolean like in this case. I don't have strong opinion about this, as far as we are consistent and the code flows easily for the reader.

Here are the 2 impls for quick comparison:

### if let ... else
```
if let Some(log_callback) = log_callback {
    logging::custom_initialize(move || {
        Logger::try_with_env()
            .unwrap()
            .log_to_writer(Box::new(CallbackLogWriter::new(log_callback)))
    });
} else {
    logging::initialize();
}
```

### match
```
match log_callback {
    Some(log_callback) => {
        logging::custom_initialize(move || {
            Logger::try_with_env()
                .unwrap()
                .log_to_writer(Box::new(CallbackLogWriter::new(log_callback)))
        });
    }
    None => {
        logging::initialize();
    }
}
```

PR #1594, opened 2025-06-25
demikernel: v5: series of patches with code improvements

anandbonde:
This is v5 of the series of patches that focus on code clarity and simplification. Refer to individual commits for reasoning.

### DIFF
--- a/src/runtime/scheduler/scheduler.rs
+++ b/src/runtime/scheduler/scheduler.rs
@@ -195,11 +195,11 @@ mod tests {
 
         // All futures are inserted in the scheduler with notification flag set.
         // By polling once, our future should complete.
-        if let Some(_) = scheduler.poll_group_once(group_id, None).pop() {
-            Ok(())
-        } else {
+        if scheduler.poll_group_once(group_id, None).pop().is_none() {
             anyhow::bail!("task should have completed")
         }
+
+        Ok(())
     }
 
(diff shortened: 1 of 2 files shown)


src/runtime/scheduler/scheduler.rs
```diff
@@ -195,11 +195,11 @@ mod tests {
 
         // All futures are inserted in the scheduler with notification flag set.
         // By polling once, our future should complete.
-        if let Some(_) = scheduler.poll_group_once(group_id, None).pop() {
-            Ok(())
-        } else {
+        if scheduler.poll_group_once(group_id, None).pop().is_none() {
```
iyzhang, line 198: I think this could be further turned into an ok_or or an ok_or_else: https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or_else

iyzhang (approved):

src/runtime/scheduler/scheduler.rs
anandbonde: That would be great, however `bail!()` does not return any value, it just returns early from the fn. `ok_or` or `ok_or_else` require a return value from the expression. We could change from `anyhow::bail!()` to `anyhow::anyhow!()` to construct the error, but then we would end up changing the semantics of this code. Do you prefer any particular style in this case?

PR #1596, opened 2025-06-25
profiler: fixing scoping issue with async profiling

### LINKED ISSUE #1593: Profiler: Crash if allocating coroutine inside a timer context
## Description

Since #1587, we now run each coroutine on insert. Since coroutines were previously rooted, it was impossible to get the old context back after allocating a coroutine, so if we allocated and run a coroutine inside of a timer context, we would lose the beginning of that context.

## How to Reproduce

1. Run any test with the profiler on.
2. Allocate a new coroutine on receiving a packet.
3. Panic when leaving the poll_once context because the new coroutine erased the previous state.

## Expected Behavior

We should not erase the previous context when running a coroutine.

iyzhang:
Fixes #1593. Previously we were not properly tracking the current running profiler scope and replacing it if we run an asynchronous block. Originally this was not an issue because we only ran async blocks in wait_* calls but now that we run them on insert, that is no longer true.

### DIFF
--- a/src/perftools/profiler/mod.rs
+++ b/src/perftools/profiler/mod.rs
@@ -1,7 +1,11 @@
 // Copyright(c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-//! This module provides a small performance profiler for the Demikernel libOSes.
+//! This module provides a small performance profiler for the Demikernel libOSes. It supports two kinds of profilers:
+//! ones for synchronous blocks of code and ones for asynchronous blocks of code. For synchronous blocks, we use a
+//! [SyncScopeGuard] to automatically scope the timer. The profiler creates the [SyncScopeGuard] when the timer starts
+//! and the timer stops when the [SyncScopeGuard] goes out of scope. For asynchronous blocks, we simply create a
+//! [SharedScope] and time the number of cycles for each poll. We do not use a guard for scoping asynchronous blocks.
 
 //======================================================================================================================
 // Exports
@@ -17,7 +21,7 @@ mod tests;
 //======================================================================================================================
 
 use crate::{
-    perftools::profiler::scope::{Guard, SharedScope},
+    perftools::profiler::scope::{SharedScope, SyncScopeGuard},
     runtime::{types::demi_callback_t, SharedObject},
 };
 use ::futures::future::FusedFuture;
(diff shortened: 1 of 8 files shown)


src/runtime/scheduler/group.rs
```diff
@@ -74,7 +74,7 @@ impl TaskGroup {
     }
 
     /// Insert a new task into our scheduler returning a handle corresponding to it.
-    pub fn insert(&mut self, task: Box<dyn Task>) -> Option<InsertResult> {
+    pub fn insert_and_poll_task(&mut self, task: Box<dyn Task>) -> Option<InsertResult> {
```
anandbonde, line 77: It would be good to commit this rename change under a separate commit id for better tracking.

src/perftools/profiler/scope.rs
```diff
@@ -193,17 +178,22 @@ impl<'a, F: Future> Future for AsyncScope<'a, F> {
     fn poll(self: Pin<&mut Self>, ctx: &mut Context<'_>) -> Poll<Self::Output> {
         let self_: &mut Self = self.get_mut();
 
-        let _guard = PROFILER.with(|p| p.clone().enter_scope(&self_.scope));
-        Future::poll(self_.future.as_mut(), ctx)
+        PROFILER.with(|p| p.clone().create_and_enter_async_scope(&self_.name));
```
anandbonde: self_.name is already a reference, so we can just use it without the &
```suggestion
        PROFILER.with(|p| p.clone().create_and_enter_async_scope(self_.name));
```

src/perftools/profiler/mod.rs
```diff
@@ -68,135 +66,128 @@ pub fn set_callback(perf_callback: demi_callback_t) {
...
+        in_ns as f64 / (end_cycle - start_cycle) as f64
     }
 }
 
-impl Profiler {
-    fn new() -> Profiler {
-        Self {
+impl SharedProfiler {
+    pub fn new() -> SharedProfiler {
+        Self(SharedObject::new(Profiler {
```
anandbonde: Can we use the `Default` impl for `SharedProfiler` instead? As this constructor is doing everything that a default would.

anandbonde (approved):

PR #1602, opened 2025-06-30
pdpix: mark the API fns unsafe

anandbonde:
The PDPIX API performs unsafe Rust operations by means of dealing with pointers passed in by the application. Hence, these operations must result into an `unsafe` block wrapper around call sites in the Rust code that uses the API.

This also encourages us to add Safety docs for the API functions.

### DIFF
--- a/src/demikernel/bindings.rs
+++ b/src/demikernel/bindings.rs
@@ -37,9 +37,12 @@ thread_local! {
     static THREAD_LOCAL_LIBOS: RefCell<Option<LibOS>> = RefCell::new(None);
 }
 
+/// # Safety
+///
+/// The `args` argument to this function must be valid and point to a `demi_args_t` structure.
 #[allow(unused)]
 #[no_mangle]
-pub extern "C" fn demi_init(args: *const demi_args_t) -> c_int {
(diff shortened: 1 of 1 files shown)


src/demikernel/bindings.rs
```diff
@@ -207,20 +216,18 @@ pub extern "C" fn demi_accept(qtok_out: *mut demi_qtoken_t, sockqd: c_int) -> c_
...
     size: Socklen,
 ) -> c_int {
-    trace!("demi_connect()");
-
-    // Check for invalid storage location.
-    if qtok_out.is_null() {
-        warn!("demi_connect() qtok_out is a null pointer");
-        return libc::EINVAL;
-    }
+    trace!("demi_connect()"); // Check for invalid storage location. if qtok_out.is_null() { warn!("demi_connect() qtok_out is a null pointer"); return libc::EINVAL; }
```
iyzhang: Is this meant to be on the next line?

iyzhang (approved):

src/demikernel/bindings.rs
anandbonde: Yes, thanks for the catch! Fixed.

PR #1603, opened 2025-07-01
demikernel: v9: series of patches with code improvements

anandbonde:
This is v9 of the series of patches that focus on code clarity and simplification. Refer to individual commits for reasoning.

### DIFF
--- a/src/demikernel/bindings.rs
+++ b/src/demikernel/bindings.rs
@@ -14,10 +14,7 @@ use crate::{
     runtime::{
         fail::Fail,
         logging::{self, CallbackLogWriter},
-        types::{
-            demi_args_t, demi_callback_t, demi_log_callback_t, demi_qresult_t, demi_qtoken_t, demi_sgarray_t,
-            demi_sgaseg_t,
-        },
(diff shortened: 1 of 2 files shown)


src/demikernel/bindings.rs
```diff
@@ -28,23 +25,22 @@ use ::socket2::SockAddr;
...
     mem::{self, MaybeUninit},
-    net::{SocketAddr, SocketAddrV4},
+    net::SocketAddr,
     ptr, slice,
     time::Duration,
 };
 
 thread_local! {
-    static THREAD_LOCAL_LIBOS: RefCell<Option<LibOS>> = RefCell::new(None);
+    static THREAD_LOCAL_LIBOS: RefCell<Option<LibOS>> = const { RefCell::new(None) };
```
iyzhang, line 34: Is there a reason that we have this as an Option and fill it later?

iyzhang (approved):

src/demikernel/bindings.rs
anandbonde: The actual LibOS instance which overwrites this value is created at runtime based on config. However, this assignment (of default value) can happen at compile time.

iyzhang: ahh yeah got it. I was wondering what was missing that we couldn't create the object at compile time.

PR #1604, opened 2025-07-01
ci: workflows: ignore missing files when release build is skipped

anandbonde:
Now that release builds are optional, there is no need to see the warning when the upload artifact step does not find files to upload becuase the previous step (i.e. the release build) was skipped.


anandbonde:
GitHub does not yet support `if-files-not-found: ignore` in the upload artifact step. So we will bring this change back when there is official support. Closing this PR for now.

PR #1613, opened 2025-07-10
scheduler: cleanup: simplify logic and cleanup

anandbonde:
Simplify scheduler and group logic, clearer naming and reduced syntactic clutter.

### DIFF
--- a/src/runtime/scheduler/group.rs
+++ b/src/runtime/scheduler/group.rs
@@ -17,7 +17,7 @@ use crate::{
     runtime::scheduler::{
         page::{WakerPageRef, WakerRef},
         waker64::{WAKER_BIT_LENGTH, WAKER_BIT_LENGTH_SHIFT},
-        InsertResult, Task,
+        ScheduleResult, Task,
     },
 };
 use ::bit_iter::BitIter;
@@ -49,7 +49,6 @@ pub struct TaskGroup {
 //======================================================================================================================
 
 impl TaskGroup {
-    /// Remove task by handle.
     pub fn remove(&mut self, idx: usize) -> Option<Box<dyn Task>> {
         let (page_idx, offset) = self.locate(idx)?;
         self.pages[page_idx].clear(offset);
@@ -68,7 +67,7 @@ impl TaskGroup {
         }
     }
 
-    pub fn insert(&mut self, task: Box<dyn Task>) -> Option<InsertResult> {
+    pub fn schedule(&mut self, task: Box<dyn Task>) -> Option<ScheduleResult> {
         let name = task.name();
         let idx = self.tasks.insert(task)?;
 
@@ -77,11 +76,11 @@ impl TaskGroup {
         let (page_idx, offset) = self.locate(idx)?;
         self.pages[page_idx].initialize(offset);
 
-        trace!("insert(): name={:?}, idx={:?}", name, idx);
+        trace!("schedule(): name={:?}, idx={:?}", name, idx);
 
         match self.run_task(idx) {
-            Some(task) => Some(InsertResult::Completed(task)),
-            None => Some(InsertResult::Inserted(idx.into())),
+            Some(task) => Some(ScheduleResult::Completed(task)),
+            None => Some(ScheduleResult::Scheduled(idx.into())),
         }
     }
 
--- a/src/runtime/scheduler/group.rs
+++ b/src/runtime/scheduler/group.rs
@@ -25,8 +25,7 @@ use ::futures::Future;
 use ::std::{
     iter::Flatten,
     pin::Pin,
-    ptr::NonNull,
-    task::{Context, Poll, Waker},
+    task::{Context, Waker},
     vec::IntoIter,
 };
 
@@ -34,15 +33,14 @@ use ::std::{
 // Structures
 //======================================================================================================================
 
-/// This represents a resource management group. All tasks belong to a task group. By default, a task belongs to the
-/// same task group as the allocating task.
+/// Resource management group; all tasks belong to one.
+/// By default, tasks inherit the allocator's group.
 #[derive(Default)]
 pub struct TaskGroup {
-    /// Stores all the tasks that are held by the scheduler.
+    /// All scheduled tasks.
     tasks: PinSlab<Box<dyn Task>>,
-    /// Holds the waker bits for controlling task scheduling.
-    waker_page_refs: Vec<WakerPageRef>,
-    /// List of ready tasks.
+    /// Waker pages controlling scheduling.
+    pages: Vec<WakerPageRef>,
     ready_tasks: Flatten<IntoIter<Vec<usize>>>,
 }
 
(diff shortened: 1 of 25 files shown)


src/runtime/scheduler/page/waker_ref.rs
```diff
@@ -92,17 +92,17 @@ impl Drop for WakerRef {
...
         // Increment reference count.
         mem::forget(self);
         waker
     }
 }
 
 unsafe fn waker_ref_clone(ptr: *const ()) -> RawWaker {
-    let p: WakerRef = WakerRef(NonNull::new_unchecked(ptr as *const u8 as *mut u8));
-    let q: WakerRef = p.clone();
+    let p = WakerRef(NonNull::new_unchecked(ptr as *const u8 as *mut u8));
```
iyzhang: Do you want to give this a full variable name?

src/runtime/scheduler/page/waker_ref.rs
```diff
@@ -142,130 +142,121 @@ mod tests {
...
     }
 
     #[test]
     fn test_wake() -> Result<()> {
-        let p: WakerPageRef = WakerPageRef::default();
-        crate::ensure_eq!(p.refcount_get(), 1);
+        let page = WakerPageRef::default();
+        crate::ensure_eq!(page.refcount(), 1);
+
+        let q = WakerRef::new(page.as_raw_waker_ref(0));
```
iyzhang: same here? Something like raw_waker_ref or page_waker_ref

src/runtime/scheduler/group.rs
```diff
@@ -51,170 +49,140 @@ pub struct TaskGroup {
...
-    pub fn remove(&mut self, pin_slab_index: usize) -> Option<Box<dyn Task>> {
-        // We should not have a scheduler handle that refers to an invalid id, so unwrap and expect are safe here.
-        let (waker_page_ref, waker_page_offset): (&WakerPageRef, usize) = {
-            let (waker_page_index, waker_page_offset) = self.get_waker_page_index_and_offset(pin_slab_index)?;
-            (&self.waker_page_refs[waker_page_index], waker_page_offset)
-        };
-        waker_page_ref.clear(waker_page_offset);
-        if let Some(task) = self.tasks.remove_unpin(pin_slab_index) {
+    /// Remove task by handle.
+    pub fn remove(&mut self, idx: usize) -> Option<Box<dyn Task>> {
```
iyzhang, line 52: I feel like the difference between idx and i is no longer clear. We should at least include the data structures that each index into in the variable name.

src/runtime/scheduler/group.rs
```diff
@@ -51,170 +49,140 @@ pub struct TaskGroup {
...
+            self.pages.push(WakerPageRef::default());
         }
     }
 
-    pub fn get_num_waker_pages(&self) -> usize {
-        self.waker_page_refs.len()
-    }
-
-    fn get_waker_page_offset(pin_slab_index: usize) -> usize {
+    fn get_page_offset(pin_slab_index: usize) -> usize {
```
iyzhang, line 109: should we get rid of the get_ on these function names as well and just name it page_offset?

src/runtime/scheduler/group.rs
```diff
@@ -51,170 +49,140 @@ pub struct TaskGroup {
...
+        while let Some(offset) = self.next(iter == 0 || refill) {
+            if let Some(task) = self.run_task(offset) {
+                tasks.push(task);
             }
-            check_for_additonal_tasks = keep_checking_for_new_tasks;
-            match max_iterations {
-                Some(max_iterations) if iterations_run >= max_iterations => return completed_tasks,
-                _ => iterations_run += 1,
+
+            if let Some(limit) = limit {
```
iyzhang: this could be written as one line with is_some_and

src/runtime/scheduler/scheduler.rs
```diff
@@ -55,40 +55,27 @@ impl Scheduler {
...
-    fn get_mut_group(&mut self, id: SchedulerId) -> Option<&mut TaskGroup> {
+    fn get_group_mut(&mut self, id: SchedulerId) -> Option<&mut TaskGroup> {
         self.groups.get_mut(id.into())
     }
 
-    /// The parent id can either be the id of the group or another task in the same group.
-    pub fn insert_and_poll_task<T: Task>(&mut self, group_id: SchedulerId, task: T) -> Option<InsertResult> {
-        let group: &mut TaskGroup = self.get_mut_group(group_id)?;
-        group.insert_and_poll_task(Box::new(task))
+    pub fn insert<T: Task>(&mut self, group_id: SchedulerId, task: T) -> Option<InsertResult> {
```
iyzhang: I think we might want to name this insert_and_run task, just to make it really clear that we are going to trigger the task right away.

src/runtime/scheduler/scheduler.rs
```diff
@@ -55,40 +55,27 @@ impl Scheduler {
...
-        group_id: SchedulerId,
-        max_iterations: Option<usize>,
-    ) -> Vec<Box<dyn Task>> {
-        // Expect is safe here because something has really gone wrong if we are polling a group that doesn't exist.
-        let group: &mut TaskGroup = self.get_mut_group(group_id).expect("group being polled doesn't exist");
-
-        // Keep polling the group and checking for runnable tasks until there are none left.
-        group.poll_group(max_iterations, true)
+    /// Do not use this function on coroutines that use poll_yield, unless limit is set.
+    pub fn poll_group(&mut self, id: SchedulerId, limit: Option<usize>) -> Vec<Box<dyn Task>> {
```
iyzhang: we should probably settle on either poll or run as the verb in all of our variable names

iyzhang (changes requested):

src/runtime/scheduler/group.rs
anandbonde: These are non-instance fns, so having `get_*` namesreduces ambiguity.

> e.g. `Self::page_offset(idx)` is confusing whether it's querying or setting the offset.

However, for instance fns it seems more natural to remove the prefix.

> e.g. `page.offset()`

I think we can leave them as they are, unless you have a preference.

src/runtime/scheduler/scheduler.rs
anandbonde: The "run" part is an implementation detail so I left it out. Another option is to invert the implementation detail responsibility and just call it `run` such that "insert" is just a side-effect of running.

We should evaluate how it may be perceived by the reader... Thoughts?

src/runtime/scheduler/scheduler.rs
anandbonde: I've changed it, let me know what you think.

src/runtime/scheduler/group.rs
iyzhang: yes, this makes sense.

src/runtime/scheduler/scheduler.rs
iyzhang: yes, maybe we could call it schedule_task to indicate that its being added to the scheduler and scheduled to run.

src/runtime/scheduler/scheduler.rs
iyzhang: Great

iyzhang (approved):

src/runtime/scheduler/scheduler.rs
anandbonde: Changed all the related fns to be `schedule*` so that it flows naturally with the functionality.

PR #1619, opened 2025-07-16
Create dependabot.yml for weekly updates on dependencies

JamieMagee:
This change enables weekly Dependabot updates for the following ecosystems:
- GitHub Actions
- Cargo
- Rust Toolchain

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,17 @@
+version: 2
+enable-beta-ecosystems: true
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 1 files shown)


anandbonde:
> This change enables weekly Dependabot updates for the following ecosystems:
> 
> * GitHub Actions
> * Cargo
> * Rust Toolchain

Hi @JamieMagee what’s the purpose of this change? Our code is sensitive to the rust tool chain version change.

Also, we don't support running CI on branches outside of our repo. And PRs cannot be pulled unless the CI passes. The only way to work around this is to create a new branch in our repo and submit a PR that way.

PR #1626, opened 2025-09-25
runtime: Add max buffer size to inetstack from the network runtime

iyzhang:
This PR fixes an issue where the inetstack was using the default implementation of `max_buffer_size_bytes()` rather than getting it from the underlying physical layer. For example, in Catnip, this meant that we were not using the configured DPDK max_body_size.


anandbonde (approved):
Looks good!

iyzhang:
Accidentally closed

